PR #4, opened 2020-02-02
Replace TSLint with ESLint

imbsky:
As the title say.


imbsky:
TSLint was deprecated last year and it's now common to check TypeScript with ESLint.

rgrinberg:
Looks good. Why did you change the indentation to 2 spaces? I'm looking at all the plugins published by Microsoft and they all seem to use 4 spaces.

imbsky:
Microsoft's repository is a bit old overall, as you can see from the fact that TSLint still exists. The most recent project basically uses Prettier, but this time when I created this PR, Prettier automatically formatted JSON in my local environment, so it's two spaces. I can override the config if you prefer 4 spaces, but it's probably better to always use the default config because of Prettier's philosophy. If this PR is merged, I will create a PR that introduces Prettier (#6), so we will be able to unify at that time.

PR #15, opened 2020-02-05
Ignore .vscode-test directory

imbsky:


imbsky:
I will do this in another PR.

PR #18, opened 2020-02-10
Improve things about languages and grammars section

imbsky:


rgrinberg:
Looks good to me. It seems like `source.ocaml.hover.type` and `source.ocaml.hover.info` are unused, is that intentional? Do you plan to use it in another PR?

imbsky:
Yes, but it's not hard to add it in another PR, and it deviates from the purpose of this PR, so I removed it!

rgrinberg:
I’ve looked into the reason server and it seems like it’s used for the hover pop up. I suppose we could do that as well in our lsp server.

Feel free to add it to vscode whenever, but make sure to add an appropriate issue to ocaml lsp so that we have proper support for it

imbsky:
Sure.

PR #19, opened 2020-02-11
Merge vscode-merlin

### LINKED ISSUE #16: Merging vscode-merlin and vscode-ocaml-platform
Hi everyone,

For sometime now, we have been working on (and using) [vscode-merlin](https://github.com/arrowresearch/vscode-merlin) - a vscode
extension that talks to a language server based on
merlin. This has been in the works since merlin lsp was still hosted
at the merlin repo 🙂 We think a lot of our goals overlap with yours.

### `vscode-merlin`'s goals
The goal is to provide a single plugin supporting `opam`, `esy` and `npm/bsb`
workflow. We hope to address some of the concerns regarding the
divide in the ecosystems. Hopefully, this extension brings both the
npm/bucklescript community and Reason/OCaml community that compiles
natively together.

This extension intends to provide first class support for all major
native development tools and relevant file extensions - including and
not limited to syntactic support for dune files, opam files, *.mly,
*.mll. It intends to leverage all features opam and esy that can
assist the respective workflows.

### Bucklescript support
We started working on this extension to gain benefits of
recoverable/incremental parsing that merlin provides for bucklescript
projects. After the merger, we intend to continue to involving
ourselves in this regard. Quick overview of how we setup merlin for
bucklescript.

Providing merlin support for bucklescript projects (supporting both
Reason and OCaml syntaxes) can get tricky - users have unpredictable
environment state and equally unpredictable tools available on their
path. We find that `esy` provides a convenient way out - it provides a
sandbox that ensures the correct version of merlin (compiled with
4.06) is used and that it can find the corresponding ocamlmerlin-reason
binary on it’s path in silo. So we turn every bucklescript projects into
a hidden esy project by dropping a `esy.json` in the `.vscode/` folder
(the UX around this is up for suggestions and feedback). We take certain
[heuristics/practices](https://github.com/arrowresearch/vscode-merlin/blob/master/src/ProjectType.re#L45-L67) into consideration before we do this - currently this is an automatic
setup behind the scenes. We can provide a more manual/intervening
workflow with ideas like a project status view that vscode provides. 

An important concern this raised is the time to get started - the
toolchain (compiler+merlin+refmt+ocamlformat) needs to be compiled
from source and this can take a while. Particularly on Windows. To
address this, we build the toolchain for the three platforms on the
CI once. Since esy's artifacts are relocatable - we download the cache
(about 130mb) and warm up the local cache. This has worked very well
for us. A preliminary setup (with manual intervention) has been tested
(did so myself) on a Windows machine too.

This makes two things necessary for the plugin

1. It relies on esy and will prompt the user to install esy. This can
   be avoided for pure opam workflows of course.
2. It needs Azure Pipelines since it provides a convenient way to
   publish and restore relocatable assets.
   
   
This has worked very well for us so far and hope the community - both
OCaml and Reason using either opam/esy/npm workflows - likes it.

One last thing left would be Windows support for Opam - we're open for
support this workflow too.

### Closing comments
As I understand, merging the projects would
affect the e2e tests - at vscode-merlin we have run e2e tests on all
three platform to ensure the extensions operations and caching
mechanism work on all three OSes. This will affect @imbsky 's re-write
of tests to TS. We'll have to use Azure Pipelines for CI - the setup
at vscode-merlin sees about 20 mins of well cached runs and a tried and
tested relocatable artifacts creation for all three platforms.

As always, I would love your thoughts :)

prometheansacrifice:
Closes #16 

Merger notes

# vscode-merlin and vscode-ocaml-platform merger

## Contribution points

``` json
{
  "contributes": {
"jsonValidation": [
  {
    "fileMatch": "esy.json",
    "url": "https://raw.githubusercontent.com/esy/esy-schema/master/esySchema.json"
  },
  {
    "fileMatch": "package.json",
    "url": "https://raw.githubusercontent.com/esy/esy-schema/master/esySchema.json"
  }
],
"commands": [
  {
    "command": "merlin-language-server.restart",
    "title": "Restart Merlin Language Server"
  }
],
"languages": [
  {
    "id": "reason",
    "aliases": [
      "Reason"
    ],
    "extensions": [
      ".re",
      ".rei"
    ],
    "configuration": "./language/reason.json"
  },
  {
    "id": "ocaml",
    "aliases": [
      "OCaml"
    ],
    "extensions": [
      ".ml",
      ".mli"
    ],
    "configuration": "./language/ocaml.json"
  }
],
"configuration": {
  "type": "object",
  "title": "Merlin configuration",
  "properties": {
    "merlin.refmt.width": {
      "type": [
    "number",
    null
      ],
      "default": null,
      "description": "Set the width of lines when formatting code with refmt"
    }
  }
},
"grammars": [
  {
    "language": "ocaml",
    "scopeName": "source.ocaml",
    "path": "./syntax/ocaml.json"
  },
  {
    "language": "reason",
    "scopeName": "source.reason",
    "path": "./syntax/reason.json"
  }
]
  }
}
```

### Re-enable Reason syntax on opam tests

``` javascript
// TODO: re-enable 
// assert.equal(
//   reasonDocument.languageId,
//   "reason",
//   "Must be identified as a Reason document"
// );
```

### Fix esy build on Linux

``` yaml
  - script: "esy build"
displayName: "esy build"
continueOnError: true
```

### Checkpoints to Merge

- [x] Remove all generated code
- [x] Remove tablecloth
- [x] Replace all reason with OCaml (can be done automatically with refmt)
- [ ] Ocamlformat the entire codebase.


prometheansacrifice:
The Azure Pipelines setup is in. Can we run them before we merge? Locally tests ran fine.

rgrinberg:
> The Azure Pipelines setup is in. Can we run them before we merge? Locally tests ran fine.

What does this require?

prometheansacrifice:
We'll need an Azure Pipelines account, create a pipeline there and install the Azure Pipelines app on this repo  in the gh settings

rgrinberg:
In the interest of maintenance, can we port all the reason code (apart from tests) to OCaml? I don't think we should require contributors to learn two languages. It doesn't have to be in this PR, but this is a requirement I'd like to keep for core projects.

prometheansacrifice:
Sure, I'm onboard. I hope it's okay to revisit this later.

rgrinberg:
@prometheansacrifice could you resolve the conflict in package.json?

prometheansacrifice:
TODO: Get back icon
TODO: implement 'esy status' inside the plugin so that opam users are not prompted for esy

.ci/build-platform.yml
```diff
@@ -0,0 +1,127 @@
...
+      - script: "npm run package"
+        displayName: "Packaging the extension"
+      - template: utils/use-cache-vscode-test.yml
+      - script: "npm test"
+        displayName: "Running vscode integration tests"
+        env:
+          DISPLAY: ':99.0'
+      - template: utils/publish-esy-cache.yml
+
+      - bash: 'mv vscode-merlin-0.1.0.vsix vscode-merlin-$(git rev-parse --short HEAD).vsix'
```
rgrinberg: Is this right? vscode-merlin is the old plugin isn't it?

Also, why are the versions hard coded in here?

azure-pipelines.yaml
```diff
@@ -0,0 +1,51 @@
+name: vscode-merlin
```
rgrinberg: The name should be updated

bsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "vscode-merlin",
```
rgrinberg: Needs to be updated

fixtures/sample-esy/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "test": "esy x RunSampleEsyTests.exe"
+  },
+  "dependencies": {
+    "@opam/dune": "*",
+    "@esy-ocaml/reason": "*",
+    "ocaml": "4.6.x"
+  },
+  "devDependencies": {
+    "@opam/merlin": "*",
+    "@opam/ocaml-lsp-server": "prometheansacrifice/ocaml-lsp:ocaml-lsp-server.opam#924ac293be5d"
```
rgrinberg: Is this fork still needed?

fixtures/sample-esy/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "start": "esy x SampleEsyApp.exe",
+    "test": "esy x RunSampleEsyTests.exe"
+  },
+  "dependencies": {
+    "@opam/dune": "*",
+    "@esy-ocaml/reason": "*",
+    "ocaml": "4.6.x"
+  },
+  "devDependencies": {
+    "@opam/merlin": "*",
```
rgrinberg: is merlin really required? ocaml-lsp-server should include it.

package.json
```diff
@@ -21,11 +21,19 @@
   "activationEvents": [
     "onLanguage:ocaml"
   ],
-  "main": "./out/client/extension",
+  "main": "./dist/extension.js",
   "contributes": {
     "configuration": {
       "title": "OCaml configuration",
       "properties": {
+        "merlin.refmt.width": {
```
rgrinberg: The prefix shouldn't be merlin here.

fixtures/sample-esy/package.json
prometheansacrifice: Not needed. But it shouldn't affect since it's a test project. Removing it so that tests run faster

fixtures/sample-esy/package.json
prometheansacrifice: I'll update once the submodule updating PR lands. Until then its needed

rgrinberg:
> I'll update once the submodule updating PR lands. Until then its needed
Which PR is that?

prometheansacrifice:
This PR I just raised: https://github.com/ocaml/ocaml-lsp/pull/90

imbsky:
I'm not really a fan of automatic toolchain setup. and I don't want to install such a tool. Is there a problem with marking the necessary tools as prerequisites?

imbsky:
How about making a branch called `vscode-merlin` and polishing it there? I'm satisfied with this extension in the current master and would like to be a little more cautious about merging. The diff is too big.

imbsky:
The current codebase is really small and I want to rewrite from scratch with OCaml. Also, installing a toolchain should not be done by extensions. If you want to do that, you should do it with your own extensions. Again, I don't want to use such a extension and it should be discussed carefully.

rusty-key:
> I'm not really a fan of automatic toolchain setup. and I don't want to install such a tool. Is there a problem with marking the necessary tools as prerequisites?

Automatic setup is happening only for BS and only because there is no good way to get required binaries in JS world. 

Also, while you are not a fan of this kind of setup, I am. One of the obstacles on the way of making Reason more popular is how hard it is to get a working extension. Think about all of the steps you need to do to make current extensions work: 
— First, you need to understand what you need to install. Not only that, you need to know the version of your compiler, otherwise nothing will work and you most likely will not get any meaningful errors. This is especially hard for people coming from JS world. 
— Then you need to get all of the binaries paths (which might be not trivial with tools like `esy`).
— Then you have to provide them to your extension's configuration.
— And you need to do all of that for each project you are working on.

Extension that works out-of-the-box would be a life saver.

While all of the dev-dependencies are not included into one package (`bs-platform`, for example), I would root for taking care of it with extension. Though, it can be optional or thinner. For example, instead of installing everything automatically, it can suggest a configuration file that can be used for installing dependencies.

prometheansacrifice:
> For example, instead of installing everything automatically, it can suggest a configuration file that can be used for installing dependencies.

This is very helpful. A configuration file sounds like a great middle ground

rgrinberg:
My preference is towards automatic setup as well whenever it's non-intrusive. I don't have any suggestions for how to handle the bucklescript/reason workflow because I don't have experience with it. I will let @prometheansacrifice and @rusty-key decide what works for this subset of users.

Something that I want to clarify however is that I am against:
* Packaging any kind of pre-built binaries with the extension. This bloats the extension and forces us to make releases just to update the binary.
* Intrusive automatic setup. The tool should suggest assistance only when it knows the user should take it.
* Eager downloading of packages of any kind. This should be done as lazily as possible and we should be stingy with the user's bandwidth.

As far as I understand, this PR does not violate any of the points above (except perhaps the 2nd point which is why it's WIP) so I'm broadly in favor.

prometheansacrifice:
TODO: Persist user choice to workspace settings

prometheansacrifice:
The TODO is a minor task. Feel free to take a look at the new setup. Relevant files are Toolchain.re and Extension.re

imbsky:
> For example, instead of installing everything automatically, it can suggest a configuration file that can be used for installing dependencies.

This idea is great. or it would be great if it could disable auto install by default and enable it in user preferences.

imbsky:
Solutions like texlab-vscode look good to me.

.ci/utils/use-esy.yml
```diff
@@ -0,0 +1,6 @@
+# steps to install esy globally
+
+steps:
+  # Ensure esy is able to pull in submodules (introduced in 0.6.0)
+- script: 'npm install -g esy --unsafe-perm'
```
rgrinberg: Would it be possible to standardize and just use yarn?

src/Toolchain.re
```diff
@@ -0,0 +1,759 @@
...
+
+  module PackageManagerSpecTuple: {
+    type t = (a, Utils.Fpath.t);
+    let compare: (('a, 'b), ('a, 'b)) => int;
+  };
+  module PackageManagerSpecTupleSet: {
+    type elt = PackageManagerSpecTuple.t;
+    type t = Set.Make(PackageManagerSpecTuple).t;
+    let empty: t;
+    let is_empty: t => bool;
```
rgrinberg: Why do we list out the entire signature of Set? Can’t we use an include?

src/tablecloth/.ocamlformat-ignore
```diff
@@ -0,0 +1 @@
+bs/src/compat.ml
```
rgrinberg: Can we move censored code to its own directory so that people don’t edit it by accident?

src/bindings/Node.re
```diff
@@ -0,0 +1,320 @@
+[@bs.module "./fs-stub.js"]
+external thisProjectsEsyJson: string = "thisProjectsEsyJson";
```
rgrinberg: Are these bindings also part of the original project or are they taken from somewhere?

As mentioned earlier, we should clearly isolate vendored code

rgrinberg:
Why do we check in the generated js? Isn’t it sufficient to just make sure the published plugins contain the js?

.ci/utils/use-esy.yml
prometheansacrifice: I'll make the change

prometheansacrifice: Can we ignore this for now? I don't want to touch the CI setup right now

src/Toolchain.re
prometheansacrifice: Sure!

src/tablecloth/.ocamlformat-ignore
prometheansacrifice: It had to be vendored because of a bug in Tablecloth. We'll get rid of tablecloth in a future PR.

src/bindings/Node.re
prometheansacrifice: These bindings are not taken from anywhere. If anything, they could be contributed to [bs-vscode](https://github.com/glennsl/bs-vscode) (or anywhere) since they are very useful

prometheansacrifice:
> Why do we need tablecloth? Why can’t we just use belt directly?

We dont strictly need. I'm willing to remove it. Maybe in another PR?

> Why do we check in the generated js? Isn’t it sufficient to just make sure the published plugins contain the js?

It's just general practice. We dont really need to check in generated JS. I'll make the change

fixtures/sample-esy/package.json
rgrinberg: I believe that PR landed

src/Error.re
```diff
@@ -0,0 +1,2 @@
+type t =
+  | Foo;
```
rgrinberg: Do we need this source?

src/Extension.re
```diff
@@ -0,0 +1,44 @@
...
+  f =>
+    Js.Promise.then_(
+      fun
+      | Ok () => Js.Promise.resolve()
+      | Error(msg) => f(msg),
+    );
+
+let activate = _context => {
+  Js.Dict.set(Process.env, "OCAMLRUNPARAM", "b");
+  Js.Dict.set(Process.env, "MERLIN_LOG", "-");
```
rgrinberg: I believe this variable has been renamed to something ocaml-lsp specific..

src/Toolchain.re
```diff
@@ -0,0 +1,759 @@
...
+    |> Js.Promise.then_(
+         fun
+         | Ok([]) =>
+           PackageManager.Manifest.lookup(projectRoot)
+           |> okThen(x =>
+                x
+                |> PackageManager.PackageManagerSpecTupleSet.elements
+                |> (
+                  fun
+                  | [] => Error("TODO: global toolchain")
```
rgrinberg: What is this todo for?

src/tablecloth/.circleci/config.yml
```diff
@@ -0,0 +1,75 @@
...
+    environment:
+      CI: true
+    working_directory: ~/repo
+    steps:
+      - checkout:
+          path: ~/repo
+      # m4 is a system dependency required by conf-m4 -> ocamlfind -> fmt -> alcotest
+      # fd (aka fdfind) is needed to support formatting. It will respect the 's .gitignore
+      - run: sudo apt-get install -y m4 fd-find
+      - run: opam update && opam install 'ocamlformat=0.10' -y
```
rgrinberg: Should be updated to a recent ocamlformat (or at least leave a TODO) here.

src/bindings/Node.re
rgrinberg: I see. Things can stay as is for now. I just want to make sure we aren't accidentally maintaining forks of anything.

.ci/utils/use-esy.yml
rgrinberg: Okay. Let's leave a TODO in here however (and/or an issue)

rgrinberg:
The additions this PR makes are quite valuable and I'd love to have them as soon as possible. However, I do feel like we need to put some more work into some yak shaves before we can merge. For me to accept this work, I need to be reasonably confident that I can maintain it. I already have to commit to learning quite a few new technologies to develop this plugin, so I think it would be best to get rid of the stuff that isn't essential. I've added a check list of things that need to be done before I can merge this PR.

I don't expect you to work on all of these points, so let me know if there are any thing you'd like the other maintainers to help you with.

I'll give this PR some more serious testing in the upcoming to make sure there are no usability regressions.

src/command/Unzip.re
```diff
@@ -0,0 +1,10 @@
+open Bindings;
+
+let run = (~p=?, file) =>
+  ChildProcess.exec(
+    {j|unzip $file|j},
```
rgrinberg: Can we always assume this binary is present?

src/Toolchain.re
```diff
@@ -0,0 +1,759 @@
...
+            })
+         << Array.filter(~f=((_pm, available)) => available),
+       );
+  };
+
+  let setupToolChain = spec => spec.setup();
+  let lsp = spec => spec.lsp();
+  let env = spec => spec.env();
+
+  module Manifest: {
```
rgrinberg: Some comments explaining the terminology would be quite useful to me. I'm not entirely sure what is a manifest in this context for example

src/bindings/Rimraf.re
```diff
@@ -0,0 +1,16 @@
+module Rimraf = {
```
rgrinberg: This module could use a comment (and an mli)

src/command/Unzip.re
prometheansacrifice: No, we cant. Thanks for catching this. Ideally, the entire Setup.re could use `module Cmd` (from Toolchain.re), but in the interest of time, I have kept it as it is

prometheansacrifice: In practice, however, we have found `unzip` on all three platforms

src/Error.re
prometheansacrifice: No, we don't. Removing it

src/command/Unzip.re
rgrinberg: Seems fine with me. If we are confident that this binary is available, then it isn't a problem.k

src/AzurePipelines.ml
```diff
@@ -0,0 +1,168 @@
...
+                                ((Error
+                                    (((E.InvalidJSONType ("id"))
+                                      [@explicit_arity ])))
+                                [@explicit_arity ]))
+                       | _ -> ((Error (E.InvalidFirstArrayElement))
+                           [@explicit_arity ]))
+                  | _ ->
+                      ((Error
+                          (((E.MissingField ("value"))[@explicit_arity ])))
+                      [@explicit_arity ])))
```
rgrinberg: Is there a way to get rid of these `explicit_arity` annotations?

prometheansacrifice:
I'm not sure, tbh

src/Toolchain.ml
```diff
@@ -0,0 +1,615 @@
...
+end
+
+module R = Result
+open Bindings
+open Utils
+
+let pathMissingFromEnv = "'PATH' variable not found in the environment"
+
+let noPackageManagerFound =
+  {j| No package manager found. We support opam (https://opam.ocaml.org/) and esy (https://esy.sh/) |j}
```
rgrinberg: This error message appears a bit too often. If no package manager found, we should just gracefully downgrade the functionality we provide. There's quite a few OCaml projects out there that use neither opam nor esy, so we shouldn't annoy them with this message.

Unless there's a less annoying way to report this message, I would suggest that we remove it.

prometheansacrifice: I didn't realise that there are a significant number of projects out there using neither opam nor esy. Can we deal with this in a future PR? Until then I think we should explicitly show this message

src/bindings/Node.re
rgrinberg: Should we port the bindings to OCaml as well then?

rgrinberg:
I got rid of all the `explicit_arity` attributes myself.

src/Toolchain.ml
rgrinberg: As discussed on slack, I'm afraid we can't delay this because it impacts the user experience for a large chunk of users.

rgrinberg:
I've had a close look at Toolchain.ml and here are a few general remarks:

* The use of first class modules everywhere seems gratuitous. I don't see why a type like this is all we need:

```ocaml
type toolchain =
  | Opam_switch of { name : string }
  | Opam_local of { path : Fpath.t }
  | Esy of { package_file : Fpath.t }
  | Global
```

* It seems like if the plugin discovers more than one toolchain, it offers the user a choice. But there's no way to remember this choice. I beileve this needs to be added, otherwise the menu will quickly get annoying.

prometheansacrifice:
I believe I mentioned that we need to persist the user's choice somewhere
here in the thread

prometheansacrifice:
Re: the choice of types

It probably evolved this way and can definitely be improved. Do we have to
deal with it right away?

prometheansacrifice:
Re: choice of types

I'm open suggestions

rgrinberg:
> Do we have to deal with it right away?

I'm just curious, what's the rush to get this merged? I would certainly prefer if the state in plugin was easy to use and understand for all contributors. I believe the changes I'm proposing here would definitely improve things in this department. Is there some other work you have this PR is holding up?

prometheansacrifice:
Quite a few actually. Starting with Windows support. Fixes made to
vscode-merlin since the merger started. Debugger integration. I wanted to
work on these in future PRs after this one is merged

prometheansacrifice:
It's not urgent as such, but that this PR is getting large - I'm trying to
make sure any change needed in this PR is absolutely needed for the
maintainers (ie you) in this PR itself.

rgrinberg:
Alright, I will only make essential changes before merging this.

prometheansacrifice:
Tangential note: one of reports (at vscode-merlin) noted that some use the
extension with monorepos - eg Yarn. Vscode support for multiroot workspaces
might help. I'm guessing there are other mono repo managements tools out
there too.

We might need to prioritize multiroot workspaces too. Speaking of which,
I'd love to learn/survey how large companies/codebases are going to use
this plugin.

rgrinberg:
I've been slow to merge this PR because I'm apprehensive about the future of this code in the plugin. To be frank, I barely have any spare time to commit to refactoring this code. At the same time, the state of the code today is one where it needs some desperate changes. The code is written in a convoluted and non-idiomatic style from the point of view of an OCaml programmer, and I have a hard time following even the relatively small codebase that is present here.

In summary, I think it would be irresponsible for me to merge the plugin at a state where I'm unable to do simple maintenance work on it. I don't mean to offend anyone and I'm always ready to collaborate, but at the same time, I must work with the free time I have available. Unless there is more effort to get this PR into better shape, I will not merge it.

prometheansacrifice:
@rgrinberg

Would you like to help me and others turn this into idiomatic OCaml? We
start we specifically why you feel it not idiomatic or why you are having
trouble understanding the PR.

src/command/unzip.ml
```diff
@@ -0,0 +1,7 @@
+open Bindings
+let run ?p  file =
+  ChildProcess.exec {j|unzip $file|j}
+    (match p with
+     | ((Some (root))[@explicit_arity ]) ->
+         ChildProcess.Options.make ~cwd:root ()
+     | None -> ChildProcess.Options.make ())
```
rgrinberg: What are all these pointless annotations and parens doing here?

src/command/esy.ml
```diff
@@ -0,0 +1,147 @@
...
+         if exitCode = 0 then
+           resolve (Ok stdout)
+         else
+           resolve (Error (E.NonZeroExit (cmd, exitCode, stdout, stderr))))
+
+type subcommand = p:string -> (string, E.t) result Js.Promise.t
+
+let install = (subcommand "install" : subcommand)
+
+let i = (subcommand "i" : subcommand)
```
rgrinberg: This is an incredibly unidiomatic way of writing type anotations

src/command/esy.ml
```diff
@@ -0,0 +1,147 @@
...
+  ChildProcess.exec cmd (ChildProcess.Options.make ())
+  |> then_ (function
+       | Error _e -> resolve (Error (E.CmdFailed cmd))
+       | Ok (exitCode, stdout, stderr) ->
+         if exitCode = 0 then
+           resolve (Ok stdout)
+         else
+           resolve (Error (E.NonZeroExit (cmd, exitCode, stdout, stderr))))
+
+type subcommand = p:string -> (string, E.t) result Js.Promise.t
```
rgrinberg: `p` is a pointless label. The label is supposed to make it easier to understand the meaning of the argument, this is not achieved here.

rgrinberg:
I’ll add some more comments as to why. But really, it should be enough for you to read the discussion in this github thread. I feel that i’ve pointed out quite a few places where the code is obtuse.

My impression is that you’re turning a blind eye towards the issues that I’m pointing out. If that’s the case, my efforts at pointing out flaws in the code aren’t well spent.

src/bindings/Node.re
```diff
@@ -0,0 +1,320 @@
...
+    onError'(t, "error", cb);
+  };
+  [@bs.send] external onEnd': (t, string, unit => unit) => unit = "on";
+  let onEnd = (t, cb) => {
+    onEnd'(t, "end", cb);
+  };
+  [@bs.send] external pipe: (t, Stream.t) => unit = "pipe";
+};
+
+module Https = {
```
rgrinberg: Reason code needs to  be removed or ported

src/bindings/Node.re
```diff
@@ -0,0 +1,320 @@
...
+        |> then_(doesExist =>
+             if (doesExist) {
+               resolve(Ok());
+             } else {
+               let homePath = Sys.getenv(Sys.unix ? "HOME" : "USERPROFILE");
+               if (path == homePath) {
+                 resolve(Error(E.PathNotFound));
+               } else {
+                 mkdir(~p=true, Filename.dirname(path))
+                 |> then_(
```
rgrinberg: Do you think this code is easy to understand? There's like 6 levels of nested if statements. I certainly don't write code like this.

src/Toolchain.ml
```diff
@@ -0,0 +1,647 @@
...
+    val lockFile : Fpath.t
+
+    val make :
+         env:string Js.Dict.t
+      -> root:Fpath.t
+      -> discoveredManifestPath:Fpath.t
+      -> (spec, string) result Js.Promise.t
+  end
+
+  module Esy : T
```
rgrinberg: This use of modules is completely unnecessary.

src/AzurePipelines.ml
```diff
@@ -0,0 +1,118 @@
...
+    | MissingField k -> {j| Response from Azure did not contain build $k |j}
+    | InvalidFirstArrayElement -> "Unexpected array value in Azure response"
+    | Failure url -> {j| Failed to download $url |j}
+end
+
+module JSONResponse = struct
+  module CamlArray = Array
+  open Js
+
+  let getBuildId responseText =
```
rgrinberg: Do you think this function is easy to understand?

src/command/__tests__/Esy_tests.re
```diff
@@ -0,0 +1,26 @@
...
+let projPath = Filename.dirname(__dirname);
+describe("Esy.status", () => {
+  Expect.(
+    testPromise(
+      "Checking if running esy status in __dirname works: " ++ projPath, () => {
+      Js.Promise.(
+        Esy.status(projPath)
+        |> then_(
+             resolve
+             << (
```
rgrinberg: Let's not introduce operators borrowed from other programming languages.

src/Toolchain.ml
```diff
@@ -0,0 +1,647 @@
...
+           | Ok x -> x
+           | Error msg ->
+             Js.log2 "Error during alreadyUsedPackageManagers()" msg;
+             []
+         in
+         match
+           List.filter
+             (fun (x, _) ->
+               match
+                 List.find_opt (fun y -> y = x) availablePackageManagers
```
rgrinberg: Is this just an obfuscated version of `List.exists`?

src/Toolchain.ml
```diff
@@ -0,0 +1,647 @@
...
+  let setupToolChain spec = spec.setup ()
+
+  let lsp spec = spec.lsp ()
+
+  let env spec = spec.env ()
+
+  let globalSpec env =
+    { setup = (fun () -> Js.Promise.resolve (Ok ()))
+    ; lsp = (fun () -> ("ocamllsp", [||]))
+    ; env = (fun () -> env |> R.return |> Js.Promise.resolve)
```
rgrinberg: The use of callbacks here is completely unjustified. As far as I can tell, there are basically 3 possible values for all these fields. Why do we need to pass functions then?

package.json
```diff
@@ -25,10 +25,29 @@
...
+          "description": "Determines what package manager to use for a given project"
+        },
+        "ocaml.refmt.width": {
+          "type": [
+            "number",
+            null
+          ],
+          "default": null,
+          "description": "Set the width of lines when formatting code with refmt"
+        },
```
imbsky: I don't think we need this.

imbsky:
Did you copy and paste CI files from other projects? You shouldn't write a large script directly to yaml.

src/AzurePipelines.ml
prometheansacrifice: Ported the JSON parsing logic to bs-json. This should make it easier to read.

src/Toolchain.ml
prometheansacrifice: This should have been a set intersection operation. Fixed it

package.json
prometheansacrifice: You are right. Thanks for catching this. This has been removed

src/Toolchain.re
prometheansacrifice: This TODO is for cases when an open file/buffer does not have any associated .merlin files since it is not a part of any project. We can reload on globally installed toolchain, but this hasn't been tested or tried

src/tablecloth/.circleci/config.yml
prometheansacrifice: Tablecloth has been removed

src/Toolchain.ml
prometheansacrifice: Fixed with `Global` variant introduced that looks for globally installed `ocamllsp`

src/bindings/Rimraf.re
prometheansacrifice: Added a comment and signature

prometheansacrifice:
@rgrinberg Refactored so that Toolchain.ml relies less on unnecessary modules as you suggested. Tried to address individual requests too. 

This should make the code more approachable. Still open to feedback! Feel free to let me know how things can be improved

prometheansacrifice:
Will shortly add more notes and comments in Toolchain.ml

rusty-key:
Looks like force-pushing messes with GH interface and makes some of the conversations hard to access. I am not completely sure that it is the reason, but if it is, it's better to merge master rather then rebase on it and force-push.
Screen Shot 2020-03-19 at 8 44 09 PM

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+let init ~env ~folder =
+  let projectRoot = Fpath.ofString folder in
+  Js.Promise.all2
+    ( PackageManager.available ~root:projectRoot ~env
+    , Manifest.lookup projectRoot
+      |> okThen (fun pms ->
+             if pms = PackageManagerSet.empty then
+               Error "TODO: global toolchain"
+             else
+               Ok pms) )
```
rusty-key: Let's extract this part to a function so `Js.Promise.all2` is easier to read? 
```
let lookupNotEmpty projectRoot = 
  Manifest.lookup projectRoot
  |> okThen (fun packageMangerSet ->
      if packageMangerSet = PackageManagerSet.empty then
        Error "TODO: global toolchain"
      else
        Ok packageMangerSet)

...
Js.Promise.all2 
  ( PackageManager.available ~root:projectRoot ~env
  , Manifest.lookupNotEmpty projectRoot)
  |> Js.Promise.then_ ...
```

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+  |> Js.Promise.then_
+       (fun (availablePackageManagers, alreadyUsedPackageManagers) ->
+         let availablePackageManagers =
+           match availablePackageManagers with
+           | Ok x -> x |> PackageManagerSet.of_list
+           | Error msg ->
+             Js.log2 "Error during availablePackageManagers()" msg;
```
rusty-key: These messages are outputted to the developer console, right? Let's use appropriate levels then? `Js.Console.error` here and below, for example.

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+             PackageManagerSet.empty
+         in
+         Js.log "availablePackageManagers";
+         PackageManagerSet.iter
+           (fun x -> x |> PackageManager.toString |> Js.log)
+           availablePackageManagers;
+         Js.log "possiblyUsed";
+         PackageManagerSet.iter
+           (fun x -> x |> PackageManager.toString |> Js.log)
+           alreadyUsedPackageManagers;
```
rusty-key: Looks like you are doing the same stuff for both results. Maybe it makes sense to return the same type from both:
`PackageManager.available` and `Manifest.lookup` and extract the similar stuff above into a function?

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+         | [] -> (
+           Js.log "Will lookup toolchain from global env";
+           match PackageManager.ofName projectRoot "<global>" with
+           | Ok t -> PackageManager.make ~env ~t
+           | Error msg -> Error msg |> Js.Promise.resolve )
+         | [ obviousChoice ] ->
+           Js.log2 "Toolchain detected" (PackageManager.toString obviousChoice);
+           PackageManager.make ~env ~t:obviousChoice
+         | multipleChoices -> (
+           let config = Vscode.Workspace.getConfiguration "ocaml" in
```
rusty-key: To make this function more concise, I'd extract everything related to choosing one of the `PackageManagerSet.t` out of many to a separate function (basically, most of the lines below)

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+           match
+             ( config
+               |. Vscode.WorkspaceConfiguration.get "packageManager"
+               |> Js.Nullable.toOption
```
rusty-key: Why not make `Vscode.WorkspaceConfiguration.get` return `option`?

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+           | _ ->
+             (Window.showQuickPick
+                ( multipleChoices
+                |> List.map (fun pm -> PackageManager.toName pm)
+                |> Array.of_list )
+                (Window.QuickPickOptions.make ~canPickMany:false
+                   ~placeHolder:
+                     "Which package manager would you like to manage the \
+                      toolchain?"
+                   ()) [@bs])
```
rusty-key: Maybe I am missing something, but why do we need `[@bs]` attribute here?

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+             |> Js.Promise.then_ (fun packageManager ->
+                    match Js.Nullable.toOption packageManager with
```
rusty-key: Again, maybe it make sense to return `option` earlier?

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+                    | None ->
+                      Js.Promise.resolve
+                        (Error "showQuickPick() returned undefined")
+                    | Some packageManager ->
+                      Vscode.WorkspaceConfiguration.update config
+                        "packageManager" packageManager 2
```
rusty-key: what's `2` here?

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+  type spec =
+    { cmd : Cmd.t
+    ; t : t
+    }
+
+  let esy root = Esy root
+
+  let opam root = Opam root
+
+  let make ~env ~t =
```
rusty-key: 1. I'd expect from `make` function to produce `t`. Instead it takes `t` and does something with it. I think this function needs a better name. 
2. `t` is a very cryptic label for an argument. As @rgrinberg mentioned earlier, labels should explain arguments, here it doesn't help. But I think with the better function name you will not need label at all. 

`t` is kinda cryptic. We either need a more descriptive name, or no name at all if the

rusty-key:
btw, can we consider increasing line width? Code looks "too vertical" :)
It's super-opinionated though, feel free to ignore

rgrinberg:
Let’s stick to 80 (78 in fact). I use 3 way diffs regularly in my workflow.

Also, if your code is nesting that far, there’s something wrong with your code :)

Thanks for the helpful review otherwise. I agree with all of your suggestions. I think I added you as a contributor, so you should be able to push fix (id you’d like) as well.

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+module R = Result
+
+type commandAndArgs = string * string array
+
+let promptSetup ~f =
+  let open Js.Promise in
+  (Window.showQuickPick [| "yes"; "no" |]
+     (Window.QuickPickOptions.make ~canPickMany:false
+        ~placeHolder:{j|Setup this project's toolchain with 'esy'?|j} ()) [@bs])
+  |> Js.Promise.then_ (fun choice ->
```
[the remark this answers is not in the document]
prometheansacrifice: Sure! @rgrinberg What do you think of aliasing modules like this?

src/Toolchain.ml
```diff
@@ -0,0 +1,500 @@
...
+         | Some choice ->
+           if choice = "yes" then
+             f ()
+           else
+             resolve (Error "Please setup the toolchain"))
+
+let setupWithProgressIndicator esyCmd ~envWithUnzip:esyEnv folder =
+  let open Setup.Bsb in
+  Window.withProgress
+    [%bs.obj { location = 15; title = "Setting up toolchain..." }]
```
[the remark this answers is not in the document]
prometheansacrifice: Those are enumerated values. I had some issue while I was setup it up. In the interest of making progress I left those values numbers. Ideally, they should be [variants compiled to integers](https://bucklescript.github.io/docs/en/generate-converters-accessors#convert-between-js-integer-enum-and-bs-variant)

https://code.visualstudio.com/api/references/vscode-api#ProgressLocation
https://code.visualstudio.com/api/references/vscode-api#ConfigurationTarget

I'll try this again today

src/Toolchain.ml
prometheansacrifice: At times, debugging got hard and readability of the output helped. Adding [@bs] removed the currying wrapper and made the output more readable

src/Toolchain.ml
prometheansacrifice: An enumerated value as explained earlier. I'll try turning them to variants that can be compiled to ints

src/Toolchain.ml
prometheansacrifice: Renamed this function to `makeSpec`

prometheansacrifice:
Thanks @rusty-key I think I have addressed your review comments. Feel free to let know if I can improve things further

src/Toolchain.ml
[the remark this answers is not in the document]
rusty-key: If there is a challenge to turn them into variants, you can just explain them (with comment or by extracting them to a variable with a good name)

src/AzurePipelines.ml
```diff
@@ -0,0 +1,102 @@
...
+    try
+      match Json.parse responseText with
+      | Some json ->
+        let downloadUrlObject json =
+          { downloadUrl = field "downloadUrl" string json }
+        in
+        let resource = field "resource" downloadUrlObject json in
+        Ok resource.downloadUrl
+      | None -> Error "getDownloadURL(): could not parse responseText"
+    with _ ->
```
rgrinberg: Can we catch specific exceptions here? Catch alls lead to hard to catch bugs

src/Cmd.ml
```diff
@@ -0,0 +1,64 @@
...
+           Js.String.split env_sep path
+           |> Js.Array.map (fun p -> Filename.concat p cmd))
+    |> Js.Array.reduce Js.Array.concat [||]
+    |> Js.Array.map (fun p ->
+           Fs.exists p
+           |> Js.Promise.then_ (fun exists -> Js.Promise.resolve (p, exists)))
+    |> Js.Promise.all
+    |> Js.Promise.then_ (fun r ->
+           Js.Promise.resolve (Js.Array.filter (fun (_p, exists) -> exists) r))
+    |> Js.Promise.then_ (fun r ->
```
rgrinberg: I think we can merge successive `then_`

src/Setup.ml
```diff
@@ -0,0 +1,215 @@
...
+                  Cmd.output unzip ~args:[| "cache.zip" |] ~cwd:hiddenEsyRoot)
+           |> then_ (function
+                | Error _unzipCmdError ->
+                  Error (E.CacheFailure "Failed to unzip downloaded cache")
+                  |> resolve
+                | Ok _unzipCmdOutput -> resolve (Ok ()))
+         | Error e -> Error e |> resolve)
+    |> then_ (function
+         | Ok () ->
+           reportProgress eventEmitter 96.66;
```
rgrinberg: Where does this number come from?

src/AzurePipelines.ml
```diff
@@ -0,0 +1,102 @@
...
+
+let getBuildID () =
+  Https.getCompleteResponse
+    {j|$restBase/$proj/_apis/build/builds?$filter&$master&$latest&api-version=4.1|j}
+  |> then_ (fun r ->
+         resolve
+           ( match r with
+           | Ok response -> (
+             try RESTResponse.getBuildId response
+             with _ ->
```
rgrinberg: Same issue. Let's not use catch alls unless absolutely necessary.

src/Toolchain.ml
```diff
@@ -0,0 +1,495 @@
...
+    | Opam _, Global -> -1
+    | Global, _ -> 1
+
+  let rec find name = function
+    | [] -> None
+    | x :: xs -> (
+      match x with
+      | Global -> None
+      | Esy root -> (
+        match ofName root name with
```
rgrinberg: In what situations `ofName` can fail?

src/Toolchain.ml
```diff
@@ -0,0 +1,495 @@
...
+    | Esy root1, Esy root2 -> Fpath.compare root1 root2
+    | Opam root1, Opam root2 -> Fpath.compare root1 root2
+    | Global, Global -> 0
+    | Opam _, Esy _ -> -1
+    | Esy _, Opam _ -> 1
+    | Esy _, Global -> -1
+    | Opam _, Global -> -1
+    | Global, _ -> 1
+
+  let rec find name = function
```
rgrinberg: is `List.find_map` available in bucklescript?

src/Setup.ml
```diff
@@ -0,0 +1,215 @@
+open Js.Promise
```
rgrinberg: This module needs an mli

src/Toolchain.ml
rusty-key: I don't think so. You can do something like that with `Belt.List.get` → `Belt.Option.map` or `Belt.List.keepMap` → `Belt.List.get`

rgrinberg: I see. We can use `Belt` or roll our own `find_map`.

rusty-key:
@prometheansacrifice, as we discussed, I made an approach on flattening `Toolchain.ml`. In the end I decided to keep `PackageManager` module, but leave there only stuff directly related to `PackageManager.t`. 

I also started refactoring some functions, but didn't finish with it yet (especially with `Manifest` and naming).

Along the way I made some code-style changes.

Please take a look: https://github.com/ocamllabs/vscode-ocaml-platform/compare/b774dee...75af9f5

prometheansacrifice:
I'm changing the signature of `setup`

```diff
- resources -> (resources, string) result Js.Promise.t
+ resources -> (unit, string) result Js.Promise.t
```
runSetup is effectful and does not touch `resources` (which is immutable). This should prevent incorrect assumptions that `runSetup` produces a different/new `resources`

src/Toolchain.ml
prometheansacrifice: There was some code duplication  - `specOfName` reads better with `ofName` and to make it future proof, as function taking in a string, it felt safer to return an option/result

src/Setup.ml
prometheansacrifice: The progress reporting seen here is not perfect. I was hoping that Azure's API's would set the zip file's size in the headers, but it doesn't. This meant I had to hard code progress as a stopgap measure.

rgrinberg: Okay. Let's leave that as a comment then.

src/Toolchain.ml
```diff
@@ -0,0 +1,376 @@
...
+
+module Binaries = struct
+  let esy = "esy"
+
+  let opam = "opam"
+end
+
+(* Terminology: - PackageManager - represents supported package managers with
+   Global as the fallback - Manifest - abstracts functions handling manifest
+   files of the supported package managers *)
```
rusty-key: Maybe make change `wrap-comments` to `false` in `ocamlformat`? 
Otherwise, it probably should be formatted like: 
```
(* Terminology: *)
(* - PackageManager - represents supported package managers with Global as the
   fallback *)
(* - Manifest - abstracts functions handling manifest files of the supported
   package managers *)
```

prometheansacrifice: Setting it to false didn't change this file (it only touched Extension.mli)

Maybe @rgrinberg can help you find what you are looking for?

rgrinberg: What about using ocamldoc comments: `(** *)` and using literal sections where appropriate.

rgrinberg: I'm also in favor of turning off `wrap-comments`

rusty-key: > Setting it to false didn't change this file (it only touched Extension.mli)

The difference is that `ocamlformat` will not touch line breaks if `wrap-comments` is off. But you have to manually edit comment first.
```ocaml
wrap-comments=true
(* Terminology:
    - PackageManager *)
↓↓↓ ocamlformat ↓↓↓
(* Terminology: - PackageManager *)

wrap-comments=false
(* Terminology:
    - PackageManager *)
↓↓↓ ocamlformat ↓↓↓
(* Terminology:
    - PackageManager *)
```

rusty-key:
@prometheansacrifice, @rgrinberg, I made some changes in `Setup.ml` in order to improve readability, please take a look: https://github.com/ocamllabs/vscode-ocaml-platform/compare/a0edab7...b7d8cb3

I also combed other files in `src` to bring them to a common code style, hope it's ok.

rgrinberg:
Looks good to me.

Instead of thenMap, what about having a Promise module with the standard set of functions (including map)

rusty-key:
> Instead of thenMap, what about having a Promise module with the standard set of functions (including map)

Good idea, added

src/Toolchain.mli
```diff
@@ -0,0 +1,65 @@
+(* Toolchain.ml exposes functions that let us
```
rgrinberg: What about making this an ocamldoc comment?

prometheansacrifice:
@rusty-key I have merged `spec` and `resources` - they dont have to be separate since a lot of `PackageManager` has got merged with `Toolchain` itself

src/Manifest.ml
```diff
@@ -0,0 +1,78 @@
+(* ?? Manifest *)
```
rusty-key: @prometheansacrifice can you add a proper description here? Or better to `mli`?

prometheansacrifice:
I'll add more comments.

Regarding @bs attributes, I had used them for a more readable output. I
don't feel very strongly about them, so let's remove them (I'll do that
tomorrow)

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "FSharp.fsacRuntime": "net"
-}
+  "editor.formatOnSave": true,
+  "ocaml.packageManager": "esy"
```
rgrinberg: is this necessary?

src/Toolchain.ml
```diff
@@ -0,0 +1,386 @@
...
+           | None ->
+             Error ("'esy command-env' returned non-json output: " ^ stdout)
+             |> P.resolve)
+  | Opam root ->
+    Cmd.output cmd ~args:[| "exec"; "env" |] ~cwd:(Fpath.toString root)
+    |> P.mapOk parseOpamEnvOutput
+
+let supportedPackageManagers ~env ~root =
+  let supportedPackageManagers =
+    [ PackageManager.Esy root; Esy Fpath.(root / ".vscode" / "esy"); Opam root ]
```
rusty-key: This path calculated three times. Maybe we can put it to `Utils` or some other shared place?

rgrinberg:
I removed a bunch of `resolve` noise and polymorphic variants.

.vscode/settings.json
rusty-key: I committed `editor.formatOnSave` on accident. I don't know about `ocaml.packageManager`. I'd just ignore `.vscode/settings`, what do you think?

rusty-key:
Removed `Promise` aliasing, some usages of `[@bs]`, added modules descriptions and made a couple of other improvements. 

@rgrinberg, @prometheansacrifice, let's take another look and understand what else is blocking?

rusty-key:
hm, apparently, one thing is dealing `fmt:check` :)
@prometheansacrifice, maybe I did something wrong during merge, can you take a look?

rgrinberg:
> I committed editor.formatOnSave on accident. I don't know about ocaml.packageManager. I'd just ignore .vscode/settings, what do you think?
Sounds good to me. Unless the setting is essential, I would prefer to skip it.

src/Toolchain.ml
prometheansacrifice: Yes, indeed. Thanks for addressing this in the commits too

.vscode/settings.json
prometheansacrifice: Removing the settings.json is fine. I noticed it was already checked in and thought it should updated with relevant ones.

prometheansacrifice:
Turns out it was the removal of [@bs]  attributes that caused the crash. I tried chasing them further (retaining them), but ran out of time. Reverted them for the time being. Maybe we should revisit this later?

prometheansacrifice:
Updated the settings section in the README. Note that `toolchainRoot` section in the settings isn't documented on purpose. Tampering with it is not advised at this point in time.

prometheansacrifice:
Should we add this screen shot as a visual aid?

Screenshot 2020-03-27 at 8 41 22 PM


Maybe be with example values filled?

prometheansacrifice:
These [terminal APIs](https://code.visualstudio.com/api/references/vscode-api#Terminal) look really promising.

One feedback I have received is to make the (esy/bsb) setup more transparent. Those APIs could really help us with that. Will open a detailed issue once we merge this PR

wokalski:
@prometheansacrifice `[@bs]` is often helpful when dealing with externals because without it Bucklescript does a runtime inspection of functions and decides to curry them depending on functions `.length`. I also faced some issues because of it recently.

rgrinberg:
@prometheansacrifice @rusty-key I've replaced parsing the shell output of `$ opam exec env` by parsing the sexp of `$ opam env --sexp`. Please review.

prometheansacrifice:
On Windows the generator command has to be `esy.cmd ocamllex $in` - trying to figure a way. If nothing comes up, let's check in `sexp.ml` and document the generator command outside the build system

rgrinberg:
On windows, why don't we run the entire build in the esy sandbox then? It seems inconsistent to me to run some commands inside the sandbox and others outside.

rgrinberg:
>  If nothing comes up, let's check in sexp.ml and document the generator command outside the build system

Yeah, that seems like the lesser evil. Why is ocamllex not portable on windows in the first place?

rgrinberg:
I'm playing around with the plugin, and I'm seeing errors like this:

```
showQuickPick() returned undefined
```

I don't know how to reproduce this issue exactly but it happens when I open the esy project.

prometheansacrifice:
@rgrinberg 

> On windows, why don't we run the entire build in the esy sandbox then? 

Yes, that's the plan. I think the wording 'outside the build system' must be confusing. I meant outside bsb, but still inside esy. 

If bucklescript doesn't get us a way to conditionally use `esy.cmd` on Windows, we could

```js
const cp = require('child_process');
cp.execSync((process.platform == 'win32' ? 'esy.cmd': 'esy') + 'ocamllex.opt sexp.mll'
```

And modify the build step in package.json
```json
// package.json
  "build": "node runOCamllex.js && bsb -make-world"
```

prometheansacrifice:
> showQuickPick() returned undefined

This must be because you unintentionally click outside the picker? `showQuickPick` returns undefined when none of the options were selected. I remember see a more helpful 'Please install the toolchain' message when I tested locally, wonder if it got lost in the refactoring. Will try that code path

rgrinberg:
There might be a bit of a misunderstanding. I was questioning the following commit: https://github.com/ocamllabs/vscode-ocaml-platform/pull/19/commits/bfbb593e85c194a85401e7dd29fb08caf9947af0

Isn't running `$ esy bsb -make-world` enough to make sure that the correct ocamllex is ran? I still find it dubious that the build system is invoking the package manager. As I've said before, all rules should preferably run in the same environment for consistency.

prometheansacrifice:
For `$ esy bsb -make-world` to work, `bsb` must be installed by `esy`. I'm not sure if globally available `bsb` (say from `npm i -g bs-platform` works reliably everywhere.

Getting `bsb` to build and run inside esy has some challenges. Let me explore both and get back here

rgrinberg:
Okay. Well, in any case we want to make sure that the plugin is buildable without esy on non windows problems.

rgrinberg:
I'm looking at the signature of `showQuickPick`:

```
  val showQuickPick :
    string array -> QuickPickOptions.t -> string option Promise.t
```

Shouldn't undefined be converted to `None` here?

prometheansacrifice:
It does return `None`. You're seeing that message from this line

https://github.com/ocamllabs/vscode-ocaml-platform/blob/1747b332a40f47af2ffd5282d1407f130922ec41/src/Toolchain.ml#L207

which can be reworded to something else

bsconfig.json
```diff
@@ -31,11 +25,5 @@
     "number": "-44-45-20",
     "error": "+101+8"
   },
-  "refmt": 3,
-  "generators": [
```
rgrinberg: What's going to happen when we want to re-generate this file?

rgrinberg:
The last workaround also seems very strange to me. We removed the rule, so how are we going to regenerate sexp.ml?  As far as I see, the error in CI is:

```
2020-03-31T13:00:17.9538284Z ##[section]Starting: Compiling ml files using bucklescript
2020-03-31T13:00:17.9649739Z ==============================================================================
2020-03-31T13:00:17.9650050Z Task         : Command line
2020-03-31T13:00:17.9650397Z Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
2020-03-31T13:00:17.9650690Z Version      : 2.164.0
2020-03-31T13:00:17.9651028Z Author       : Microsoft Corporation
2020-03-31T13:00:17.9651346Z Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
2020-03-31T13:00:17.9651753Z ==============================================================================
2020-03-31T13:00:18.8395486Z Generating script.
2020-03-31T13:00:18.8491353Z Script contents:
2020-03-31T13:00:18.8502286Z esy bsb -make-world
2020-03-31T13:00:18.8834926Z ========================== Starting Command Output ===========================
2020-03-31T13:00:18.9093107Z ##[command]"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "d:\a\_temp\7d753503-f0b5-445f-8740-695aee08ffd6.cmd""
2020-03-31T13:00:19.0136952Z error: Exec format error
2020-03-31T13:00:19.0137337Z   
2020-03-31T13:00:19.0141781Z esy: exiting due to errors above
2020-03-31T13:00:19.1523747Z ##[error]Cmd.exe exited with code '1'.
2020-03-31T13:00:19.1887110Z ##[section]Finishing: Compiling ml files using bucklescript
```

This means that one of the following is buggy on Windows:

* bsb
* bucklescript's ocamllex
* esy

Why do we need any workaround at all on our side? Can't we report this problem upstream first and see if it can be fixed easily first?

prometheansacrifice:
bsb is installed via npm globally. On Windows, npm uses .cmd scripts to run
the binaries. Which means, the command would have to be `esy bsb.cmd`?

Will sexp.mll be updated a lot? I can look into this, but running global
commands (in this case bsb) inside esy is not reliable on Windows - more so
with npm installed binaries.

package.json
```diff
@@ -25,10 +25,21 @@
...
   "contributes": {
     "configuration": {
       "title": "OCaml configuration",
       "properties": {
+        "ocaml.toolChainRoot": {
+          "type": "string",
+          "default": null,
+          "description": "Determines where to find the toolchain for a given project"
+        },
+        "ocaml.packageManager": {
```
giltho: Given the following snippet from `Toolchain.ml`
```ocaml
let ofName ~root = function
    | "opam" -> Some (Opam root)
    | "esy" -> Some (Esy root)
    | "global" -> Some Global
    | _ -> None
```
I would add
```json
"enum": [ "esy", "opam", "global" ]
```
to the `ocaml.packageManager` configuration. Somehow this is not documented on the official vscode website, but `enum` allows to have a dropdown selection in the vscode configuration menu

rgrinberg:
> bsb is installed via npm globally. On Windows, npm uses .cmd scripts to run
the binaries. Which means, the command would have to be `esy bsb.cmd`?

So why are we using esy on windows then? If it doesn't support bucklescript properly, what benefit are we getting out of it?

> Will sexp.mll be updated a lot? I can look into this, but running global
commands (in this case bsb) inside esy is not reliable on Windows - more so
with npm installed binaries.

It will not be updated often, but this problem is something that I imagine many other users will run into (with or without ocamllex). So yeah, I would prefer not to hack around this, especially since esy does not seem like an essential component for this plugin.

prometheansacrifice:
The plugin relies on esy for bucklescript support. I was hoping this was
clear.

rgrinberg:
No, sorry. Could you clarify it to me? Currently I'm using & building this plugin without esy and it's working just fine.

package.json
rgrinberg: I'm not sure why we need both this option and the toolchain root actually. The two options seem redundant to me. What about something like a single option interpreted as follows:

* null - the global toolchain is used
* a file that ends with json - esy is used
* a directory or a raw name will be interpreted as an opam switch

prometheansacrifice: @rgrinberg We need to provide a way to override the choice of package manager/toolchain manager. Dropdown/explicit field in the settings is a great way to do this, IMO, and not redundant.

The reason `toolchainRoot` exists there is not for users to override per se - vscode's settings persistence API requires us to expose the field for users to edit.

prometheansacrifice:
> No, sorry. Could you clarify it to me? Currently I'm using & building this plugin without esy and it's working just fine.

Esy is not essential for building the plugin itself, but is essential for the plugin to work for users who don't have opam installed.

The reason I'm sticking to using esy on the CI is that there is no OCaml toolchain installed on the Windows image on CI globally. `ocamllex` is present in the local sandbox on the otherhand. I'm open to ideas

rgrinberg:
This is the offending line I guess:
```
       | None -> Error "showQuickPick() returned undefined" |> Promise.resolve

```

We shouldn't be showing this kind of error to users. What about just defaulting to Global if the user fails to make a choice.

Also, I can observe that `esy` is being presented twice in the options. That seems wrong to me.

rgrinberg:
I don't understand this logic in `Cmd.make`:

```
    let cmds =
      match Sys.unix with
      | true -> [| cmd |]
      | false -> [| cmd ^ ".exe"; cmd ^ ".cmd" |]
    in
    cmds
```

Why do we need to look for both .exe or .cmd? When is `.cmd` an option for a binary?

rgrinberg:
Also, why do we need `cmd` in `Toolchain.t`? If it's just for optimization, then I'm not sure it's necessary.

prometheansacrifice:
> We shouldn't be showing this kind of error to users. What about just defaulting to Global if the user fails to make a choice.

You are right, we need to re-word the message

> Also, I can observe that esy is being presented twice in the options. That seems wrong to me.
This must be so when you open the plugins codebase and not generally - this happens because there is an esy.json and a package.json. That makes the project both a valid esy project and a bucklescript project. In reality, those options are different - one would setup the project as a bsb project via esy and other as a native esy project.

The misleading option names would be fixed as a part of unifying the two prompts we discussed yesterday

> Why do we need to look for both .exe or .cmd? When is .cmd an option for a binary?

On Windows, when binaries are installed by npm (example esy.exe) npm doesnot place the .exe files on the $PATH. They are found in the node_modules. NPM instead places a wrapper bash script for cygwin and .cmd and .ps1 for cmd shell and powershell respective.


> Also, why do we need cmd in Toolchain.t? If it's just for optimization, then I'm not sure it's necessary.

It is not an optimisation. `Cmd.t` provides guarantee of a command existence statically. This has helped us avoid multiple option checks in the codebase as it happens upfront once after the initial `.make`

prometheansacrifice:
> What about just defaulting to Global if the user fails to make a choice.
We can do this, provide we show a small alert in saying that a global toolchain is in us

rgrinberg:
I've polished this PR a bit and will merge shortly. Resolving merge conflicts periodically was getting to be a hassle. For now, the extension just defaults to the global toolchain. We can start fixing issues in separate PR's as @rusty-key mentioned. @imbsky some dependencies are now perhaps out of date again, sorry for that.

imbsky:
No worries.

PR #20, opened 2020-02-14
Add support for OCamllex syntax highlighting

### LINKED ISSUE #13: Ocamllex syntax highlighting
ocamllex scripts occur often enough that we should ave support for them as well

imbsky:
Closes #13


rgrinberg:
Looks very good so far. Some thing that would be nice to improve:

- [ ] Add a snippet for producing rules inside mll files
- [ ] Enable lsp completion inside the OCaml blocks
- [ ] Make commenting work both inside the OCaml blocks and outside them.

Also, is there a way to add a default template for such files? It would be nice if:

```
{}
rule token = parse
 | _ { assert false }
{}
```

was inserted when creating empty mll files.

imbsky:
Sorry for the late response! My mailbox was filled by many other projects and I was slow to notice... I think I can probably work on this after tomorrow!

rgrinberg:
Merged this manually with a few tweaks

imbsky:
Thanks

PR #23, opened 2020-02-18
Add support for Reason

imbsky:


rgrinberg:
@prometheansacrifice mind having a look?

imbsky:
I prefer to merge this simple change than a large diff. and perhaps writing a VSCode extension in OCaml or Reason isn't reasonable. But I would like to hear many opinions. What do you think?

prometheansacrifice:
Sure, I will. Please give me a few hours.

PR #25, opened 2020-02-19
Add vsce as dev dependency

imbsky:
This allows us to execute the package script.


rgrinberg:
> This allows us to execute the package script.

How do we do that?

imbsky:
In other words, we needed this package in our package script but we somehow forgot to install the library.

imbsky:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/package.json#L226

PR #31, opened 2020-02-19
Change license from MIT to ISC

imbsky:
The copyright year doesn't need to be updated. Also, in the previous Markdown version, the two names looked to be connected due to the format, so I changed them a little.


rgrinberg:
Let's change the license to ISC instead.

imbsky:
Done.

PR #64, opened 2020-03-17
Add CODE_OF_CONDUCT.md

imbsky:


rgrinberg:
I think this is something that is best standardized on the org level. Anyways, we only have 2-3 regular contributors. We can manage things informally for now ;)

imbsky:
Indeed

PR #82, opened 2020-03-25
Add indentationRules to ocaml.json

### LINKED ISSUE #60: Auto-indentation rules
When I write the following, the line with ```type t``` doesn't auto indent when I hit enter after ```struct```.

```ocaml
module X = struct
type t
end
```
Below is what I would expect - is it possible to get this feature? Or is it a problem with my environment?

```ocaml
module X = struct
  type t
end
```

imbsky:
Closes #60

### DIFF
--- a/languages/ocaml.json
+++ b/languages/ocaml.json
@@ -20,5 +20,10 @@
     ["(", ")"],
     ["'", "'"],
     ["\"", "\""]
-  ]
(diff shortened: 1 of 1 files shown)


rgrinberg:
Looks good. I wish there was a way to write these regexes in a cleaner form though.

PR #91, opened 2020-04-01
Add Menhir/OCamlyacc syntax highlighting

mnxn:
Addresses #21 

Parts of the regex were taken from the Menhir tmLanguage file in https://github.com/hackwaly/vscode-ocaml

Some of the interactions between this grammar and the existing OCaml grammar are unexpected. 
For example, everything after `List` will not be highlighted, but adding `;;` after `List` will make it work fine. 

```yacc
%{
open List
%}
%start <Ast.program> program
%%
program: module_body EOF { Program $1 }
```
I've deduced that this is a problem with the existing OCaml syntax and I'm working on a much simpler ocaml grammar which I can confirm would fix this problem and other inconsistencies.

I have not found any highlighting issues with the Menhir/OCamlyacc grammar itself.


rgrinberg:
Looks good to me. @imbsky feel free to merge once you've tested it.

imbsky:
Thanks @mnxn for your contribution!

PR #95, opened 2020-04-03
Tweak OCaml syntax highlighting and snippets

mnxn:
I looked over the OCaml grammar and snippets, improving a few things:
- the lambda snippet includes the `fun` keyword
- unicode escape sequence has between 1 and 6 digits
- an underscore is considered a constant pattern instead of a keyword
- the empty list constructor is highlighted


imbsky:
Thanks @mnxn!

PR #96, opened 2020-04-03
Rewrite OCamllex grammar

mnxn:
This PR fixes some bugs with the ocamllex configuration and grammar. Also, since the OCaml grammar already defines patterns for strings and characters, I made it so the ocamllex and menhir grammars just reference the existing patterns (`source.ocaml#strings` & `source.ocaml#characters`).

- Adds keyword highlighting for `shortest`, `refill`, `as`
- OCaml block comments are used instead of C-style block comments
- Comments are correctly highlighted (dimmed for my theme)
- Curly brackets in string literals don't escape the string and break highlighting
- In the OCaml grammar, character literals are now `constant.character` instead of `string.quoted`

Comparison:
| Old | New |
| - | - |
| old | new |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unrealeased
+
+- Improve ocamllex syntax highlighting
+
 ## 0.4.0
 
 - Add syntax highlighting and basic language support for ocamlyacc/menhir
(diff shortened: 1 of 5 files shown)


imbsky:
The code looks good to me!

rgrinberg:
Looks good to me as well. This change should be included in the change log by the way.

mnxn:
Ready to merge?

rgrinberg:
LGTM

Issue #97, opened 2020-04-03
Polish opam syntax highlighting


imbsky:
The current syntax highlighting is quite awkward.
<img alt="Screen Shot 2020-04-03 at 17 03 13" src="https://user-images.githubusercontent.com/38746192/78338087-189e6280-75cd-11ea-8490-c75afd79f49c.png">

mnxn:
I can take a look tomorrow.

imbsky:
Thanks @mnxn!

PR #101, opened 2020-04-03
Improve opam syntax

mnxn:
Addresses #97

I'm not entirely familiar with the opam file format, so if I missed something, please let me know.

Comparison:

| Old | New |
| - | - |
| old | new |


imbsky:
Looks good to me! Thanks!

rgrinberg:
Looks good to me as well. Just needs a change log entry.

PR #102, opened 2020-04-03
Improve README.md

imbsky:


imbsky:
@rgrinberg The contents haven't changed much, but please check it just in case.

imbsky:
The idea of adding a quick start section came from @Khady's tweet.

PR #107, opened 2020-04-04
Remove Fpath in favor of Path

rgrinberg:
Use one Path module consistently


rgrinberg:
Merging this as it's trivial and is required for more important work

PR #109, opened 2020-04-04
Add a command to select sandbox

rgrinberg:
The goals of this PR are:

- [x] Allow the user to switch sandboxes at anytime with a command
- [x] If the user does not make a switch selection, then no setting is persisted.
- [x] Do not discover "manifests" for opam projects. This concept does not align with opam based workflow. Users should be able to select any switch on their system
- [x] Gentler failure. Previously, the entire workflow would stop if some part failed. Change this to a "best effort" policy with appropriate warnings
- [x] Simplify sandbox configuration to a single setting. User preferences regarding package managers should be a completely separate setting that only effects what we present in the selection
- [ ] Suggest users to attach a sandbox to a project when none is attached.


prometheansacrifice:
@rgrinberg Whenever you are ready, can you please update the Nodejs version so that the CI is unblocked again?

https://github.com/ocamllabs/vscode-ocaml-platform/blob/1b79d3d090acd7cdb5214937d8a1c66714d228ac/.ci/utils/use-node.yml#L7

rgrinberg:
If you know what's the right version of node, would you mind pushing a commit to this branch? I'm not entirely sure what it should be.

rgrinberg:
>  Suggest users to attach a sandbox to a project when none is attached.

I'm trying to address this point, but unfortunately it's not working. I'm unable to create a notification with options for the users to select from.

prometheansacrifice:
@rgrinberg I couldn't find a notification box with a dropdown - should we turn it into an additional step? What's the flow of action you have in  mind?

prometheansacrifice:
Unit tests are failing because `import.ml` includes `Vscode` - if a module includes vscode, it will only be available during e2e tests and not during unit tests.

We can revisit this later if you like.

src/Extension.ml
```diff
@@ -21,21 +21,75 @@ module Server = struct
...
+      Instance.start instance t
+  in
+  let (_ : unit Promise.t) = handleError Window.showErrorMessage setToolchain in
+  ()
+
+let suggestToSetupToolchain instance =
+  let open Promise.O in
+  Vscode.Window.showInformationMessage'
+    "There is no toolchain attached to this project."
+    [ ("select toolchain", ()) ]
```
rgrinberg, line 68: @prometheansacrifice This is what doesn't seem to be working at the moment. The function should suggest to setup a toolchain, but the "select toolchain" option doesn't show up.

rgrinberg:
> Unit tests are failing because import.ml includes Vscode - if a module includes vscode, it will only be available during e2e tests and not during unit tests.

Why is that?

I'm fine with a workaround, but I'd like to understand what's the cause first.

prometheansacrifice:
vscode module is only available when the extension is loaded in a vscode
instance.

rgrinberg:
Can't the bindings just fail at runtime? I don't think our unit tests are calling vscode API methods anyway

prometheansacrifice:
Bindings are not unit tested at the moment. IIUC, `vscode` module is only available inside a running VSCode instance. Unit testing them will involve running vscode's integration test runner such that the APIs are called as separate units. Ideally, we should cover them

rgrinberg:
For now, this work is a strict improvement so I'm merging it. I'll debug the issue with the prompt later.

PR #114, opened 2020-04-05
Add mli to Node bindings

rgrinberg:
Simplify the stream bindings as well by removing an extra functor

### DIFF
--- a/src/bindings/Node.ml
+++ b/src/bindings/Node.ml
@@ -47,24 +47,11 @@ module Buffer = struct
   let ofString = from
 end
 
-module type STREAM = sig
+module Stream = struct
   type t
-
-  val on : t -> string -> (Buffer.t -> unit) -> unit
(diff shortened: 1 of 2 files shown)


rgrinberg:
I just realized that bucklescript has its own node bindings. Shouldn't we be using those?

prometheansacrifice:
If we consider not having Node bindings as a part of this repo - we could contribute them here - https://github.com/sikanhe/reason-node

src/bindings/Node.ml
```diff
@@ -47,25 +47,12 @@ module Buffer = struct
   let ofString = from
 end
 
-module type STREAM = sig
```
prometheansacrifice, line 50: This was functor was handy in implementing a lot of subtype relations in the Stream based APIs in Node.js like I/O, network etc

PR #115, opened 2020-04-05
Improve OCaml syntax highlighting in strings and doc comments

mnxn:
- Printf placeholders (those that begin with `%`) are no longer highlighted in `external` declarations
 - Invalid printf placeholders are no longer highlighted because `%` can appear in other contexts where it is not an error
- OCaml sources are highlighted inside ocamldoc comments
- ocamldoc comments no longer have the `emphasis` tag because ocamldoc comments can have text formatting for italics

Comparison:
| Old | New |
| - | - |
| printf_old | printf_new | 
| doc_old | doc_new | 

Still need to find a way to easily distinguish ocamldoc comments from normal comments at a glance...

### DIFF
--- a/syntaxes/ocaml.json
+++ b/syntaxes/ocaml.json
@@ -49,6 +49,7 @@
               "comment": "embedded ocaml source",
               "begin": "\\[",
               "end": "\\]",
+              "contentName": "source.embedded.ocaml",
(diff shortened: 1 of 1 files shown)


imbsky:
My workstation was completely broken so I reviewed this from GitHub Mobile. So I may have missed something.

PR #131, opened 2020-04-12
readme: avoid confusion between ocaml-lsp and ocaml-lsp-server

tatchi:
At first glance, I thought those two were different things while actually they redirect to the same project.

I guess that using the same name for both might avoid some confusion.


imbsky:
Thanks @tatchi!

PR #132, opened 2020-04-13
Update esy.json

imbsky:


imbsky:
I updated the configuration for OCamlFormat 0.14.0 in a recent PR, but it seems like we still can't use it without some annoying overrides. I think we should revert that PR. What do you think?

prometheansacrifice:
The overrides look good. I noticed `ocamlformat` hasn't been updated here, is it because it was updated in the other PR mentioned? Either ways, this PR is good to be merged

prometheansacrifice:
Feel free to post the issues faced in the other PR - I'd like to understand issues faced

imbsky:
Updated.

PR #134, opened 2020-04-13
Fix support for multi dir workspaces

rgrinberg:
This removes all the reliance on a "projectRoot". This does not make
sense for workspaces with multiple directories.

For now, this creates an awkward situation for esy projects as it means
that only a single LSP server is used for all projects in the workspace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.vsix
 *~
 _esy/
+esy.lock
 dist/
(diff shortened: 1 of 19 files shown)


rgrinberg:
Okay, with this PR the plugin is in a usable but annoying state again.  We need to have a better plan for multi dir workspaces and settings.

PR #135, opened 2020-04-14
Add Makefile

imbsky:


imbsky:
Hmm? Somehow the permissions of some files seem to have been wrong before and that change is included in the commit. Should I divide it into separate commits?

prometheansacrifice:
I think it's fine. Regarding the wrapping parens, I'll leave it up to you
all. I'm fine either way.

package.json
```diff
@@ -280,9 +280,7 @@
     "vscode:prepublish": "yarn run build",
     "package": "vsce package --yarn",
     "fmt:check": "prettier --check .",
-    "fmt:ocaml": "git ls-files '*.ml' '*.mli' | xargs ocamlformat -i",
```
rgrinberg, line 283: Why did you replace `git ls-files` with find?

imbsky: I don't know, but it seems to make more sense to use `git ls-files'. I will update this.

Makefile
```diff
@@ -0,0 +1,12 @@
+OCAML_SRCFILES = $(shell git ls-files "*.ml" "*.mli")
+REASON_SRCFILES = $(shell git ls-files "*.re" "*.rei")
```
rgrinberg, line 2: Why do we have formatting of reason sources? We eliminated all reason sources already.

Makefile
```diff
@@ -0,0 +1,12 @@
...
+
+build:
+	yarn build
+.PHONY: build
+
+fmt:
+	esy ocamlformat --inplace --enable-outside-detected-project $(OCAML_SRCFILES)
+	esy refmt --in-place $(REASON_SRCFILES)
```
rgrinberg, line 10: Actually, I don't use esy so this is making formatting unusable for me. If someone wants to use ocamlformat from esy, they should just run `$ esy make fmt`.

Makefile
imbsky: There are several Reason codes in the fixtures directory. Well, the interface file doesn't exactly exist yet.

Makefile
imbsky: Updated.

Makefile
rgrinberg: That's a good opportunity to convert them to OCaml then.

prometheansacrifice: Fixture are for e2e test. We need to ensure the plugin works with Reason syntax

rgrinberg:
In that case, the script should use the refmt that comes with bucklescript.

rgrinberg:
Could you add `doc-comments-val=after` to the ocamlformat config so the comment style doesn't change?

imbsky:
The `doc-comments-val` option was added in 0.14.0, so I don't think we can use it.

rgrinberg:
I might be missing something - why did you downgrade to 0.13.0?

imbsky:
The minimum version of dune was changed from 0.14.0, but esy can't handle it without a dirty hack. and I didn't have a strong motivation to update to 0.14.0 or above, so I reverted it.

rgrinberg:
Okay. Again, it would have been easier to understand this from the description :)

imbsky:
The version of ocamlformat was downgraded by another PR. But I agree with that.

imbsky:
I use opam for all OCaml projects, so if we need to use 0.14.1 in this project, I don't need any dirty hacks. Is there any reason why we are using esy in this project?

rgrinberg:
We are only using esy to test that using the extension is working with esy. We don't need esy to develop the extension itself. If there are problems installing the appropriate ocamlformat with esy, that should be solved in esy.

imbsky:
Oh, I see.

I don't know how to use a secondary compiler with esy, could you help me? @rusty-key @prometheansacrifice

imbsky:
Rebased to current master. The purpose of this PR is to add a Makefile, so I would like to upgrade ocamlformat in another PR. What do you think? @rgrinberg

rgrinberg:
Okay. We can ask one of our esy experts to update to fix the ocamlformat + esy situation

prometheansacrifice:
Sure!

You just need the resolutions mentioned at 
https://github.com/esy/esy/issues/1062#issuecomment-601378849

Secondary compilers, unlike as in opam, are installed separately in the libroot. So, findlib needs to be told of this location. Those overrides do just that.

imbsky:
@prometheansacrifice Thank you so much!

PR #145, opened 2020-04-16
Update build-platform.yml

imbsky:


rgrinberg:
Looks good. I think we need an issue to remove this duplication of information. It will be quite easy to forget to change this otherwise

imbsky:
Agree.

imbsky:
I made an issue, so I would like to merge it once. As for the test, it doesn't seem to work properly, so I think it needs to be fixed. We can fix the test using JavaScript, or rewrite test using MDX instead.

imbsky:
Well, I'm not familiar with MDX, so I don't know if we can test this with MDX.

PR #147, opened 2020-04-17
Fix LSP command check

rusty-key:
Following up on conversation in Slack. Existing command check was blocking initialization of extension. 

Instead of using external command I decided to just use `--help` argument as it will be hard to check something like `/usr/local/bin/opam exec --switch=[switch-path] -- ocamllsp` otherwise.

### DIFF
--- a/src/Toolchain.ml
+++ b/src/Toolchain.ml
@@ -243,7 +243,7 @@ let getLspCommand (t : PackageManager.t) : Path.t * string array =
   | Global -> (Path.ofString name, [||])
 
 let addLspCheckArg args =
-  Array.append args [| "--help" |]
+  Array.append args [| "--help=plain" |]
 
 let runSetup resources =
   let open Promise.Result.O in
--- a/src/Toolchain.ml
+++ b/src/Toolchain.ml
@@ -242,12 +242,15 @@ let getLspCommand (t : PackageManager.t) : Path.t * string array =
   | Esy (esy, manifest) -> Esy.exec esy ~manifest ~args:[| name |]
   | Global -> (Path.ofString name, [||])
 
+let addLspCheckArg args =
+  Array.append args [| "--help" |]
+
 let runSetup resources =
   let open Promise.Result.O in
   setupToolChain resources
   >>= (fun () ->
         let cmd, args = getLspCommand resources in
-        Cmd.make ~cmd () >>= fun cmd -> Cmd.output cmd ~args)
+        Cmd.make ~cmd () >>= fun cmd -> Cmd.output cmd ~args:(addLspCheckArg args))
   |> Promise.map (function
        | Ok _ -> Ok ()
        | Error msg -> Error {j| Toolchain initialisation failed: $msg |j})


prometheansacrifice:
To check if `ocamllsp` binary is available, running `Cmd.make` is enough. It acts like popular commands like `which` and `where`

prometheansacrifice:
We dont have to any another command subsequently

rusty-key:
@prometheansacrifice this will correctly check only existence of global `ocamllsp`. In case of `opam` and `esy` we are not running `ocamllsp` directly. So `Cmd.make` will check existence of `opam` and `esy` which is not what we need

prometheansacrifice:
@rusty-key `Cmd.make ~cmd: "ocamllsp" ~env: processEnv` will lookup ocamllsp in the specified env.

rusty-key:
I don't understand what are you suggesting. We want to check if after toolchain setup lsp binary exists. This command looks like: 
```
/usr/local/bin/opam exec --switch=[opam-switch] -- ocamllsp
```
or
```
/usr/local/bin/esy -P [esy-switch] ocamllsp
```
or 
```
ocamllsp
```
Most of them cannot be tested by `which` or alternatives.

Are you suggesting to change `getLspCommand` to produce absolute path to the binary so we could use `which` on it?

prometheansacrifice:
I'm trying to say running `Cmd.make` on a given binary name is enough to guarantee it exists in a a given environment

https://github.com/ocamllabs/vscode-ocaml-platform/blob/718aff3cd2deea01cd1178d0045fb5b206ff2256/src/Cmd.ml#L36-L46

Regarding looking up a binary in an opam switch or an esy sandbox, both the tools can provide us a dictionary of the environment.  Here is how it used to be - I think we can go back to how it was

https://github.com/ocamllabs/vscode-ocaml-platform/blob/a344356ee0b9cb4f263f5149a4c845dd981f04db/src/Toolchain.ml#L348-L358

In `runSetup`, where we knew which toolchain manager was supposed to be used for a project, we have a way to know what environment to look into for `ocamllsp`. We used to use `Toolchain.env` to get the environment, and lookup `ocamllsp` in it by simply trying to construct a `Cmd.t` - since `Cmd.make` looks  into the specified environment anyway

prometheansacrifice:
`Cmd` along with being a utility to run commands also provides static guarantee that a command exists on the shell

rusty-key:
Man, I feel stupid, I don't understand your point. We are talking about two different solutions, right? One is to solve current approach, another — use absolute paths to binaries. Is that right? 

If yes, then

1. Use current approach, meaning use package manager (opam or esy) to access binary. In that case I still do not understand how am I supposed to use `Cmd.make`? 
`getLspCommand` returns something like `(opam, [| "exec"; "--switch=...", "--", "ocamllsp" |])`. I do not understand how we can use `Cmd.make` to check if `ocamllsp` exists. Can you show me? If it's simple fix, maybe you can just implement instead of trying to explain it to me? 😅 

2. Change the approach and use absolute paths to binaries. While I understand that, it feels like something out of the scope of this PR. Also, before doing that we need to figure out why it was changed from absolute paths to using PMs in the first place? cc @rgrinberg

prometheansacrifice:
Sorry about the disconnect.

> We are talking about two different solutions, right? One is to solve current approach, another — use absolute paths to binaries. Is that right?

If I understand correctly, you are trying to see if there is a usable `ocamllsp`, right? And you are doing so, by running it with `--help` to see if it runs fine.

This execution of the command is unnecessary - it ties us to whether the binary supports `--help` for example (which may not be a huge issue I agree)

Instead, I suggest not run the command and simply stop at `Cmd.make`

However, the command must be looked up in the correct environment. And this, it appears, is not being done right

https://github.com/ocamllabs/vscode-ocaml-platform/blob/718aff3cd2deea01cd1178d0045fb5b206ff2256/src/Toolchain.ml#L245-L250

(from the master)
We don't have to run it. Previously, things were like this

https://github.com/ocamllabs/vscode-ocaml-platform/blob/a344356ee0b9cb4f263f5149a4c845dd981f04db/src/Toolchain.ml#L348-L358

> Use current approach, meaning use package manager (opam or esy) to access binary. In that case I still do not understand how am I supposed to use Cmd.make?

That snippet should answer how to use `Cmd.make`.  Both toolchain managers provide `env` function (Opam.env and Esy.env) - they return the environment dictionary. Providing it as a second arg to `Cmd.make` would correctly look for `ocamllsp` in esy's sandbox or opam's switch

I don't have this repo handy right away. I can raise a PR soon. But I think I had to point this out - running the binary isn't strictly required.

And by no means, is this a blocker right now. Quick shallow look at the diff seems okay (I couldn't test it though)

rgrinberg:
> Most of them cannot be tested by which or alternatives.

Sure they can be. Both esy and opam give us a full environment for a sandbox (`{Opam,Esy}.env`). Regardless, the workaround here is good enough. Can you use `--help=plain` however?

PR #149, opened 2020-04-21
Do not show error on fallback

rusty-key:
We discussed merging error and suggestion messages but I think that error still useful sometimes.  Here's my proposal:
— At the start, check if there is active configuration
— If there is, try to run `ocamllsp` with it, if it fails, show the error
— If there is not, show the suggestion and fall back to Global
— If there is global `ocamllsp`, run it
— If there is no global `ocamllsp`, do nothing, wait until configured.
— After user configures extension, we are back at square one.

Here are all of the possible cases:

**User has no active configuration, has no global `ocamllsp`**
User gets "Select toolchain" message.

**User has no active configuration, has global `ocamllsp`**
User gets "Select toolchain" message.
Extension starts working using global `ocamllsp` 

**User has active configuration, `ocamllsp` is not available in selected toolchain**
User gets error message

**User has active configuration, `ocamllsp` is available**
Extension starts working without any messages.

What do you think?

### DIFF
--- a/src/Extension.ml
+++ b/src/Extension.ml
@@ -61,7 +61,8 @@ let selectSandbox (instance : Instance.t) () =
 let suggestToSetupToolchain instance =
   let open Promise.O in
   Vscode.Window.showInformationMessage'
-    "Extension is unable to find ocamllsp automatically. Please select package manager you used to install ocamllsp for this project."
+    "Extension is unable to find ocamllsp automatically. Please select package \
+     manager you used to install ocamllsp for this project."
     [ ("Select package manager", ()) ]
   >>| function
   | None -> ()
--- a/src/Extension.ml
+++ b/src/Extension.ml
@@ -75,18 +75,20 @@ let activate _context =
   let open Promise.O in
   let toolchain =
     Toolchain.ofSettings () >>| fun pm ->
-    let resources =
+    let (resources, isFallback) =
       match pm with
       | None ->
         let (_ : unit Promise.t) = suggestToSetupToolchain instance in
-        Toolchain.PackageManager.Global
-      | Some toolchain -> toolchain
+        (Toolchain.PackageManager.Global, true)
+      | Some toolchain -> (toolchain, false)
     in
-    Toolchain.makeResources resources
+    (Toolchain.makeResources resources, isFallback)
   in
-  toolchain >>= fun toolchain ->
+  toolchain >>= fun (toolchain, isFallback) ->
   Instance.start instance toolchain
-  |> handleError Window.showErrorMessage
+  |> handleError (fun e ->
+      if isFallback then Promise.resolve () else Window.showErrorMessage e
+  )
   |> Promise.catch (fun e ->
          let message = Node.JsError.ofPromiseError e in
          Window.showErrorMessage {j|Error: $message|j})
--- a/src/Extension.ml
+++ b/src/Extension.ml
@@ -61,8 +61,8 @@ let selectSandbox (instance : Instance.t) () =
 let suggestToSetupToolchain instance =
   let open Promise.O in
   Vscode.Window.showInformationMessage'
-    "There is no toolchain attached to this project."
-    [ ("select toolchain", ()) ]
+    "Extension is unable to find ocamllsp automatically. Please select package manager you used to install ocamllsp for this project."
+    [ ("Select package manager", ()) ]
   >>| function
   | None -> ()
   | Some () -> selectSandbox instance ()


rgrinberg:
> — If there is no global ocamllsp, do nothing, wait until configured.

Isn't this a bit counter intuitive? The problem seems like it's not working, and it's not giving the user any hints why that's the case.

rusty-key:
But at this point user already sees a suggestion to choose sandbox

rusty-key:
Maybe we should just change make the suggestion message more clear?
```
There is no toolchain attached to this project 
[select toolchain]
↓↓↓
Extension is unable to find `ocamllsp` automatically. Please specify the toolchain you are using.
[Select toolchain]
```
Also, maybe change "toolchain" to "package manager"?

rusty-key:
@rgrinberg, I added breakdown of possible cases to the description, please take a look:

> **User has no active configuration, has no global `ocamllsp`**
> User gets "Select toolchain" message.
> 
> **User has no active configuration, has global `ocamllsp`**
> User gets "Select toolchain" message.
> Extension starts working using global `ocamllsp` 
> 
> **User has active configuration, `ocamllsp` is not available in selected toolchain**
> User gets error message
> 
> **User has active configuration, `ocamllsp` is available**
> Extension starts working without any messages.

rgrinberg:
@rusty-key that last breakdown seem reasonable.

I prefer the term "sandbox" over package manager or toolchain actually. In the end, we don't care about installing any packages after all - only what is available.

prometheansacrifice:
> I prefer the term "sandbox" over package manager or toolchain actually. In the end, we don't care about installing any packages after all - only what is available.

I'm working on removing redundant installation of tools - thinking of setting up the toolchain in a shared location, but sandboxed by esy. At that point, sandbox would be shared and not project local (much like opam). I tend to associate the word sandbox with a special project - hence prefer the terms toolchain/toolchain manager

rusty-key:
We can call it whatever we want in the code, I am talking about what term we should expose to users. It should be clear in their world, not in our.

Toolchain is confusing, because we don't call neither opam nor esy "toolchain".
Sandbox is better, but I think is still confusing as people do have specific terms for them (such as switch for opam).
Package manager is the clearest for me even though it's a bit clunky. 

> In the end, we don't care about installing any packages after all - only what is available.

Agree, but we are using package manager to get what we need. 

Maybe? 

> Extension is unable to find `ocamllsp` automatically. Please select package manager you used to install `ocamllsp` for this project.
> [Select]

prometheansacrifice:
I see what you mean now - for someone now familiar with the internals,
package manager is probably the most they can relate to

rusty-key:
So, are we good here?

src/Extension.ml
```diff
@@ -61,8 +61,8 @@ let selectSandbox (instance : Instance.t) () =
 let suggestToSetupToolchain instance =
   let open Promise.O in
   Vscode.Window.showInformationMessage'
-    "There is no toolchain attached to this project."
-    [ ("select toolchain", ()) ]
+    "Extension is unable to find ocamllsp automatically. Please select package manager you used to install ocamllsp for this project."
```
rgrinberg, line 64: Could you format the code with ocamlformat?

rusty-key: Formatting triggers too many unrelated changes, I did in separate PR: https://github.com/ocamllabs/vscode-ocaml-platform/pull/152

rusty-key: Formatted added string: https://github.com/ocamllabs/vscode-ocaml-platform/pull/149/commits/e8f5755c55c520725b135f4064840a339c31919a

yannl35133:
I have a problem with the way you phrased the notification :
- Why does it say it cannot find `ocamllsp`, if it finds it after all?
- If the problem is that we are using the global `ocamllsp`: why is it a problem?

ManasJayanth:
@yannl35133 It is problematic in the sense there could be compiler version mismatches

At the time, we believed, simply finding the lsp binary wasn't enough and that it had to be the exact version (ie the binary compiled with the compiler in the current switch/sandbox)

I can see how this is confusing. We can definitely improve the message.

yannl35133:
I don't get where the mismatch happens: does `ocamllsp` behave differently with sources depending on the version of ocaml it is compiled with?

ManasJayanth:
Indeed it does - Merlin used underneath needs binary compatibility with the .cmt files

yannl35133:
I don't know what the .cmt files are; simply, what components do we need to make sure they are using the same version of ocaml?

yannl35133:
Ok, now that I've done my research, I think I get it : we need to call ocamlc to generate the .cmt file for the source we are reading, and then ocamllsp and merlin under the hood to analyse this .cmt file. Is that right?
In this case, there would be a problem if one were to use (eg) the APT ocamlc and opam ocamllsp which could have different versions of ocaml?

If this is right, are there other possibilities of discrepancy? Can we address this issue more evidently in the message?

ManasJayanth:
Another point I recall now is missing refmt for Reason syntax - if for some reason refmt is missing, merlin fails.

But @rgrinberg suggested we handle this in a different place (which could handle [this](https://github.com/ocamllabs/vscode-ocaml-platform/pull/184) as well.

So, I think, it was these two scenarios (at the time) that made us pick a message that covers both.

Ideally, the message should instruct the user to select the appropriate opam switch or esy sandbox and rebuild if needed (which I believe the package manager takes care of)

yannl35133:
As far as I understand, the default is a mix between complaining and doing as if the global toolchain (the selected switch in opam) were selected in VSCode. Why don't we do as if the user selected "Global" by default?

PR #152, opened 2020-04-23
Fix format script

rusty-key:
Script relied on outside binary. Fixed it to use local `ocamlformat` installed with `esy` and ran formatter for all files.

Here's the actual change: https://github.com/ocamllabs/vscode-ocaml-platform/pull/152/commits/31b0849ad2524b9627ce86e1cfea053047ec817b

Also adding `ocamlformat` check to CI workflow

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -28,7 +28,10 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install packages
-        run: yarn --frozen-lockfile && yarn deps
+        run: yarn --frozen-lockfile
+
+      - name: Install esy deps
+        run: yarn deps
 
       - name: Check formatting
         run: yarn fmt:check
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -31,7 +31,7 @@ jobs:
         run: yarn --frozen-lockfile
 
       - name: Install esy deps
-        run: yarn deps
+        run: yarn esy:deps
 
       - name: Check formatting
         run: yarn esy:fmt:check
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -28,7 +28,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install packages
-        run: yarn --frozen-lockfile
+        run: yarn --frozen-lockfile && yarn deps
 
       - name: Check formatting
         run: yarn fmt:check
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -34,7 +34,7 @@ jobs:
         run: yarn deps
 
       - name: Check formatting
-        run: yarn fmt:check
+        run: yarn esy:fmt:check
 
       - name: Build
         run: yarn package
--- a/package.json
+++ b/package.json
@@ -284,7 +284,7 @@
     "test": "yarn run test:unit && yarn run test:opam ",
     "vscode:prepublish": "yarn run build",
     "package": "vsce package --yarn",
-    "ocamlformat": "git ls-files '*.ml' '*.mli' | xargs esy exec-command -- ocamlformat",
+    "ocamlformat": "git ls-files '*.ml' '*.mli' | xargs yarn esy exec-command -- ocamlformat",
     "fmt:check": "yarn ocamlformat --check && prettier --check .",
     "fmt": "yarn ocamlformat -i && prettier --write .",
     "deps": "esy install"
--- a/package.json
+++ b/package.json
@@ -279,10 +279,9 @@
     "test": "yarn run test:unit && yarn run test:opam ",
     "vscode:prepublish": "yarn run build",
     "package": "vsce package --yarn",
-    "fmt:check": "prettier --check .",
-    "fmt:ocaml": "git ls-files '*.ml' '*.mli' | xargs esy exec-command -- ocamlformat -i",
-    "fmt:metadata": "prettier --write .",
-    "fmt": "yarn run fmt:ocaml && yarn fmt:metadata"
+    "ocamlformat": "git ls-files '*.ml' '*.mli' | xargs esy exec-command -- ocamlformat",
+    "fmt:check": "yarn ocamlformat --check && prettier --check .",
+    "fmt": "yarn ocamlformat -i && prettier --write ."
   },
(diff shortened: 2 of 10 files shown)


rusty-key:
Oh, looks like we don't have `esy` on CI. Is it intentional?

rgrinberg:
Yes. The plugin should work without esy. Although we could have a separate plugin + esy pass if it’s necessary to verify some esy related functionality.

rusty-key:
This is not about extension end-user functionality, it's about extension development. Right now new developer will have to know about this external dependency and somehow install it. 

I think we should unified way of installing all of the dependencies required for development. I thought in our case it's `esy install` since we have `esy.json`. But it can be local opam switch too, I don't mind, but we need to set it up then.

rgrinberg:
Isn't esy already usable with the current setup? You can just run the formatting script inside an esy environment.

rusty-key:
> Isn't esy already usable with the current setup

I am not sure what you mean. Esy is usable, yes and in this PR I am utilizing this fact by running `ocamlformat` via `esy`. The only problem is that we don't have `esy` on CI:
```
$ git ls-files '*.ml' '*.mli' | xargs esy exec-command -- ocamlformat --check
9
xargs: esy: No such file or directory
```
https://github.com/ocamllabs/vscode-ocaml-platform/runs/612071421

We can install `esy` on CI at one of the prep step, but I wanted to make sure that it's ok.

rusty-key:
Here's what I am suggesting: 

Add esy as dev dependency:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/089e5692753ddc55153a9e861c8be8598acbd80d/package.json#L304

Install esy deps on CI:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/089e5692753ddc55153a9e861c8be8598acbd80d/package.json#L290
https://github.com/ocamllabs/vscode-ocaml-platform/blob/089e5692753ddc55153a9e861c8be8598acbd80d/.github/workflows/workflow.yml#L33-L34

Run ocamlformat check via esy:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/089e5692753ddc55153a9e861c8be8598acbd80d/package.json#L287

rgrinberg:
Let me just state some facts about the situation:

* I don't use esy
* I've added the ocamlformat command to yarn so that I can format my code
* This script now requires esy to be installed to use my script.
* The consequence of this is that the script is now broken for me

> Here's what I am suggesting:

It was already pretty before, thank you. I have no desire to install esy just to format my code.

imbsky:
I recently opened a PR that adds a Makefile, it should be able to use formatters from both esy and opam. Is there any reason you prefer this changes over that?

rusty-key:
@rgrinberg, this is probably misunderstanding. I am not asking you to change your workflow, I am concerned about two things:
1. Ensuring that files are formatted on CI (run `ocamlformat --check`) in order to avoid situations we are in right now, when a bunch of files are committed with wrong formatting . For that we need some way of bootstrapping proper environment on CI.
2. New developer experience. Right now, in order to run formatter new developer will have to install external dependency, and he will have to know about it, and have proper version compiled with proper version of ocaml. Otherwise, he will be getting wrong output (in case of wrong ocamlformat) or no output at all (in case of wrong ocaml). All of this is hidden knowledge.

With esy (or with local opam switch) we can provide a single command that will bootstrap project and install all dependencies required for development. That's it. It doesn't mean you have to change anything in your workflow. 

> I've added the ocamlformat command to yarn so that I can format my code

I didn't know that. I can revert this changes. I also missed @imbsky's PR, maybe we can have a single yarn command that can run ocamlformat from local switch/esy with fallback to global. 

> I recently opened a PR that adds a Makefile, it should be able to use formatters from both esy and opam. Is there any reason you prefer this changes over that?

@imbsky, I missed your PR sorry. I am definitely prefer your way of formatting, but I don't see how it can use esy's ocamlformat if it is installed. It looks like it just uses `ocamlformat` available in global.

imbsky:
I think you can do it with `esy make fmt` as @rgrinberg said.

rusty-key:
@imbsky, but it doesn't solve any problems I mentioned in my previous comment

rgrinberg:
> Ensuring that files are formatted on CI (run ocamlformat --check) in order to avoid situations we are in right now, when a bunch of files are committed with wrong formatting . For that we need some way of bootstrapping proper environment on CI.

To solve this problem, we can use either opam or esy to install ocamlformat in CI like you've done in this PR. However, we should make sure the current package manager independent commands remain usable.

> New developer experience. Right now, in order to run formatter new developer will have to install external dependency, and he will have to know about it, and have proper version compiled with proper version of ocaml. Otherwise, he will be getting wrong output (in case of wrong ocamlformat) or no output at all (in case of wrong ocaml). All of this is hidden knowledge.

Do you think documentation in the readme is sufficient for this? I don't know how this problem is solved for other bucklescript projects, but we could certainly leave some instructions for how to set things up with esy (and opam, in case anyone is interested).

> With esy (or with local opam switch) we can provide a single command that will bootstrap project and install all dependencies required for development. That's it.

What should this command be? I thought that `$ esy yarn ocamlformat` would be sufficient for this, but if there are other suggestions then I'm all ears.

> It doesn't mean you have to change anything in your workflow.

That's all I'm asking.

imbsky:
Sorry for the late reply. I also think it's enough just to write instructions in a README.

rusty-key:
Changed PR to provide esy alternatives specifically for CI. 

> Do you think documentation in the readme is sufficient for this?

esy.json or local switch could be single source of truth for development dependencies and their versions. Writing instructions about in README will provide second with inevitable conflicts in the future and will require additional maintenance. Probably I am pre-optimizing though. If you think that having instructions is good enough, then I am ok with that.

> What should this command be?

There are options, depending on what we want to use:

— If we require global `esy`, than all of the deps can be installed with single `esy`.
— Alternatively, we can use local `esy`: specify it as dev dependency in package.json then use it via yarn as I am doing in this PR.
— With `opam` it's a bit trickier because we will have to provide bootstrap script. But it can be something like:
```
"opam:bootstrap": " opam switch create ./ [version]",
"opam:deps": "opam install ."
```

rgrinberg:
Looks good to me. Could you add `doc-comments-val=after` to the ocamlformat configuration so that the spurious formatting changes aren't introduced?

imbsky:
Similar changes were made when the Makefile was added.

Issue #106, opened 2020-04-04
Improve dune highlighting


mnxn:
The syntax for dune files is missing highlighting for certain keywords and other parts of the file format (operators, variables, etc): 
![image](https://user-images.githubusercontent.com/25037249/78419709-5d41f200-75fd-11ea-87a8-acedf976dd10.png)

rgrinberg:
Another nice thing to have would be proper syntax highlighting for dune files that use OCaml syntax. In other words, files that have the following first line: `(* -*- tuareg -*- *)`

SaySayo:
Can i work on this issue?

SaySayo:
Thank you! I have a basic understanding of RegEx and I'd appreciate if you recommend more resources on it.

tmattio:
https://www.sitepoint.com/learn-regex/ also seems like a great tutorial, and I'd recommend using https://regex101.com/ as a playground to test different RegEx while learning about them.

mnxn:
To be clear, the improvements I had in mind are to rewrite [dune.json](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/syntaxes/dune.json) in a similar fashion to the [dune-project.json](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/syntaxes/dune-project.json) and [dune-workspace.json](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/syntaxes/dune-workspace.json) syntaxes.

That is, only valid stanzas should be highlighted in the appropriate contexts. The `dune-project` and `dune-workspace` syntax highlighting already does this (although it may be slightly outdated, I'm not sure). This requires heavily referencing https://dune.readthedocs.io/en/stable/dune-files.html.

It's a lot of tedious work (the `dune` format is much larger than both `dune-project` and `dune-workspace`), so I'm not sure this issue is viable as a "good first issue".

tmattio:
Closed by #742

PR #154, opened 2020-04-24
dune-project syntax highlighting

mnxn:
_Partially_ addresses #106.

Sorry it took so long but here it is:

dune-project

The `general` syntax rules will be extracted into the dune grammar once that it rewritten (soon).


rgrinberg:
Looks great!

PR #159, opened 2020-04-30
Drop BSB autosetup

rusty-key:
There are two main problems with BSB autosetup:
1. It's has many points when it could fail: network issues, esy version mismatch, build errors. When it's failing, it's hard for user to pinpoint the problem as most of the possible errors are hidden by design. 
2. Extension requires esy for _some_ projects, but does it implicitly which is also not clear for user.

Ultimately, this is a workaround: the only reason we need it, is because there is no other way of installing required dependencies in JS world. 

This PR removes autosetup and adds instructions for installing dependencies with `esy` to the README. 

The further plan is:
— add esy.json file generation/modification;
— with @rgrinberg's help, explore a way of bringing `ocamlformat`/`ocamllsp` to npm;
— after binaries are available via npm, add "native" support for bsb projects.

@rgrinberg, @prometheansacrifice, please take a look


prometheansacrifice:
I like PR #164 better, as leaving in the current mechanism in a disabled code path would be convenient when we try to improve/fix it.

Voting in favour of that PR.

rusty-key:
👌

PR #161, opened 2020-05-01
Remove Azure Pipelines

imbsky:
The test still failed, but it's not a problem with this PR itself.


prometheansacrifice:
We rely on Azure for the prebuilts. I'm not aware of this being done on GH actions - has it been addressed?

imbsky:
Not yet, but it will be possible soon.

rgrinberg:
More to the point, are we running the integration tests with this pipeline?

imbsky:
That's right. Also, automatic toolchain setup is currently disabled, and I remember that we have decided not to use Azure's artifact to distribute pre-built binaries. So we might be able to merge this, what do you think?

rgrinberg:
Well, it’s more like that this package doesn’t produce any binaries so the publishing should be done outside this package. It doesn’t really matter if it’s done in azure or github actions. The integration tests still need to be be ported to CI however

imbsky:
The integration test is already ported to Actions but is now broken regardless of this PR.
https://github.com/ocamllabs/vscode-ocaml-platform/pull/161/files#diff-f09ce9d24f2aeb928cee946c13324a06R54-R57

rgrinberg:
I see. You need to clarify stuff like this in the description 🙂

Okay, so the pr indeed looks good. Let me just have a more in depth review later.

imbsky:
As said in https://github.com/ocaml/dune/pull/3451, there are many benefits to migrating to Actions for CIs that need an OCaml environment. You will also benefit more in the future.

imbsky:
> You need to clarify stuff like this in the description

Yes, indeed...

.github/workflows/workflow.yml
```diff
@@ -13,27 +13,50 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - windows-latest
+        ocaml-version:
+          - 4.06.1
```
rgrinberg, line 17: Why 4.06?

imbsky: I thought we needed this version for testing with BuckleScript. Maybe it's not?

rgrinberg: Oh yes. Although I'm not sure we actually have any bucklescript tests at the moment.

imbsky:
Actions VM doesn't have the global install directory of yarn in the path, so we will install esy with npm.

PR #167, opened 2020-05-02
Add dune and dune-project snippets

imbsky:


snippets/dune-project.json
```diff
@@ -0,0 +1,26 @@
...
+    "body": ["(name $1)"]
+  },
+  "using": {
+    "prefix": "using",
+    "body": ["(using $1 $2)"]
+  },
+  "info": {
+    "prefix": "info",
+    "body": [
+      "(license $1)\n(authors \"$2\")\n(maintainers \"$3\")\n(source (github $4))\n(bug_reports \"$5\")\n(homepage \"$6\")\n(documentation \"$7\")"
```
mnxn, line 17: I don't know how prevalent github is in the ocaml community, but users aren't necessarily going to be using it.
Maybe this should default to `uri` instead of `github`.

imbsky: Indeed! I made it possible to choose. What do you think?
image

mnxn: Looks good

PR #169, opened 2020-05-03
Dune Formatting

mnxn:
Adds formatting for dune/dune-project/dune-workspace with `dune format-dune-file`. Dune must be on the PATH for the formatting to work; otherwise, it silently fails and no formatting is preformed.

There is no formatting configuration as everything is handled by `format-dune-file`. 1-space indentation is always used. Line endings (CRLF or LF) seem to be preserved on format.

<details><summary>Example</summary>
<p>

```
(package
(name ocaml-lsp-server)
(synopsis "LSP Server for OCaml")
(description "An LSP server for OCaml.")
(depends
yojson
stdlib-shims
menhir
ocaml-syntax-shims
ppx_yojson_conv_lib
(ocamlformat :with-test)
(ocamlfind (>= 1.5.2))
(ocaml (>= 4.06))
(dune (>= 2.0))))
```
will be turned into:

```
(package
 (name ocaml-lsp-server)
 (synopsis "LSP Server for OCaml")
 (description "An LSP server for OCaml.")
 (depends
  yojson
  stdlib-shims
  menhir
  ocaml-syntax-shims
  ppx_yojson_conv_lib
  (ocamlformat :with-test)
  (ocamlfind
   (>= 1.5.2))
  (ocaml
   (>= 4.06))
  (dune
   (>= 2.0))))
```
</p>
</details>

Parts of the VSCode bindings had to be updated in order to support interfacing with `registerDocumentFormattingEditProvider`. The Node bindings were updated to support writing to stdin for a `ChildProcess`


rgrinberg:
> Indentation size and whether to use tabs/spaces is detected from vscode options

Would it be possible to set the default to be 1 space for dune files? This would make this feature much more useful.

mnxn:
> Would it be possible to set the default to be 1 space for dune files? This would make this feature much more useful.

Yep! https://code.visualstudio.com/api/references/contribution-points#contributes.configurationDefaults

I'll write up the defaults in a second

mnxn:
Took a lot longer than a second haha.

Turns out that setting a configuration default for a languageId with a dot in it doesn't seem to work. I renamed all occurrences of `ocaml.dune` to `dune` and it works now.

I also added a default for ocaml for 2 space indents.

mnxn:
I just found out that dune has a formatting command already: https://github.com/ocaml/dune/blob/master/bin/format_dune_file.ml

I'll rewrite this PR to use that. It seems much better to rely on that command than on the naive implementation I had before.

mnxn:
@rgrinberg @rusty-key @prometheansacrifice 

It would be great if you could look over this please.

src/DuneFormatter.ml
```diff
@@ -0,0 +1,24 @@
...
+  in
+  (* selects entire document range *)
+  let fullDocumentRange =
+    Range.make ~startLine:0 ~startCharacter:0 ~endLine:lastLine
+      ~endCharacter:lastCharacter
+  in
+  (* text of entire document *)
+  let documentText = TextDocument.getText document fullDocumentRange in
+
+  ChildProcess.exec "dune format-dune-file" ~stdin:documentText
```
rgrinberg, line 16: Don't we have the `Cmd` module for this?

mnxn: Ahh. Must have overlooked it sorry.

rgrinberg:
> Ahh. Must have overlooked it sorry.

No worries. If it doesn’t work for your use case, feel free to introduce more bindings or to improve the API of cmd.

mnxn:
> No worries. If it doesn’t work for your use case, feel free to introduce more bindings or to improve the API of cmd.

I'm rewriting it to use Cmd right now. I'll add an optional field to Cmd.output for stdin.

mnxn:
> This needs a CHANGES entry as we're adding new functionality

I already updated CHANGELOG.md, is there something else that is needed?

rgrinberg:
Oops. Looks like, I've scrolled too fast and this is good to go.

By the way, the dune formatting functionality is actually something that you can directly add to ocaml-lsp right away (if you're interested). It's unlikely that we'll have a separate LSP server for dune, so we might as well use the server that we already have.

mnxn:
I'll look into it. It would be useful to have the feature in other editors that use ocaml-lsp too.

PR #176, opened 2020-05-05
Discover esy in parent folders

rusty-key:
Right now we only looking for esy in projectRoot, but sometimes esy can be configured in some other parent folder.

This is a dirty implementation of looking in parent directories. Is there a better way to do it? cc @rgrinberg @prometheansacrifice

### DIFF
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -80,7 +80,7 @@ module Discover = struct
       [||] results
     |> Array.to_list
 
-  let run ~dir : Path.t list Promise.t =
+  let parseDir dir =
     let open Promise.O in
     (Path.toString dir |> Fs.readDir >>| function
      | Error _ ->
@@ -89,10 +89,18 @@ module Discover = struct
          (Path.toString dir);
        [||]
      | Ok res -> res)
-    >>= fun files ->
-    files
-    |> Js.Array.map (parseFile dir)
-    |> Promise.all |> Promise.map foldResults
+    >>= fun files -> files |> Js.Array.map (parseFile dir) |> Promise.all
+
+  let getUpstreamDirs dir =
+    let dirList = dir |> Path.split in
+    dirList |> List.length
+    |. Belt.List.makeBy (fun i ->
+           dirList |. Belt.List.take (i + 1) |. Belt.Option.map Path.concat)
+    |. Belt.List.keepMap (fun x -> x)
+
+  let run ~dir : Path.t list Promise.t =
+    dir |> getUpstreamDirs |> Array.of_list |> Array.map parseDir |> Promise.all
+    |> Promise.map (fun res -> res |> Array.to_list |> Array.concat |> foldResults)
 end
 
 let discover = Discover.run
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -71,13 +71,14 @@ module Discover = struct
       | _ -> None |> Promise.resolve )
 
   let foldResults results =
-    Js.Array.reduce
-      (fun acc x ->
-        Js.Array.concat acc
-          ( match x with
-          | Some x -> Array.of_list [ x ]
-          | None -> [||] ))
-      [||] results
+    results |> Array.to_list |> Array.concat
+    |> Js.Array.reduce
+         (fun acc x ->
+           Js.Array.concat acc
+             ( match x with
+             | Some x -> Array.of_list [ x ]
+             | None -> [||] ))
+         [||]
     |> Array.to_list
 
   let parseDir dir =
@@ -91,17 +92,17 @@ module Discover = struct
      | Ok res -> res)
     >>= fun files -> files |> Js.Array.map (parseFile dir) |> Promise.all
 
-  let getUpstreamDirs dir =
-    let dirList = dir |> Path.split in
-    dirList |> List.length
-    |. Belt.List.makeBy (fun i ->
-           dirList |. Belt.List.take (i + 1) |. Belt.Option.map Path.concat)
-    |. Belt.List.keepMap (fun x -> x)
+  let rec parseDirsUp' parsedDirs dir =
+    let parsedDirs = parseDir dir :: parsedDirs in
+    match Path.parent dir with
+    | None -> parsedDirs
+    | Some dir -> parseDirsUp' parsedDirs dir
+
+  let parseDirsUp dir = parseDirsUp' [] dir
 
   let run ~dir : Path.t list Promise.t =
-    dir |> getUpstreamDirs |> Array.of_list |> Array.map parseDir |> Promise.all
-    |> Promise.map (fun res ->
-           res |> Array.to_list |> Array.concat |> foldResults)
+    dir |> parseDirsUp |> Array.of_list |> Promise.all
+    |> Promise.map foldResults
 end
 
 let discover = Discover.run
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -70,17 +70,6 @@ module Discover = struct
       | ".opam" -> Promise.return (Some manifestFile)
       | _ -> None |> Promise.resolve )
 
-  let foldResults results =
-    results |> Array.to_list |> Array.concat
-    |> Js.Array.reduce
-         (fun acc x ->
-           Js.Array.concat acc
-             ( match x with
-             | Some x -> Array.of_list [ x ]
-             | None -> [||] ))
-         [||]
-    |> Array.to_list
-
   let parseDir dir =
     let open Promise.O in
     (Path.toString dir |> Fs.readDir >>| function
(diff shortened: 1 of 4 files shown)


src/esy.ml
```diff
@@ -91,10 +89,19 @@ module Discover = struct
          (Path.toString dir);
        [||]
      | Ok res -> res)
-    >>= fun files ->
-    files
-    |> Js.Array.map (parseFile dir)
-    |> Promise.all |> Promise.map foldResults
+    >>= fun files -> files |> Js.Array.map (parseFile dir) |> Promise.all
+
+  let getUpstreamDirs dir =
```
prometheansacrifice, line 94: This would go all the way up to root, right? I think we should stop at home. 

On Windows, `Path.split` will all result in `C:` being one of the entries in the loop, which I think is not valid too

prometheansacrifice:
I'm not sure how I feel about this - if a valid esy manifest file (package.json/esy.json) is not found in the project root, I don't think we should look at other locations. Is there a use case you have in mind @rusty-key ?

rusty-key:
@prometheansacrifice in my mind it's primarily about supporting monorepos. For example, we have multiple frontend projects with shared dependencies: 
```
frontend/
- project1/
- project2/
- project3/
- package.json
```
But developers usually work on one project, so they are opening only `frontend/project1` in their workspace. But that means that every workspace will have to have their own esy.json. In my mind it makes sense to put `esy.json` to the root and be able to find it from under project's workspace.

prometheansacrifice:
> In my mind it makes sense to put esy.json to the root and be able to find it from under project's workspace.

I agree.

My initial thoughts were - when CLIs do this (like esy or dune), there is an explicit log message in the terminal. I think, similarly, if the discovered esy manifest is not in the project root, we must issue an alert message saying "esy.json was found at /some/path"

What do you think?

prometheansacrifice:
Without an explicit message, I think, it could lead to confusion

rusty-key:
We don't do anything automatically. All found "sandboxes" are presented to user and he is selecting whatever he wants: 
Screen Shot 2020-05-05 at 6 47 22 PM

src/path.ml
```diff
@@ -26,3 +26,8 @@ let relative_all p xs = List.fold_left Filename.concat p xs
 let join x y = Filename.concat x y
 
 let withExt x ~ext = x ^ ext
+
+let split x =
+  x |> Js.String.split sep |> Array.to_list |> List.filter (fun p -> p != "")
```
rgrinberg, line 31: How is it possible to get an empty component here?

src/path.ml
```diff
@@ -26,3 +26,8 @@ let relative_all p xs = List.fold_left Filename.concat p xs
...
+
+let concat x = sep ^ (x |> Array.of_list |> Js.Array.joinWith sep)
```
rgrinberg, line 33: Can't you just use `String.concat`? Also, the name `Path.concat` seems confusing. The resultant path is going to be absolute, which is definitely not something I'd expect.

src/esy.ml
```diff
@@ -91,10 +89,19 @@ module Discover = struct
...
+    let dirList = dir |> Path.split in
+    dirList |> List.length
+    |. Belt.List.makeBy (fun i ->
+           dirList |. Belt.List.take (i + 1) |. Belt.Option.map Path.concat)
+    |. Belt.List.keepMap (fun x -> x)
```
rgrinberg, line 99: This is a bit convoluted. It would be simpler if you'd introduce a primitive like this:

```
module Path : sig
  type t
  val parent : t -> t option
end
```

And then recurse with it until you reach root.

src/esy.ml
rgrinberg: I don't see why we should stop at $HOME. It will make the detection work differently depending on where the user has his project. That is incredibly counter intuitive and convoluted.

src/path.ml
rusty-key: When the path is absolute, first element will be empty:
```ocaml
Path.split "/path/to" 
(* ["", "path", "to"] *)
```
Also, I think it's possible to construct `/path//to` paths, so empty path could be in the middle too. Probably the module needs some revising.

src/path.ml
rusty-key: Sorry, I made this PR quickly, just to explore the idea, so code and naming may be not optimized

src/esy.ml
rusty-key: good idea, will do

rgrinberg:
Well the idea in this PR looks good to me. If the implementation is good, this is ready to merge.

rusty-key:
Ok, I'll find some time to refactor it tomorrow

src/esy.ml
prometheansacrifice: How would stopping at $HOME yield different results from the current implementation?

Files beyond HOME can be off-limits in terms of ownership

prometheansacrifice:
> We don't do anything automatically. All found "sandboxes" are presented to user and he is selecting whatever he wants

Will this show all the paths possible from the project all the way till the root, even if there is no esy.json? I think we should filter the candidate paths by the existence of the esy.json.

prometheansacrifice:
It also looks like there will be a lot of choices in that list - parent directories shouldn't make it there if the esy.json is found at the project root (is this taken care of already. Hope I didn't miss it)

src/path.ml
rgrinberg: I see. I think it's better to do away with `split` completely then. It's semantics aren't that clear on non normalized paths. For example, I would expect `foo/./bar` would remove the redundant `.`

prometheansacrifice: In case it helps, here is how esy looks for a manifest file

https://github.com/esy/esy/blob/master/bin/ProjectConfig.re#L110

src/esy.ml
rgrinberg: > How would stopping at $HOME yield different results from the current implementation?

Isn't it possible for users to store their sources outside $HOME? For example, a network drive.

> Files beyond HOME can be off-limits in terms of ownership

That's the case for files inside $HOME as well. It's possible to chmod inside a home directory after all. It's a good exceptional case to point out, but we should handle it regardless.

I don't think it ever pays off to be overly clever with these things. Especially when the benefit is not clear.

rusty-key:
@prometheansacrifice, in select we are only showing sandboxes, which means only directories that contain `esy.json`. In most cases users will see only one esy sandbox

src/esy.ml
jordwalke: Just lurking over here. Feel free to use this little one-file library I made called `Fp` that helps with a lot of this stuff if you just need a self-contained util: https://github.com/facebookexperimental/reason-native/tree/master/src/fp

Currently requires that you normalize platform paths on the way in, but we'll be adding those utility features directly.

rusty-key: Awesome! I guess we can mostly replace `Path` with `Fp` then. But I think I will do it in a separate PR as it will touch a lot of stuff

rusty-key:
Updated parsing approach to climb recursively up, please take a look.

src/esy.ml
rgrinberg: Just as an aside, we don't permit any reason sources in this repository. So you will have to port this module to OCaml first.

rusty-key:
@rgrinberg, @prometheansacrifice, what do you think? Is there anything else to improve in this PR?

prometheansacrifice:
Looks like we look for esy.json in the parent directory even if one of found in the current project root.

screenshot

Is there a reason to lookup parent directories when a manifest file is present in the project root?

prometheansacrifice:
> By the way, this got me thinking that we should really just allow the user to browse the filesystem and just point at whatever json file they want. This would cover all special cases (.vscode/esy.json for example) and would be the most straightforward escape hatch.

I think this idea of explicitly asking the user for the location of the esy.json

rusty-key:
> By the way, this got me thinking that we should really just allow the user to browse the filesystem and just point at whatever json file they want

I am trying to think of a case when esy is not in the root or upstream and cannot come up with one. Do you? 

I feel that current approach covers 99% of the cases and for the rest of them user can go and manually edit `settings.json` if needed. If you agree, I don't see why we would want to change automatic detection with manual.

> Is there a reason to lookup parent directories when a manifest file is present in the project root?

I can imagine minor monorepo case when you want to handle extension dependencies at root level while having another esy configuration locally. I guess we can remove it as it's very minor, but I don't see why if it's already implemented. What do you think?

prometheansacrifice:
I wish we could handle monorepos better - reliably figure the project root and expect users the place their esy.json there. For now, I have dont have better ideas, so I dont mind this feature for now (it is useful after all). My only qualm is the growing list (some possibly misleading) but we can address them in future maybe (I'll open an issue if I think of something)

It's 👍  from me for now

Khady:
> I wish we could handle monorepos better - reliably figure the project root and expect users the place their esy.json there.

A monorepo can contain multiple projects which will have different dependencies/settings/... So it seems expected that there will be different esy.json files in multiple places and that not always the same one should be used inside of a given project?

prometheansacrifice:
I was referring to the manifest file that contains the dev tools (lsp,
formatters etc) - these shouldnt be specified in multiple places IIUC

rgrinberg:
> I am trying to think of a case when esy is not in the root or upstream and cannot come up with one. Do you?

I don’t use esy very much at all so it’s hard for me to come up with use cases. The idea was to augment automatic detection rather than replace it

PR #181, opened 2020-05-06
Improve OCaml Syntax Highlighting

### LINKED ISSUE #180: Syntax highlighting for new quoted extensions syntax
https://github.com/ocaml/ocaml/blob/4.11/Changes#L8

mnxn:
Closes #180 
OCaml Syntax Highlighting for Quoted Extensions: 
| Old | New |
| - | - |
| old | new | 

Also fixes a fatal flaw in #175 that prevented function declarations from being highlighted:

| Old | New |
| - | - |
| old | new |

### DIFF
--- a/syntaxes/ocaml.json
+++ b/syntaxes/ocaml.json
@@ -211,7 +211,7 @@
         },
         {
           "comment": "let expression",
-          "match": "\\b(let)\\s+(rec\\s+)?(lazy\\s+)?([a-z_][A-Za-z0-9_']*)\\s*(?!\\s*,)",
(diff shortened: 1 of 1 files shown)


imbsky:
@mnxn This looks good. Are there any additional changes you would like to make in this PR?

imbsky:
Oh, you probably need to update the changelog.

mnxn:
No, this is it. I just had to fix some problems that were introduced in previous changes.

rgrinberg:
Just to clarify the policy on the change log:

* It should include bug fixes only when the bug was introduced in a previously released version. If the bug was only present in master, there's no need to mention it.

* It should mention new features being introduced in this release.

imbsky:
Oops, you're right.

PR #184, opened 2020-05-08
Allow setting refmt width

rusty-key:
There is currently no way of setting up line width for refmt other than passing flag to the command.

While we at this point, I am suggesting passing refmt width from extension's settings via ocamllsp to refmt arguments. Here's related PR in ocaml-lsp repo: https://github.com/ocaml/ocaml-lsp/pull/172

Related issue in reason repo: facebook/reason#2567


src/Extension.ml
```diff
@@ -15,9 +15,19 @@ module Client = struct
...
+      ~toJson:Json.Encode.int
+
   let make (toolchain : Toolchain.resources) :
       Vscode.LanguageClient.serverOptions =
     let command, args = Toolchain.getLspCommand toolchain in
+    let args =
+      match Settings.get refmtSettings with
+      | None -> args
+      | Some refmtWidth ->
+        Array.concat [ args; [| "--refmt-width"; string_of_int refmtWidth |] ]
```
rgrinberg, line 29: A command line option is not the correct place to pass this option. Preferably:

* It should be passed as a custom option to the formatting request (see the LSP spec).
* If this isn't possible, there's a mechanism for setting workspace configuration options.

prometheansacrifice: @rgrinberg I'm not sure about adding options to the requests from the client. We do have an option to send `initializationOptions` when we start the server

https://github.com/microsoft/vscode-languageserver-node/blob/master/client/src/client.ts#L508

rgrinberg:
> @rgrinberg I'm not sure about adding options to the requests from the client. We do have an option to send initializationOptions when we start the server

That’s better than the current PR, but it’s still not enough. When the user changes the setting, it shouldn’t be required to restart the server.

prometheansacrifice:
@rgrinberg Any particular reason for objecting to server restarts? The protocol is stateless, so I presumed it was ok

rgrinberg:
It's just bad user experience. The cache is blown out, there's a little latency to restarts, and the diagnostics must be recalculated. Not to mention that it's also more complex code wise because we must forget to trigger the restart. In this case, it's easy to do the right thing.

prometheansacrifice:
Do we support `workspace/configuration`?

Ref
1. https://microsoft.github.io/language-server-protocol/specification#workspace_configuration
2. https://github.com/microsoft/vscode-languageserver-node/blob/8e99b7645cf85d07cea1f7f93c71d632e58a9e8c/protocol/src/protocol.configuration.ts#L29-L37

rgrinberg:
workspace/configuration would be the 2nd best option if we can't just pass the option on every formatting call.

giltho:
See discussion on the ocaml-lsp PR : 
https://github.com/microsoft/vscode-languageserver-node/blob/363e9228a87d3df51427474ecc596740c3fecfd5/client/src/codeConverter.ts#L616

Using the high-level class LanguageClient in VSCode, the number of parameters passed to codeConverter is hardcoded, it's only part of what the protocol describes :/

So I'm workspace/configuration is the right answer until this is fixed

giltho:
I was going to raise an issue on the client implementation but there already is one and it's pretty recent.
https://github.com/microsoft/vscode-languageserver-node/issues/617

So it's not going to be fixed for a little while

TheSpyder:
It's a messy workaround, but could the extension instead provide a way to inject folders into the environment `PATH` that is used to run `ocamllsp`?

That way we could use a shell script to override what the `refmt` command does. This would have the same effect as setting the `PATH` before loading vscode from the command line, but it's much easier to set up.

rgrinberg:
@TheSpyder do you use `refmt` in a team? If yes, then I'm curious how you'd like to share this setting with the rest of your team. Is there a vscode config file shared by the team?

rgrinberg:
Closing this PR for now as it's no longer relevant. Please re-open if the code is updated.

TheSpyder:
@rgrinberg we are using vscode workspace settings. That does mean each repository needs to configure it separately but we are fine with that. The workaround I suggested wouldn't be shareable, but neither is the patch to do it by loading vscode from the command line (although perhaps the `refmt` shell script could be put into the repository).

rgrinberg:
What about developers that don't use vscode? What do they do?

Khady:
In emacs, the editor is passing the width to refmt based on the editor's settings. It combines very well with [editorconfig](https://editorconfig.org/). We can have the width we want in the .editorconfig file, it sets the settings in emacs, reason-mode picks it from the editor, everyone is happy.

https://github.com/reasonml-editor/reason-mode/blob/5690544a7091630e0ea0023bbbd57a733cea8bde/refmt.el#L191

rgrinberg:
> If something similar can happen in vscode that would be great.

Perhaps an even greater idea is just to have refmt to understand this. Then, all editors will be in the same happy boat. Moreover, we wouldn't need to implement this delicate implementation feature across N(for every editor) repositories.

TheSpyder:
We tried that approach, linked in the PR description. Nothing is happening on that end either.

A config option to specify the location of `refmt` would also work, since we're using a shell script to set the width anyway. That might also help people using a custom `refmt`.

PR #189, opened 2020-05-09
Fix tests

jchavarri:
`CheckBucklescriptCompat` was bringing `Import` module which imported `vscode`, which made jest failed for some reason.

Just inlining `mergeDicts` in `CheckBucklescriptCompat` (it's not used anywhere else) fixes the problem.

Also removed svg usage in readme, which apparently made a subsequent "Package extension" step [fail](url). 

Successful run: https://github.com/jchavarri/vscode-ocaml-platform/actions/runs/99952913.


prometheansacrifice:
Thanks!

PR #190, opened 2020-05-09
Fixes e2e tests for esy

prometheansacrifice:
(Review notes inlined)

### DIFF
--- a/fixtures/sample-esy/.vscode/settings.json
+++ b/fixtures/sample-esy/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "ocaml.sandbox": {
+    "root": "/path/will/be/populated/by/the/test",
+    "kind": "esy"
+  }
+}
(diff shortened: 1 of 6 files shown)


fixtures/sample-esy/.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
prometheansacrifice, line 1: We have to check in this file so that e2e dont assume global toolchain

test/suite/esy.test.js
```diff
@@ -7,15 +7,14 @@ const fs = require("fs-extra");
 const { Uri } = vscode;
 
 let root = path.dirname(path.dirname(__dirname));
-let fixtureSrcDir = path.join(root, "fixtures");
+let fixtureDir = path.join(root, "fixtures");
```
prometheansacrifice, line 10: Previously, these fixtures would be copied over to TMPDIR. But now that e2e need settings.json with the path of the esy toolchain specified in it relative to the project itself, e2e tests will setup the project in the work tree and run tests there.

prometheansacrifice:
Looks like adding the delay worked. Removing the --verbose. We can merge after that

PR #191, opened 2020-05-09
[CI] Cache esy, opam and yarn folders

jchavarri:
This PR adds caching for all esy, yarn and opam.

- yarn caching is taken from GH actions [example](https://github.com/actions/cache/blob/master/examples.md#node---yarn).
- esy from hello-reason [workflow](https://github.com/esy-ocaml/hello-reason/tree/master/.github/workflows).
- opam is more rough, it just takes the whole opam root folder calling `opam config var root` and stores it.

There are still ~5min that are spent calling the "use OCaml" action, as I guess it creates the switch etc. This is unnecessary as the cache has everything already, but I don't know if there's a cross-platform way to just install opam. If there is, we could use it when the workflow detects a cache hit.

### Times

Before ([example](https://github.com/ocamllabs/vscode-ocaml-platform/actions/runs/100010339)): 32m 0s 

After ([example](https://github.com/jchavarri/vscode-ocaml-platform/actions/runs/100038544)): 11m 57s

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -29,6 +29,11 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
+      - name: Use OCaml ${{ matrix.ocaml-version }}
+        uses: avsm/setup-ocaml@v1
+        with:
+          ocaml-version: ${{ matrix.ocaml-version }}
+
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
@@ -42,20 +47,19 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
+      - name: Get opam root directory path
+        id: opam-root-dir-path
+        run: echo "::set-output name=dir::$(opam config var root)"
+
       - name: Try to restore .opam folder
         id: cache-opam
         uses: actions/cache@v1
         with:
-          path: ~/.opam
+          path: ${{ steps.opam-root-dir-path.outputs.dir }}
           # opam cache is coupled to esy lock files, both caches will be wiped whenever esy deps change
           key:
             ${{ runner.os }}-opam-${{ matrix.ocaml-version }}-${{
             hashFiles('**/index.json') }}
-
-      - name: Use OCaml ${{ matrix.ocaml-version }}
-        uses: avsm/setup-ocaml@v1
-        with:
-          ocaml-version: ${{ matrix.ocaml-version }}
 
       - name: Install latest esy
         run: npm install --global esy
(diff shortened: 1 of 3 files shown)


.github/workflows/workflow.yml
```diff
@@ -24,26 +24,80 @@ jobs:
...
+      - name: Get opam root directory path
+        id: opam-root-dir-path
+        run: echo "::set-output name=dir::$(opam config var root)"
+
+      - name: Try to restore .opam folder
+        id: cache-opam
+        uses: actions/cache@v1
+        with:
+          path: ${{ steps.opam-root-dir-path.outputs.dir }}
+          # opam cache is coupled to esy lock files, so both caches will be wiped whenever esy deps change
```
jchavarri, line 59: not sure if there's a way to generate a "lock file" for opam and checking it in... for now it's tied to esy `index.json` file.

Khady: opam lock or opam switch export

Khady:
> There are still ~5min that are spent calling the "use OCaml" action, as I guess it creates the switch etc. This is unnecessary as the cache has everything already, but I don't know if there's a cross-platform way to just install opam.

The opam init call could be done with --bare so that it doesn't install a switch. But if you backup the whole opam root dir, then I think that you can even skip the whole opam init step.

prometheansacrifice:
@jchavarri Can you add a cache task for the downloaded vscode instance too? It is downloaded to the `.vscode-test` folder. In case, it helps, I did this for Azure Pipelines [here](https://github.com/arrowresearch/vscode-merlin/blob/master/.ci/utils/use-cache-vscode-test.yml#L9)

imbsky:
For npm modules of this size, it's faster not to cache. and cache thrashing may occur due to the current cache limit.

imbsky:
I'm currently working with the GitHub team to address the slow speed of setting up an OCaml environment using setup-ocaml. It's done completely in my spare time so there is no ETA, but there's definitely progress. The point here is that caching the opam directory on the user side is not currently recommended. Doing in so many repositories can cause a lot of confusion when we later migrate to the next version.

.github/workflows/workflow.yml
```diff
@@ -24,26 +24,80 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
```
imbsky, line 31: Don't create unnecessary diff.

imbsky:
I think this can be merged now. We will definitely need to make some changes in the near future, but I will take care of them.

imbsky:
It seems to take very little time to download the VSCode. I decided not to cache it in order not to make the workflow more complicated than necessary.

imbsky:
Thanks @jchavarri!

PR #193, opened 2020-05-11
Add Cram test syntax highlighting

### LINKED ISSUE #179: Syntax highlighting for cram tests.
Cram tests are commonly used in dune. Here's an example of the syntax:

```
Unless a line starts with an indentation, the line is considered as a comment. Anything that starts with indentation is considered to be a test that runs the command. The line below it is the expected output
  $ echo expected output
  expected output
  $ cat >file <<EOF
  > heredocs are commonly used to create larger files
  > EOF
```

The usual extension for such files is `.t`.

mnxn:
Closes #179 

Example:

example


imbsky:
I don't know much about Cram so I'll leave it to @rgrinberg.

PR #197, opened 2020-05-12
Simplify installation instructions for BuckleScript projects

wokalski:
I tested this locally. Please try to run it on your machines. I have a couple of projects with a similar setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,18 @@ npm i -g esy
 
 ```json
 {
-  "devDependencies": {
-    "ocaml": "4.6.x",
-    "@opam/ocaml-lsp-server": "ocaml/ocaml-lsp:ocaml-lsp-server.opam#360dd5",
+  "dependencies": {
+    "@opam/ocaml-lsp-server": "ocaml/ocaml-lsp:ocaml-lsp-server.opam",
+    "@opam/reason": "*",
+    "ocaml": "4.6.x"
+  },
+  "resolutions": {
+    "@opam/dune-configurator": "2.0.1",
+    "@opam/dune-private-libs": "2.0.1",
     "@opam/dune": "2.0.1",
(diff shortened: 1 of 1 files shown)


rgrinberg:
@imbsky feel free to merge if you've tested this out. I can't review since I'm not an esy user.

imbsky:
Since dune isn't strictly a development dependency, it should be placed in a resolutions field. So I think the current configuration is the simplest.

imbsky:
FYI @prometheansacrifice

wokalski:
It’s a bit disappointing to see the complexity (scary for new users) just for the sake of imaginary correctness.

rgrinberg:
> I see. But I'm sure there is a problem with the release process.

It's been a problem for years and it's unlikely to be fixed. I don't think we should point users to use the opam version of reason.

prometheansacrifice:
I think @wokalski 's changes can be re-introduced. Placing Dune in the dependencies correctly instructs the solver to pick `2.0.1` for other dune libraries too (dune-configurator and dune-private-libs). The Menhir overrides are not necessary since the newer menhirs work out of the box without any overrides. One of the Menhir releases in Feb introduced `fix` which was missing in the opam manifest and wouldn't build - pinning to an earlier version was a quick fix. It is not needed anymore

wokalski:
I just want to echo what @prometheansacrifice said. Additionally, since ocaml-lsp-server depends on newer dune now, the `esy.json` needs to include a commit hash of some sorts.

rgrinberg:
> I just want to echo what @prometheansacrifice said. Additionally, since ocaml-lsp-server depends on newer dune now, the `esy.json` needs to include a commit hash of some sorts.

Ocaml lsp server already includes this as a constraint in the opam file. Isn’t that enough?

Feel free to make a PR for the reason change however

PR #198, opened 2020-05-12
Add ATD syntax highlighting

### LINKED ISSUE #196: syntax highlighting for atd files
It would be nice if you also support syntax highlighting for [atd](https://marketplace.visualstudio.com/items?itemName=ocamllabs.ocaml-platform) files.

mnxn:
Closes #196 

Example: 
example

The ATD and OCaml syntaxes are fairly similar, so the ATD syntax `include`s parts of the OCaml syntax for simplicity.


mnxn:
It does highlight `let` expressions. I just thought of a way to fix it, will be updated soon.

mnxn:
let expressions and other ocaml-specific statements are no longer highlighted in ATD files

rgrinberg:
Yay. Would be nice if there was a way to make this work for mli files as well :)

mnxn:
> Would be nice if there was a way to make this work for mli files as well :)

Unfortunately the interaction between the syntax highlighting of .ml and .mli files is more complex, especially since extension nodes can include expressions inside a .mli. I plan to revisit the OCaml grammar soon, so I will try my best to include that feature.

PR #202, opened 2020-05-14
Output window should not steal focus

rgrinberg:


src/Extension.ml
```diff
@@ -11,6 +11,7 @@ module Client = struct
         [| { scheme = "file"; language = "ocaml" }
          ; { scheme = "file"; language = "reason" }
         |]
+    ; revealOutputChannelOn = Some Never
```
TheSpyder, line 14: `jsConverter` isn't automatic, unfortunately
https://bucklescript.github.io/docs/en/generate-converters-accessors#convert-between-js-integer-enum-and-bs-variant

rgrinberg:
I'll merge and make a release since people are very annoyed by this.

PR #208, opened 2020-05-17
Improve Dune formatting

### LINKED ISSUE #206: Improve dune formatting
* Use dune from the sandbox
* Handle missing dune binary gracefully

mnxn:
Close #206 

Adds support for running `format-dune-file` from sandboxes. I have tried it with opam, it would be great if somebody could try it with esy. 

Errors are now shown when the file fails to parse or dune cannot be found.


src/bindings/Node.ml
```diff
@@ -201,12 +200,8 @@ module ChildProcess = struct
     Promise.make @@ fun ~resolve ~reject:_ ->
     let cp = ref [%bs.obj { exitCode = 0 }] in
     cp :=
-      exec cmd options (fun err stdout stderr ->
-          if Js.Nullable.isNullable err then (
```
mnxn, line 205: This error checking is redundant since `Cmd.output` also checks if there is an error and provides more information.

src/DuneFormatter.ml
```diff
@@ -1,6 +1,7 @@
 open Import
 
-let format (document : TextDocument.t) _options _token =
+let getFormatter toolchain =
+ fun [@bs] (document : TextDocument.t) _options _token ->
```
rgrinberg, line 4: What does the `[@bs]` annotation do here?

mnxn: It guarantees that the resulting function is uncurried from the JS side, which is what `registerDocumentFormattingEditProvider` expects. It's possible that it works fine without the annotation, but I don't want to take any chances.

https://bucklescript.github.io/docs/en/function#solution-guaranteed-uncurrying

src/Extension.ml
```diff
@@ -50,52 +50,53 @@ module Instance = struct
...
+        let (_ : unit Promise.t) =
+          suggestToSetupToolchain instance duneFormatter
+        in
         (Toolchain.PackageManager.Global, true)
       | Some toolchain -> (toolchain, false)
     in
     (Toolchain.makeResources resources, isFallback)
   in
   toolchain >>= fun (toolchain, isFallback) ->
+  DuneFormatter.register duneFormatter toolchain;
```
rgrinberg, line 99: Perhaps `Instance.start` should call `DuneFormatter.register`? Otherwise, it seems a bit fragile to remember that we need to re-register the formatter everytime the toolchain is switched.

mnxn: Good point. I just updated it.

src/Extension.ml
```diff
@@ -27,27 +27,32 @@ module Server = struct
 end
 
 module Instance = struct
-  type t = LanguageClient.t option ref
+  type t =
+    { client : LanguageClient.t option ref
```
rgrinberg, line 31: Should be `mutable client` instead of the `ref`.

src/bindings/Node.ml
```diff
@@ -33,8 +33,7 @@ module JsError = struct
 
   external make : string -> t = "Error" [@@bs.new]
 
-  let ofPromiseError =
-    [%raw fun error -> "return error.message || 'Unknown error'"]
+  let ofPromiseError _error = [%raw "_error.message || 'Unknown error'"]
```
rgrinberg, line 36: Thanks for fixing the warning!

src/DuneFormatter.ml
```diff
@@ -13,14 +14,30 @@ let format (document : TextDocument.t) _options _token =
...
   let documentText = TextDocument.getText document fullDocumentRange in
 
-  Cmd.make ~cmd:(Path.ofString "dune") ()
-  |> Promise.Result.bind (fun duneCmd ->
-         Cmd.output duneCmd
-           ~args:[| "format-dune-file"  |]
-           ~stdin:documentText)
+  let cmd, args = Toolchain.getDuneFormatter toolchain in
+  Cmd.make ~cmd ()
+  |> Promise.then_ (function
```
rgrinberg, line 19: Can we use `Cmd.make ~cmd () >>= function` would read a bit cleaner. In general, we prefer the monadic API when possible

PR #209, opened 2020-05-18
Improve OCaml Syntax Highlighting

### LINKED ISSUE #14: Improve syntax highlighting for mli files
Currently, we highlight syntax elements of .ml files in the .mli. This isn't correct and .mli files are limited to the signature language. We should create a separate grammar for the signature items use it for .mli files. Then this grammar file can be embedded to the syntax of .ml files.

### LINKED ISSUE #199: Syntax highlighting for cppo directives
Such directives should be recognized in `.cppo.ml{,i}` files.

### LINKED ISSUE #207: Support .mld files
`.mld` files are standalone documentation files. They are attached directly to packages using the `(documentation)` stanza in dune. Their syntax is the same as ocamldoc comments.

mnxn:
<ul>

<li> 

Separates ocamldoc grammar from main grammar, adding support for standalone `.mld` documentation files (Closes #207). </li>

<li> 

Separates .mli specific syntax into its own grammar, preventing `.ml` syntax from being highlighted in `.mli` files or in `sig .. end` (Closes #14).

**IMPORTANT**: this introduces a new language id for `.mli` files: `ocaml.interface`. ocaml-lsp must be updated to support this language id [here](https://github.com/ocaml/ocaml-lsp/blob/6f54399a99bbc7c0eb455a568ec723529877708f/ocaml-lsp-server/src/document.ml#L24). Otherwise, features like `textDocument/formatting` do not work.
</li>

<li>

 Adds highlighting for all [Cppo directives](https://github.com/ocaml-community/cppo#user-guide) (Closes #199). </li>

<li> 

Adds highlighting for all [toplevel directives](https://caml.inria.fr/pub/docs/manual-ocaml/toplevel.html#s%3Atoplevel-directives). </li>

<li> 

Adds highlighting for all [topfind directives](http://projects.camlcity.org/projects/dl/findlib-1.8.1/doc/guide-html/x231.html). </li>

<li>

Adds highlighting for [extension float/int literals](http://caml.inria.fr/pub/docs/manual-ocaml/extensionsyntax.html#ss:extension-literals) </li>

</ul>


rgrinberg:
Nice. This change will require a corresponding update in lsp (to understand the new language identifier), so let's approach this change carefully.

rgrinberg:
Does this handle code inside attributes? For example:

```
type t [@@deriving let x = 123]
```

should still highlight the `let` inside

What about things like:
```
module S : sig
  val foo : int -> int
end = struct
end

module type S = sig
  ..
end
```

Is it possible to handle the signatures in `S` as well?

mnxn:
Yep, both of those syntax constructs are handled.

rgrinberg:
Corresponding LSP PR https://github.com/ocaml/ocaml-lsp/pull/189

rgrinberg:
I've added an `ocaml.interface` as an experimental capability [1]. Let's make sure that this field is present in this PR. If it's absent, we should just tell the user to upgrade their LSP server.

[1] https://github.com/ocaml/ocaml-lsp/pull/189/files#diff-fdb440823f9d68a0a4c8f26862ef8ef0R35

rgrinberg:
@mnxn I've tweaked the implementation a bit. The fields in `experimental` are an implementation detail that is independent of the bindings. Therefore I moved this to a module that is pat of the extension.

rgrinberg:
@mnxn thanks for that fix. I've also pushed a commit to make the bindings a little safer.

mnxn:
It doesn't look like the code is currently accessing the `ocamllsp` field of the json data, so the value is always false.

rgrinberg:
Thanks for letting me know. My json skills with bucklescript are a little weak. Will debug

rgrinberg:
How about now? I' think I fixed the json parsing.

mnxn:
Seems to work well now.

rgrinberg:
Sweet.

PR #212, opened 2020-05-18
Fix Path.parent

rgrinberg:
This broke the esy configuration in some cases for me


rgrinberg:
Thanks. I got rid of it.

PR #217, opened 2020-05-20
Improve esy sandbox discovery

rgrinberg:
* Include error messages for manifest that failed to vaildate for some reason
* Do not read arbitrary json files


imbsky:
We may want to load arbitrary JSON files with such a repository.
https://github.com/cristianoc/reanalyze

PR #218, opened 2020-05-20
Improve launch configuration

rgrinberg:
* Set outFiles correctly
* Disable other extensions when testing this extension

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,6 +15,14 @@
         "--disable-extensions",
         "--extensionDevelopmentPath=${workspaceFolder}"
       ]
+    },
+    {
+      "name": "Extension (no disable-extensions)",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "outFiles": ["${workspaceRoot}/**/*.bs.js"],
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}"]
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


mnxn:
> Disable other extensions when testing this extension

I just tested this and unfortunately it degrades the experience for editing the syntax highlighting. The theme changes to the default "Dark+" which has problems: namely that operators don't have a distinct color and that keywords are the same color as some constants. 

I think disabling extensions would be useful when editing the logic of this extension, but not for editing the syntax highlighting. 

I propose adding a _second_ configuration in addition to this one that doesn't disable extensions:

<details><summary>Expand</summary>

```
  "configurations": [
    {
      "name": "Extension",
      "type": "extensionHost",
      "request": "launch",
      "runtimeExecutable": "${execPath}",
      "outFiles": ["${workspaceRoot}/**/*.bs.js"],
      "args": [
        "--disable-extensions",
        "--extensionDevelopmentPath=${workspaceFolder}"
      ]
    },
    {
      "name": "Extension (no disable-extensions)",
      "type": "extensionHost",
      "request": "launch",
      "runtimeExecutable": "${execPath}",
      "outFiles": ["${workspaceRoot}/**/*.bs.js"],
      "args": ["--extensionDevelopmentPath=${workspaceFolder}"]
    }
  ]
```
</details>

rgrinberg:
Sure, we can have an additional configuration for testing syntax highlighting. It's just that there's enough issues from the extension not reloading when you expect it to, that I think that this should the default setting.

Feel free to push your additional configuration to this branch.

PR #220, opened 2020-05-20
Show statusbar item for current toolchain

### LINKED ISSUE #216: Display sandbox for current project
Sometimes the extension is working when I open a project but I don't know why and which sandbox it is currently using. It happened to me on a project without any .vscode/settings.json file. It would be convenient to either see in the status bar which sandbox is currently used or have a command to get this information.

mnxn:
Closes #216

Example:

example

Clicking on the status bar item will open the "select sandbox" menu.


src/Extension.ml
```diff
@@ -29,12 +29,19 @@ end
...
+    ; statusBarItem : StatusBarItem.t
     ; duneFormatter : DuneFormatter.t
     }
 
-  let create () = { client = None; duneFormatter = DuneFormatter.create () }
+  let create () =
+    let statusBarItem =
+      Window.createStatusBarItem ~alignment:StatusBarAlignment.(tToJs Left) ()
+    in
+    statusBarItem ## command #= "ocaml.select-sandbox";
```
rgrinberg, line 40: We should avoid copying this string more than once in this module. Let's create a variable for it.

src/Extension.ml
```diff
@@ -29,12 +31,19 @@ end
...
-  let create () = { client = None; duneFormatter = DuneFormatter.create () }
+  let create () =
+    let statusBarItem =
+      Window.createStatusBarItem ~alignment:StatusBarAlignment.(tToJs Left) ()
+    in
+    statusBarItem##command#=selectSandboxCommandId;
+    { client = None; statusBarItem; duneFormatter = DuneFormatter.create () }
 
   let stop t =
+    StatusBarItem.hide t.statusBarItem;
```
rgrinberg, line 46: Why is this hide and not dispose?

mnxn: I thought It would be more efficient to reuse the StatusBarItem object. Now that I think about it, it does make more sense for `Instance.stop` to completely dispose its owned resources.

Just fixed it.

PR #221, opened 2020-05-21
Simplifies esy.json for bucklescript users

prometheansacrifice:
I pushed overrides for public consumption. It should now be possible for bucklescript users to user a simpler esy.json such as this

```json
{
    "@opam/ocaml-lsp-server": "ocaml/ocaml-lsp:ocaml-lsp-server.opam",
    "@opam/reason": "*",
    "ocaml": "4.6.x"
}
```

The overrides are currently on pushed for Dune 2.5.x, but can be just as easily push for the lower ones too. A more resilient, long term solution has to do with how esy looks for secondary toolchains in general and is being discussed too.

Would really appreciate if a few folks try it out before merging. I'm open to adding this to the CI preferably in a way if overrides for new Dune releases are not present, it should just warn/notify instead of a hard failure.


rgrinberg:
Could you also update the esy setup for the repo itself? It should use the latest ocamlformat which requires dune 2.5

prometheansacrifice:
This PR needs two things

1. GCC 10 users are complaining that the secondary compiler (untouched by the overrides) doesn't build due to new defaults (
2. Windows users are unable to build the secondary compiler.

I'll be spending time on these two days this week. Soon after, we I'll update the esy.json in the project root - the CI could break (unless I'm wrong and that esy.json is not used anywhere in the CI)

prometheansacrifice:
@rgrinberg Dune 2.5.1 works fine on Windows too now. As suggested, esy.json has been updated and it pulls in the latest ocamlformat 0.14.2 (can be seen in the lock files) We can merge this I believe.

prometheansacrifice:
Not sure what's going on at the CI. Stale cache?

imbsky:
The cache isn't available in the esy install step.

PR #228, opened 2020-05-24
Improve logging and error messages

rgrinberg:
Low level information is now just logged instead of being presented to
the user.

### DIFF
--- a/src/Cmd.ml
+++ b/src/Cmd.ml
@@ -50,18 +50,29 @@ let output ~args ?cwd ?stdin { cmd; env } =
   let shellString =
     Js.Array.joinWith " " (Js.Array.concat args [| Path.toString cmd |])
   in
-  Js.Console.info shellString;
   let cwd =
     match cwd with
     | None -> None
     | Some cwd -> Some (Path.toString cwd)
   in
   ChildProcess.exec shellString ?stdin (ChildProcess.Options.make ?cwd ~env ())
-  |> Promise.map (function
-       | ChildProcess.{ exitCode = 0; stdout; stderr = _ } -> Ok stdout
-       | ChildProcess.{ exitCode; stdout = _; stderr } ->
-         Error
-           {j| Command $shellString failed:
-exitCode: $exitCode
-stderr: $stderr
-|j})
+  |> Promise.map (fun (res : ChildProcess.return) ->
+         let () =
+           let message =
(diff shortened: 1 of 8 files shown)


src/import.ml
```diff
@@ -69,3 +74,19 @@ let message kind fmt =
...
+  type field
+
+  external field : _ -> field = "%identity"
+end
+
+let log fmt = Printf.ksprintf (fun x -> Js.Console.log x) fmt
+
+let logJson msg (fields : (string * Log.field) list) =
+  let json = Js.Dict.fromList fields in
+  Js.Console.log2 msg json
```
mnxn, line 92: Would it be possible to use the native vscode output pane for logs? It seems like the more appropriate place for logs.

https://code.visualstudio.com/api/references/vscode-api#window.createOutputChannel
https://code.visualstudio.com/api/references/vscode-api#OutputChannel

src/Cmd.ml
```diff
@@ -50,18 +50,29 @@ let output ~args ?cwd ?stdin { cmd; env } =
...
-         Error
-           {j| Command $shellString failed:
-exitCode: $exitCode
-stderr: $stderr
-|j})
+  |> Promise.map (fun (res : ChildProcess.return) ->
+         let () =
+           let message = ("result", Log.field res) :: message in
+           logJson "external command" message
+         in
```
mnxn, line 73: What's the purpose of making this a let expression? Wouldn't a semicolon suffice?

src/import.ml
rgrinberg: Isn't that where the console log statements go already? Or do you think we should have a dedicated output channel?

mnxn: Yes, currently all log statements go to the developer console. I think it would be useful to have a dedicated output channel as it would make it more user-friendly.

rgrinberg:
> Yes, currently all log statements go to the developer console. I think it would be useful to have a dedicated output channel as it would make it more user-friendly.
Sounds reasonable to me. Should make it easier for us to debug stuff.

However, it’s a bit outside the scope of this PR as it requires adding bindings. This PR should at least make it easier though.

src/Cmd.ml
rgrinberg: See the new code. The purpose of the `let ()` is to avoid leaking the `message` binding.

PR #229, opened 2020-05-24
Add support of 4.09+ multi lines errors in problem matcher

Khady:
Since ocaml 4.09, when an error covers multiple lines, the message
emitted by the compiler looks like this:

```
File "robustmatch.ml", lines 33-37, characters 6-23:
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
(AB, MAB, A)
```

This commit detects the presence of an eventual end line. Support of
previous version of the compiler or single line errors is preserved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add .mld syntax highlighting
 - Add highlighting for Cppo directives in OCaml files
 - Add highlighting for more toplevel and topfind directives in OCaml files
+- OCaml problem matcher now understands multi line errors emitted by 4.09 (#229)
 
 ## 0.6.1
 
--- a/package.json
+++ b/package.json
@@ -78,11 +78,12 @@
         ],
         "pattern": [
           {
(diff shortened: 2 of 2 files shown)


mnxn:
Works well for me.
Thank you for the contribution!

Khady:
I can't reproduce the error on the CI. Not sure why it happens.

mnxn:
I restarted the workflows, lets see if it works. 

The error is `There should only be one diagnostic in the Reason source file`, which pops up in the CI every once in a while. I'm personally not sure why it happens, but it's not related to this change.

rgrinberg:
It would be nice to somehow to unit test stuff like this. @mnxn do you know if it's possible?

mnxn:
> It would be nice to somehow to unit test stuff like this. @mnxn do you know if it's possible?

Unfortunately I'm not familiar with a way to do that.

rgrinberg:
Thank you for your contribution

PR #230, opened 2020-05-24
Improve dispose handling

rgrinberg:
* Make command registration return disposables
* Allow to attach disposablas to the extension context


rgrinberg:
We'll also need to make sure that disposing of the extension disposes of the dune formatters.

mnxn:
> We'll also need to make sure that disposing of the extension disposes of the dune formatters.

This can be done by calling `Instance.stop` in a disposable. I can implement that right now.

PR #232, opened 2020-05-24
Stop Instance on extension dispose

mnxn:
https://github.com/ocamllabs/vscode-ocaml-platform/pull/230#issuecomment-633185813

### DIFF
--- a/src/Extension.ml
+++ b/src/Extension.ml
@@ -83,6 +83,8 @@ module Instance = struct
          available in mli sources. Please update to a recent version and \
          restart the server.";
     Ok ()
+
+  let disposable t = Disposable.create ~dispose:(fun () -> stop t)
 end
 
 let selectSandbox (instance : Instance.t) () =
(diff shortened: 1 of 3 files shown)


src/Extension.ml
```diff
@@ -116,6 +116,8 @@ let activate (extension : Vscode.ExtensionContext.t) =
   Vscode.ExtensionContext.subscribe extension
     (Vscode.Commands.register ~command:selectSandboxCommandId
        ~handler:(selectSandbox instance));
+  Vscode.ExtensionContext.subscribe extension
+    (Vscode.Disposable.create ~dispose:(fun () -> Instance.stop instance));
```
rgrinberg, line 120: Could you add an intermediate function to `Instance`:

```ocaml
val disposable : Instance.t -> Disposable.t
```

If we follow this convention, it should be easier for us to remember to dispose of things a bit more diligently.

mnxn: Makes sense. Just updated.

PR #238, opened 2020-05-28
slightly simplify vscode bindings

tatchi:
Some of the functions/properties there return `undefined` if no values. Bucklescript compiles `None` option type to `undefined` (and vice-versa). So you can just type their return values as `option(t)` and bucklescript will do the conversion automatically.

See: https://bucklescript.github.io/docs/en/null-undefined-option#interoperate-with-javascript-undefined-and-null

`Js.Nullable.t` should be used when the value can be `undefined` OR `null`. For functions that can return `undefined` or `null`, a `[@bs.return nullable]` directive exists to convert it to `option` : https://bucklescript.github.io/docs/en/return-value-wrapping

### DIFF
--- a/src/bindings/Vscode.mli
+++ b/src/bindings/Vscode.mli
@@ -211,9 +211,9 @@ module Workspace : sig
 
   type cancellationToken = { isCancellationRequested : bool }
 
-  val name : unit -> string option
+  val name : string option
 
-  val workspaceFolders : unit -> Folder.t array
+  val workspaceFolders : Folder.t array option
 
   val onDidOpenTextDocument : (TextDocument.event -> unit) -> unit
 
--- a/src/Toolchain.ml
+++ b/src/Toolchain.ml
@@ -249,7 +249,9 @@ let makeResources kind = kind
 
 let selectAndSave () =
   let open Promise.O in
-  let workspaceFolders = Vscode.Workspace.workspaceFolders () in
+  let workspaceFolders =
+    Vscode.Workspace.workspaceFolders |. Belt.Option.getWithDefault [||]
+  in
   sandboxCandidates ~workspaceFolders >>= fun candidates ->
   selectPackageManager candidates >>| function
   | None -> None
--- a/src/bindings/Vscode.ml
+++ b/src/bindings/Vscode.ml
@@ -74,9 +74,7 @@ end
 module WorkspaceConfiguration = struct
   type t
 
-  external get' : t -> string -> Js.Json.t Js.Nullable.t = "get" [@@bs.send]
-
-  let get workspaceConfig key = get' workspaceConfig key |> Js.Nullable.toOption
+  external get : t -> string -> Js.Json.t option = "get" [@@bs.send]
 
   type configurationTarget =
     | Global [@bs.as 1]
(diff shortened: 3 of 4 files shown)


src/bindings/Vscode.mli
```diff
@@ -211,9 +211,9 @@ module Workspace : sig
 
   type cancellationToken = { isCancellationRequested : bool }
 
-  val name : unit -> string option
+  val name : string option
 
-  val workspaceFolders : unit -> Folder.t array
+  val workspaceFolders : Folder.t array option
```
rgrinberg, line 216: Why did you change the type of this function?

tatchi: Hmm good question. I guess mainly to avoid exposing a function which wraps it and returns an empty array if `undefined` or `None`; but lets the caller decides how to handle the `None` case. I can totally revert that if you prefer the previous version :)

rgrinberg:
Is there a semantic difference between Some [| |] and None? If there is, then our bindings should retain this distinction. Otherwise, we are just complicating things.

tatchi:
> Is there a semantic difference between Some [| |] and None? If there is, then our bindings should retain this distinction. Otherwise, we are just complicating things.
> […](#)
> On May 29, 2020, 9:36 AM -0700, Corentin Leruth ***@***.***>, wrote: @tatchi commented on this pull request. In src/bindings/Vscode.mli: > - val workspaceFolders : unit -> Folder.t array + val workspaceFolders : Folder.t array option Hmm good question. I guess mainly to avoid exposing a function which wraps it and returns an empty array if undefined or None; but lets the caller decides how to handle the None case. I can totally revert that if you prefer the previous version :) — You are receiving this because your review was requested. Reply to this email directly, view it on GitHub, or unsubscribe.

Good point. I would tend to think that there's a difference since the JS API make a distinction between both cases. But actually looking at the doc:

>List of workspace folders or undefined when no folder is open. Note that the first entry corresponds to the value of rootPath.

From my understanding, it won't be possible to get an empty list. If there is no open folder, `undefined` is returned. So I guess the best would be to revert my change and return an empty list in the bindings when `undefined` is returned from the JS side. What do you think ?

rgrinberg:
> From my understanding, it won't be possible to get an empty list. If there is no open folder, undefined is returned. So I guess the best would be to revert my change and return an empty list in the bindings when undefined is returned from the JS side. What do you think ?

That would make it more straightforward for callers. I agree that it would be an improvement.

tatchi:
@rgrinberg thanks for the review, I made `workspaceFolders ` always return an array.

src/bindings/Vscode.mli
```diff
@@ -211,7 +211,7 @@ module Workspace : sig
 
   type cancellationToken = { isCancellationRequested : bool }
 
-  val name : unit -> string option
+  val name : string option
```
rgrinberg, line 214: As far as I understand, `name` is a mutable field. That is, we cannot guarantee that:

```
let x = Workspace.name in
let y = Workspace.name in
assert (x = y)
```

Since ocaml's strings are immutable (with safe string), this means that exposing the binding as `val name : string option` exposes a mutable value as an immutable one. This seemed like a bad idea to me, so I added the `unit` argument so that callers are not confused. What do you think about this?

rgrinberg: This was resolved in private.

rgrinberg:
Thanks for your contribution!

PR #239, opened 2020-05-28
Add taskProvider bindings

Khady:
I want to write a task provider for dune. As a first step here are some bindings to the task provider api.
https://code.visualstudio.com/api/references/vscode-api#TaskProvider

In Task, I don't know what Workspace and Global refer to. So I only handle a workspace folder for now. But I still use a variant and bs.unwrap to mark that the api is incomplete here.
https://code.visualstudio.com/api/references/vscode-api#Task

Also I didn't bind the custom execution. Only shell and process execution are supported
https://code.visualstudio.com/api/references/vscode-api#CustomExecution

If the bindings can get merged I will then provide a second PR with the dune provider itself.

### DIFF
--- a/src/bindings/Vscode.ml
+++ b/src/bindings/Vscode.ml
@@ -133,6 +133,14 @@ module Workspace = struct
   external getConfiguration : string -> WorkspaceConfiguration.t
     = "getConfiguration"
     [@@bs.module "vscode"] [@@bs.scope "workspace"]
+
+  external findFiles :
+       inc:string
+    -> excl:string option
+    -> maxResults:int option
+    -> cancellationToken option
+    -> TextDocument.uri array Js.Promise.t = "findFiles"
+    [@@bs.module "vscode"] [@@bs.scope "workspace"]
 end
 
 module StatusBarAlignment = struct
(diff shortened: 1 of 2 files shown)


src/bindings/Vscode.ml
```diff
@@ -133,6 +133,14 @@ module Workspace = struct
   external getConfiguration : string -> WorkspaceConfiguration.t
     = "getConfiguration"
     [@@bs.module "vscode"] [@@bs.scope "workspace"]
+
+  external findFiles :
+       inc:string
+    -> excl:string option
+    -> int option
```
rgrinberg, line 140: Is it possible to add a label to the int argument? It's not so easy to understand the purpose of it just by reading the signature.

src/bindings/Vscode.ml
```diff
@@ -407,3 +415,83 @@ module LanguageClient = struct
     let open Promise.O in
     onReady t >>| fun () -> t.initializeResult
 end
+
+module ShellExecution = struct
+  type shellQuotingOptions
```
rgrinberg, line 420: How do we create values of this type?

src/bindings/Vscode.mli
```diff
@@ -306,3 +313,74 @@ module LanguageClient : sig
...
+
+  type taskDefinition = { type_ : string [@bs.as "type"] }
+
+  val make :
+       taskDefinition:taskDefinition
+    -> scope:[ `Folder of Folder.t ]
+    -> name:string
+    -> source:string
+    -> execution:[ `Shell of ShellExecution.t | `Process of ProcessExecution.t ]
+    -> problemMatchers:string list option
```
rgrinberg, line 368: Is there a difference between `Some []` and `None` here?

src/bindings/Vscode.ml
Khady: for now we don't, I didn't bind this part

src/bindings/Vscode.mli
Khady: I think it should be the same, will remove the option

Khady:
Sorry I was sure that I pushed my work last week. Should be good now

rgrinberg:
Thanks!

PR #246, opened 2020-06-05
Update esy.lock

Khady:
Necessary to install ocamlformat 0.14.2


imbsky:
CI failed for some reason. I re-triggered it, but I don't know if it will succeed.

Khady:
Sorry it seems that I didn't proceed correctly. Here is what I did now:
```
rm -r esy.lock
esy
git add -f esy.lock
git commit
```

The `-f` during `git add` is necessary because `esy.lock` is gitignored. Without it the new packages are missing.

PR #248, opened 2020-06-07
Dune task provider

Khady:
Basic task provider for dune. This task provider doesn't try in any way to parse the dune file and detect where executables/rules/tests are defined. It simply creates one `dune build` task per dune{,-project,-workspace} file that it finds in the workspace.

In my simple example projects the tasks look like this when executing `Run Build Task`:
image

I'm open to a different naming scheme. For now the path displayed is the path of the dune file relative to the root of the project.

The PR contains a few fixes of the taskProvider bindings that could be extracted before this PR is accepted.

It is not clear to me how the provider must be enabled and disposed from the extension. I tried to copy something existing. But the sample extension does something different https://github.com/microsoft/vscode-extension-samples/blob/3f794b7a5a4b5c98b14c71b174b68888e0d24440/task-provider-sample/src/extension.ts#L24


imbsky:
Re-triggered CI. Since enabling the cache it has become unstable and I’m considering removing the cache steps.

src/TaskProviderDune.ml
```diff
@@ -0,0 +1,95 @@
...
+  let open Promise.O in
+  let folders = Workspace.workspaceFolders () in
+  let excl =
+    (* ignoring dune files from _build, _opam, _esy *)
+    Some "{**/_*,}"
+  in
+  let inc = "**/{dune,dune-project,dune-workspace}" in
+  Workspace.findFiles ~inc ~excl ~maxResults:None cancellationToken
+  >>= fun dunes ->
+  commandLine () >>= fun commandLine ->
```
rgrinberg, line 57: I think this should be simpler with just map `(>>|)`.

src/TaskProviderDune.ml
```diff
@@ -0,0 +1,95 @@
...
+      match acc with
+      | Some _ -> acc
+      | None -> (
+        match startsWith ~s:file ~prefix:folder.uri.fsPath with
+        | false -> acc
+        | true -> Some (folder, removePrefix ~s:file ~prefix:folder.uri.fsPath)
+        ))
+    None folders
+
+let commandLine () =
```
rgrinberg, line 37: Perhaps we should assume that:

* the dune binary exists 
* there's a dune-project/dune-workspace in the workspace

before providing these tasks.

Khady: > the dune binary exists

I guess I can, but not sure to see the benefit of it

> there's a dune-project/dune-workspace in the workspace

if I take the example of ahrefs, we have a mono repo with a single dune-project file at the root and multiple projects as sub folders. The rational being to share some config between all projects (like compilation flags). I think there are some people who open one instance of vscode per project. In this case there is no dune-project/dune-workspace in the workspace.

rgrinberg: > I guess I can, but not sure to see the benefit of it

What's the current behavior if dune is not installed?

> I think there are some people who open one instance of vscode per project. In this case there is no dune-project/dune-workspace in the workspace.

You mean at the workspace root, right? They ought to have to dune-project files in the subdirs. Okay, this isn't essential.

Khady: > What's the current behavior if dune is not installed?

The build tasks are listed. There would be an error only if one tries to run a task. There are a few task providers in vscode itself. In [gulp](https://github.com/microsoft/vscode/blob/ea0880611f706e0529f89aa25d1989363ab53b11/extensions/gulp/src/main.ts) and [npm](https://github.com/microsoft/vscode/blob/e77c58dd0b6d955c331a3c11321eccae710ecb85/extensions/npm/src/tasks.ts) they don't seem to check for the binary.

To me it seems more natural to have it this way rather than silently doing nothing.

For the record this is how it looks like when one executes `Run Task` in a project with a build tasks configured by hand. The first line is for a task I configured by hand. The ones from an extension are under a directory. Which may or may not by empty.
image

If I have a build task configured by hand in .vscode/tasks.json, then the tasks from the dune provider are not displayed when executing `Run Build Task`
image

It seems that the noise is pretty low.

> You mean at the workspace root, right?

No dune-project/workspace file in the vscode workspace. With the vscode workspace being a project/a subdir if only one project is opened in vscode

Khady:
I just realized there is this message in the output pane. Not sure what should be done about it.
```
Error: there is no registered task type 'dune'. Did you miss to install an extension that provides a corresponding task provider?
```

src/TaskProviderDune.ml
Khady: actually I checked a bit more what happens with npm. With this package.json file. The build tasks appears when executing `Run Build Task`. And I don't have npm installed.

```json
{
    "scripts": {
        "demo-script": "some command",
        "test": "foo",
        "build": "bar"
    }
}
```

image

src/TaskProviderDune.ml
```diff
@@ -0,0 +1,95 @@
...
+(* the ocamlc matcher is not able to parse ocaml compiler errors unless they
+   follow the short style. *)
+let env = Js.Dict.fromList [ ("OCAML_ERROR_STYLE", "short") ]
+
+let startsWith ~s ~prefix =
+  try
+    let prefixLen = String.length prefix in
+    let start = String.sub s 0 prefixLen in
+    String.equal start prefix
+  with _ -> false
```
rgrinberg, line 18: Could you move general helpers to `Import`?

src/TaskProviderDune.ml
```diff
@@ -0,0 +1,95 @@
...
+        ))
+    None folders
+
+let commandLine () =
+  let open Promise.O in
+  Toolchain.ofSettings () >>| function
+  | None -> "dune build"
+  | Some pm ->
+    let resources = Toolchain.makeResources pm in
+    let cmd, args = Toolchain.getCommand resources "dune" [ "build" ] in
```
rgrinberg, line 43: There seems to be a lot of duplication of this "dune" string. I propose that we have 2 constants:

* One for the binary name `dune`.
* Another one for the task type `dune`.

src/TaskProviderDune.ml
```diff
@@ -0,0 +1,95 @@
...
+              ; shellQuoting = None
+              }
+          in
+          ShellExecution.make ~commandLine ~options
+        in
+        let task =
+          Task.make ~taskDefinition ~scope ~source ~name ~problemMatchers
+            ~execution:(`Shell execution)
+        in
+        task.group <- Some TaskGroup.build;
```
rgrinberg, line 84: Why can't `group` be part of the constructor?

rgrinberg:
> To me it seems more natural to have it this way rather than silently doing nothing.

Alright, I trust your opinion on this matter. I just wanted to make sure that bucklescript users aren’t confused by the dune stuff.

src/bindings/Vscode.ml
```diff
@@ -469,22 +469,52 @@ module ProcessExecution = struct
     [@@bs.module "vscode"] [@@bs.new]
 end
 
-module Task = struct
+module TaskGroup = struct
   type t
 
+  external build : t = "Build" [@@bs.module "vscode"] [@@bs.scope "TaskGroup"]
+
+  external clean : t = "clean" [@@bs.module "vscode"] [@@bs.scope "TaskGroup"]
```
rgrinberg, line 477: Is the lowercase of `"clean"` intended here?

src/TaskProviderDune.ml
Khady: it could. I followed the typescript api. Will do the change

rgrinberg:
> it could. I followed the typescript api. Will do the change

Thanks. I’d like to keep a high level style for the bindings when possible.

Khady:
Pushed a few commits which should take care of your comments.

Khady:
> Error: there is no registered task type 'dune'. Did you miss to install an extension that provides a corresponding task provider?

This was solved by contributing a task definition in package.json

Khady:
> Alright, I trust your opinion on this matter. I just wanted to make sure that bucklescript users aren’t confused by the dune stuff.

To be safe I added a setting which allows the detection to be disabled.

rgrinberg:
I tried this out, and I'm now questioning the benefit of presenting the `dune` files. I think the dune-project and dune-workspace files being presented should be sufficient. It really doesn't make any difference to run dune in a subdirectory that contains a dune file, because dune will climb to the upmost dune-{project,workspace} file. It would make more sense if we somehow narrowed the targets to that directory, but we aren't doing that at the moment so it seems redundant.

rgrinberg:
@mnxn could you try this PR out and comment on the usability? No need for a full review of the code if you're busy.

mnxn:
Seems to work fine. I agree that it's probably unnecessary to show `dune` files for tasks. It could get overwhelming for large projects.

src/Extension.ml
```diff
@@ -119,6 +119,7 @@ let activate (extension : Vscode.ExtensionContext.t) =
     (Vscode.Commands.register ~command:selectSandboxCommandId
        ~handler:(selectSandbox instance));
   Vscode.ExtensionContext.subscribe extension (Instance.disposable instance);
+  Vscode.ExtensionContext.subscribe extension (TaskProviderDune.create ());
```
mnxn, line 122: I feel this should be part of `Instance`. Similarly to how `DuneFormatter` is currently.

src/TaskProviderDune.ml
```diff
@@ -0,0 +1,104 @@
...
+
+let dune_bin = "dune"
+
+let task_type = "dune"
+
+let taskDefinition = { Task.type_ = task_type }
+
+let source = task_type
+
+let problemMatchers = [| "ocamlc" |]
```
mnxn, line 11: Shouldn't this be `$ocamlc`? [The docs](https://code.visualstudio.com/api/references/vscode-api#Task) use '$tsc' and '$eslint' as examples.

Khady:
> I tried this out, and I'm now questioning the benefit of presenting the dune files. I think the dune-project and dune-workspace files being presented should be sufficient. It really doesn't make any difference to run dune in a subdirectory that contains a dune file, because dune will climb to the upmost dune-{project,workspace} file. It would make more sense if we somehow narrowed the targets to that directory, but we aren't doing that at the moment so it seems redundant.

I don't understand. If I run `dune build` or `dune build @all` in a subdir, it only builds the content of this dir for me. It makes a huge difference.

src/TaskProviderDune.ml
Khady: https://github.com/Khady/vscode-ocaml-platform/blob/ef120ec9810dec8ec222c689e16873e8990473aa/package.json#L80

The name must be the one declared by the extension. Also it seems that the names starting with a $ are supposed to be reserved for extensions that are part of vscode.

Khady:
> It could get overwhelming for large projects.

The whole list is displayed only until once task is used. Listing only the dune-project/workspace files would be totally useless at ahrefs.

src/TaskProviderDune.ml
mnxn: > Also it seems that the names starting with a $ are supposed to be reserved for extensions that are part of vscode.

I don't think so. When I use the task provider as-is, I get no problems reported, but it works fine if I add the dollar sign.

 Also if I do `Tasks: Configure Task` for `"dune: build \dune-project`, it adds the following to my tasks.json:

```json
{
	"type": "dune",
	"problemMatcher": [
		"ocamlc"
	],
	"group": "build",
	"label": "dune: build \\dune-project"
}
```

But there is an error:
> Value is not accepted. Valid values: "$msCompile", "$lessCompile", "$gulp-tsc", "$jshint", "$jshint-stylish", "$eslint-compact", "$eslint-stylish", "$go", "$lessc", "$node-sass", "$tsc", "$tsc-watch", "**$ocamlc**".

mnxn:
Sorry if this is nitpicky, but is there a reason it's named `TaskProviderDune` instead of `DuneTaskProvider`? It would be similar to `DuneFormatter` the second way.

Khady:
I wasn't sure which name to pick. Now changed to DuneTaskProvider.

src/DuneTaskProvider.ml
```diff
@@ -0,0 +1,108 @@
...
+    Js.Array.joinWith " " (Js.Array.concat args [| Path.toString cmd |])
+
+let computeTasks cancellationToken =
+  let open Promise.O in
+  let folders = Workspace.workspaceFolders () in
+  let excl =
+    (* ignoring dune files from _build, _opam, _esy *)
+    Some "{**/_*}"
+  in
+  let inc = "**/{dune,dune-project,dune-workspace}" in
```
Khady, line 61: this could be built based on a configuration setting if you want to have a way to only pick the dune-project/workspace files

package.json
```diff
@@ -50,6 +50,15 @@
           "type": "boolean",
           "default": true,
           "description": "Controls whether dune tasks should be automatically detected."
+        },
+        "ocaml.dune.files": {
+          "scope": "resource",
+          "type": "array",
```
Khady, line 56: let me know if instead of an array you would prefer an object like `{dune: false, dune-workspace: true, dune-project: true}`

rgrinberg:
Looks good. Just add a CHANGES entry and remove the extra configuration that will be removed anyway. If users complain ask for it, we can consider restoring.

Khady:
Done. Thanks to both of you

PR #259, opened 2020-06-15
Add commands to open terminals

### LINKED ISSUE #219: Launch terminal in a toolchain
vscode has a nice API for creating terminals. It would be nice to introduce a command that lunches a terminal in a toolchain. See this page https://code.visualstudio.com/api/references/vscode-api#Terminal search for the `createTerminal` function

mnxn:
Closes #219 
Sorry for the long turnaround. 

The commands use the configured extension settings (or global integrated terminal settings) to decide which shell to use and launch it with the current sandbox or a selected sandbox.


rgrinberg:
I tried this out and it looks good. A couple of comments:

- I'm wondering if we need the `TerminalSandbox.t` to be a manager of terminals. It seems simpler to just have it represent a single terminal instance and the extension can manage them.

- Should we close the terminals when the extension is reset? Seems like it's safer to let the user close them himself.

- Currently, the launched terminal doesn't work well for me because it's launched with a login shell: `zsh -l`. It would be nice if there was a way to launch a custom command in a sandbox.

- It would also be nice if there was a variant of the command that let you picked the sandbox first. It's a bit odd to have to switch to a sandbox before launching a terminal.

Otherwise, this looks really good! Thanks.

mnxn:
I believe I have addressed all 4 of those comments. Please let me know if there is anything else to change.

src/DuneTaskProvider.ml
```diff
@@ -90,19 +83,24 @@ let computeTasks cancellationToken =
...
  fun [@bs] task _cancellationToken -> Js.Promise.resolve (Some task)
 
-let create () =
-  let provider = { TaskProvider.provideTasks; resolveTask } in
-  Tasks.registerTaskProvider ~typ:task_type ~provider
+let create () = ref None
+
+let register t toolchain =
+  let provideTasks = provideTasks toolchain in
+  let provider = TaskProvider.{ provideTasks; resolveTask } in
```
rgrinberg, line 101: Slightly better is `{ TaskProvider.provideTasks ; relativeTasks }`. When it's possible to avoid an open, we should do it.

src/Toolchain.ml
```diff
@@ -247,44 +249,42 @@ let setupToolChain (kind : PackageManager.t) =
...
-  let dune = "dune" in
-  let format = "format-dune-file" in
-  match t with
-  | Opam (opam, switch) -> Opam.exec opam ~switch ~args:[| dune; format |]
-  | Esy (esy, manifest) -> Esy.exec esy ~manifest ~args:[| dune; format |]
-  | Global -> (Path.ofString dune, [| format |])
+let selectAndSave () =
+  let open Promise.Option.O in
+  select () >>| fun packageManager ->
+  let (_ : unit Promise.t) = toSettings packageManager in
```
rgrinberg, line 267: Why don't we wait for the save to finish here?

mnxn: I don't believe the previous version waited for save to finish either.

The function already returns a promise, so this should be an easy fix.

src/import.ml
```diff
@@ -25,13 +25,48 @@ let getSubDict dict key =
...
+  let ( >>= ) = flatMap
+
+  let ( >>| ) = map
+
+  let alternative l r =
+    match (l, r) with
+    | None, r -> r
+    | l, _ -> l
+
+  let flatten = function
```
rgrinberg, line 40: I think the normal name here is `join` rather than `flatten`.

src/Extension.ml
```diff
@@ -87,6 +94,15 @@ module Instance = struct
          restart the server.";
     Ok ()
 
+  let openTerminal ?toolchain t () =
```
rgrinberg, line 97: `()` isn't necessary because we have a positional argument `t` here.

src/Toolchain.ml
rgrinberg: I see, well it's worth fixing as it's confusing otherwise.

src/Extension.ml
rgrinberg: Oh, it's because it's registered as a command later. Never mind

mnxn: The `()` is needed for it to be a `unit -> unit` function for when it is used as a handler for `Vscode.Commands.register`.

I can remove the `()`, but that will require replacing its `registerCommands` entry with a lambda.

rgrinberg:
I removed alternative because I find error prone and not very useful in general. It's much more useful in a lazy language like haskell where the 2nd argument isn't evaluated.

One last thing I'm wondering about is the necessity of having per platform options. Usually, a user just customizes things for the platform that they're on. Are there other examples of per platform customization in vscode?

Apart from that, the PR looks good.

mnxn:
> One last thing I'm wondering about is the necessity of having per platform options. Usually, a user just customizes things for the platform that they're on. 

I thought the same, but I'm following the conventions of the VSCode settings themselves. VSCode uses `terminal.integrated.shell.(linux|osx|windows)` and `terminal.integrated.shellArgs.(linux|osx|windows)` for its terminal. 

I think it's best to follow the VSCode style in the case that a user has separate settings for `terminal.integrated` and they do not wish to override the configuration for all operating systems.

PR #264, opened 2020-06-23
ocaml.trace.server configuration option.

mnxn:
Adds a configuration option, `ocaml.trace.server`, to control the verbosity of the language server logs. 

Addresses https://github.com/ocaml/ocaml-lsp/pull/181#issuecomment-647770104

I put the default option to `messages`, but it appears that the previous behavior was equivalent to `off`. I believe that the `messages` option is more useful as default behavior because it shows what type of message the language server is handling instead of just the time taken.

### DIFF
--- a/src/Extension.ml
+++ b/src/Extension.ml
@@ -28,37 +28,6 @@ module Server = struct
     { command = Path.toString command; args; options = { env = Process.env } }
 end
 
-module TraceSetting = struct
-  type t = LanguageClient.trace =
-    | Off
-    | Messages
-    | Verbose
-
-  let of_string = function
-    | "off" -> Some Off
-    | "messages" -> Some Messages
-    | "verbose" -> Some Verbose
-    | _ -> None
-
-  let ofJson json =
-    let open Json.Decode in
-    match of_string (string json) with
-    | Some s -> s
-    | None ->
-      raise (DecodeError "off | messages | verbose are the only valid values")
-
-  let to_string = function
-    | Off -> "off"
-    | Messages -> "messages"
-    | Verbose -> "verbose"
-
-  let toJson (t : t) =
-    let open Json.Encode in
-    string (to_string t)
-
-  let t = Settings.create ~scope:Workspace ~key:"trace.server" ~ofJson ~toJson
-end
-
 module Instance = struct
   type t =
     { mutable toolchain : Toolchain.resources option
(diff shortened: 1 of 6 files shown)


rgrinberg:
> I put the default option to messages, but it appears that the previous behavior was equivalent to off. I believe that the messages option is more useful as default behavior because it shows what type of message the language server is handling instead of just the time taken.

IMO it should be left as `off`. Unless this information is somehow actionable to the user, we shouldn't present it by default.

src/bindings/Vscode.ml
```diff
@@ -466,15 +466,20 @@ module LanguageClient = struct
...
+    | Messages [@bs.as 1]
+    | Verbose [@bs.as 2]
+  [@@bs.deriving { jsConverter = newType }]
+
   type t =
-    { start : (unit -> unit[@bs])
-    ; stop : (unit -> unit[@bs])
-    ; initializeResult : InitializeResult.t
+    { initializeResult : InitializeResult.t
+    ; mutable trace : abs_trace
```
rgrinberg, line 477: What is the point of saving the `trace` here? As far as I know, the language client is going to pass it automatically to the server anyways. This is the code that I'm referring to:

https://github.com/microsoft/vscode-languageserver-node/blob/master/client/src/common/client.ts#L3249-L3269

Let me know if I'm misunderstanding something.

src/Extension.ml
```diff
@@ -28,6 +28,37 @@ module Server = struct
     { command = Path.toString command; args; options = { env = Process.env } }
 end
 
+module TraceSetting = struct
```
rgrinberg, line 31: As mentioned earlier, I think this option is already handled by the client implementation. So we don't actually need to do any of this.

src/bindings/Vscode.ml
mnxn: I believe I was the one that misunderstood. I wrote this code under the assumption that [LanguageClient.trace](https://github.com/microsoft/vscode-languageserver-node/blob/master/client/src/common/client.ts#L2772) was the only way to change the trace value. 

I guess I really overthought it :/.
Once I get back to my dev computer, I will remove the redundant code.

Is this behavior documented somewhere? It would be nice to know in the future.

rgrinberg: The documentation is lacking unfortunately. I learned about this from the official LSP sample https://github.com/microsoft/vscode-extension-samples/blob/master/lsp-sample/client/package.json

giltho:
For completeness, I believe that the option is automatically registered by vscode-language-server, and set by default to 'off'. That being said, adding it here will also make it explicit in the vscode-ocaml-platform documentation which is good

PR #265, opened 2020-06-23
Restart command

### LINKED ISSUE #224: Command to restart LSP server
This is needed often enough that it would be useful if there was an a command to do this.

mnxn:
Closes #224

Current behavior is to prompt the user to select a toolchain to start if no language server is running.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,11 @@
         "title": "Select sandbox for this workspace"
       },
       {
+        "command": "ocaml.restart",
+        "category": "OCaml",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -39,6 +39,11 @@
         "category": "OCaml",
         "title": "Select sandbox for this workspace"
       },
+      {
+        "command": "ocaml.restart",
```
rgrinberg, line 43: How about `ocaml.server.restart`?

PR #266, opened 2020-06-23
Unify output channels for logging

### LINKED ISSUE #231: Redirect logs to own channel
Would it be possible to use the native vscode output pane for logs? It seems like the more appropriate place for logs.

https://code.visualstudio.com/api/references/vscode-api#window.createOutputChannel
https://code.visualstudio.com/api/references/vscode-api#OutputChannel

_Originally posted by @mnxn in https://github.com/ocamllabs/vscode-ocaml-platform/pull/228#discussion_r429597805_

mnxn:
Closes #231 

The `log` function now outputs to the *OCaml Extension* output channel.

All `LanguageClient` instances now share a single *OCaml Language Server* output channel, to avoid adding duplicate channels. Before, there were many duplicate channels with 1 or 2 log statements, but only one output channel was being used for all of the log statements.

| Before |
| - |
| many channels |

These changes should make it easier for users and developers to understand what the extension and language server are doing.

### DIFF
--- a/src/Output.ml
+++ b/src/Output.ml
@@ -1,0 +1,4 @@
+let languageServerOutputChannel =
+  Vscode.Window.createOutputChannel "OCaml Language Server"
+
+let extensionOutputChannel = Vscode.Window.createOutputChannel "OCaml Extension"
--- a/src/Output.ml
+++ b/src/Output.ml
@@ -1,4 +1,4 @@
 let languageServerOutputChannel =
   Vscode.Window.createOutputChannel "OCaml Language Server"
 
-let extensionOutputChannel = Vscode.Window.createOutputChannel "OCaml Extension"
+let extensionOutputChannel = Vscode.Window.createOutputChannel "OCaml Platform Extension"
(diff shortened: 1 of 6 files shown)


src/Output.ml
```diff
@@ -0,0 +1,4 @@
+let languageServerOutputChannel =
+  Vscode.Window.createOutputChannel "OCaml Language Server"
+
+let extensionOutputChannel = Vscode.Window.createOutputChannel "OCaml Extension"
```
rgrinberg, line 4: Let's name this `OCaml Platform Extension`.

src/Output.mli
```diff
@@ -0,0 +1,5 @@
+(** [languageServerOutputChannel] is the output channel that should be used by all ocamllsp LanguageClients *)
+val languageServerOutputChannel : Vscode.Window.OutputChannel.t
+
+(** [extensionOutputChannel] is the output channel that should be used for logs by the extension *)
+val extensionOutputChannel : Vscode.Window.OutputChannel.t
```
rgrinberg, line 5: Can we make both values in this module lazy? It's better to avoid module side effects when possible.

PR #270, opened 2020-06-24
Implement command to switch between implementation and interface files.

### LINKED ISSUE #119: Add ml/mli switching
Ideally, this should be implemented on the LSP side. However, it's quiet useful to also have it when there's no LSP server installed.

mnxn:
Closes #119 

Adds the `ocaml.switch-impl-intf` command to switch between implementation and interface files.

The command can be used from a keybind, Alt+O, when a suitable file is open.
The command is also added to the `editor/title` `navigation` menu for easy access on those files.

There is a fallback mechanism implemented for when the language server cannot provide a file to switch to. The fallback mechanism considers `.ml`, `.mll`, `.mly`, and `.re` files to be implementations and it considers `.mli` and `.rei` to be interfaces. The mechanism searches for the first opposite file in the same folder to switch to.

The extension warns the user if the language server does not provide `handleSwitchImplIntf` as an experimental functionality. For the language server version of the mechanism to be used, the language server must respond to a `ocaml/didSwitchImplIntf` message that is currently unimplemented on the lsp side.


package.json
```diff
@@ -53,8 +55,41 @@
...
+        "key": "Alt+O",
+        "when": "resourceExtname == .mli || resourceExtname == .rei"
+      }
+    ],
+    "menus": {
+      "editor/title": [
+        {
+          "command": "ocaml.switch-impl-intf",
+          "key": "Alt+O",
+          "when": "resourceExtname == .ml || resourceExtname == .mll || resourceExtname == .mly || resourceExtname == .re",
```
rgrinberg, line 82: Why can't we use `editorLangId` instead?

mnxn: I was only testing with the fallback implementation and made this according to `Fallback.impl` and `Fallback.intf`.

It didn't occur to me at the time that users can have files that don't match those extensions. Will fix.

src/SwitchImplIntf.ml
```diff
@@ -0,0 +1,59 @@
+open Import
+
+module Lsp : sig
+  val switch : LanguageClient.t -> string -> string option Promise.t
+end = struct
+  let switch client fileName =
+    let targetFileName : string option Promise.t =
+      LanguageClient.sendRequest client "ocaml/didSwitchImplIntf"
+        (fileName : string)
+        ()
```
mnxn, line 10: If the language server will be able to support switching for file types other than the 6 listed in `Fallback`, we will need to figure out a way to send the language ID too.

I'm not completely familiar with the `vscode-languageclient` API, but I'm guessing that sending a value of type `Json.t` will work.

rgrinberg: > If the language server will be able to support switching for file types other than the 6 listed in Fallback, we will need to figure out a way to send the language ID too.

Could you elaborate on this? I'm not sure I'm getting the problem here.

> I'm not completely familiar with the vscode-languageclient API, but I'm guessing that sending a value of type Json.t will work.

That sounds right.

src/SwitchImplIntf.mli
```diff
@@ -0,0 +1,6 @@
+(** [requestSwitch] will attempt to switch between the implementation and
+	interface for a given file. If a valid language client is provided, the
+	function will send a `ocaml/didSwitchImplIntf` message to the language
+	server. If a valid language client is not provided or the language server
+	message fails, a less accurate fallback mechanism will be used. *)
+val requestSwitch : Vscode.LanguageClient.t option -> string -> unit Promise.t
```
rgrinberg, line 6: What about just passing the entire `document`? Otherwise, the `string` argument is a little opaque.

src/SwitchImplIntf.ml
```diff
@@ -0,0 +1,59 @@
...
+    in
+    targetFileName |> Promise.catch (fun _ -> Promise.return None)
```
rgrinberg, line 12: Better to annotate the type of `_` here. It's a bit dangerous to ignore a value of an unknown type.

src/SwitchImplIntf.ml
```diff
@@ -0,0 +1,59 @@
...
+  val switch : string -> string option
+end = struct
+  let impl = [ ".ml"; ".mll"; ".mly"; ".re" ]
+
+  let intf = [ ".mli"; ".rei" ]
+
+  let switchTo exts withoutExt =
+    List.find_map exts ~f:(fun ext ->
+        let targetFileName = withoutExt ^ ext in
+        if Fs.existsSync targetFileName then
```
rgrinberg, line 25: Can we use the async version? It probably doesn't matter, but I don't think it's any harder. We just need an appropriate `Promise.List.find_map`.

rgrinberg:
The current behavior will not create the `.ml` or `.mli` file unless it already exists? Is that correct? I feel like it would be better to do one of the following:

* Open an `.ml` or `.mli` file.
* Prompt the user with possible file names. Either `foo.ml`, `Foo.re`, or `foo.mli`, `Foo.rei`.

By the way, file names starting with a capital letter is a silly reason convention. We should stick to file names starting with lowercase letters.

src/SwitchImplIntf.ml
mnxn: For example, if a user requests to switch from `main.ml.in`, I imagine it would be helpful if the language server knew if it was set to the `ocaml` language ID by the user.

Passing the entire `document` is a good idea as it would give all the contextual information to the language server.

mnxn:
> The current behavior will not create the .ml or .mli file unless it already exists?

I'm not sure I understand. Why should the command create a file? Isn't the purpose to switch between existing implementation and interface files?

rgrinberg:
(closing the PR was accidental)

> I'm not sure I understand. Why should the command create a file? Isn't the purpose to switch between existing implementation and interface files?

I guess I'm just used to how it works in Emacs. If the `.ml` or `.mli` file don't exist, it will prompt you to open a new file. However, the file is only created once the user saves it.

mnxn:
I can see how that can be useful. 

The extension could also fill a newly created `.mli` file with an inferred interface (output of `ocamlc -i`). That would be a nice productivity booster.

rgrinberg:
> That would be a nice productivity booster.

Indeed. Let's leave that for the LSP server however.

src/bindings/Vscode.ml
```diff
@@ -480,8 +469,21 @@ module LanguageClient = struct
...
 
+  external start : t -> unit = "start" [@@bs.send]
+
+  external stop : t -> unit = "stop" [@@bs.send]
+
   external onReady : t -> unit Promise.t = "onReady" [@@bs.send]
 
+  external sendRequest :
+       t
+    -> string
```
rgrinberg, line 480: Let's introduce a label for the string argument and polymorphic arguments. `~meth` and `~data` to match the names in the LSP spec

src/switchImplIntf.ml
```diff
@@ -0,0 +1,60 @@
+open Import
+
+module Lsp : sig
+  val switch : LanguageClient.t -> TextDocument.t -> string option Promise.t
+end = struct
+  let switch client document =
+    let targetFileName : string option Promise.t =
+      LanguageClient.sendRequest client ~meth:"ocaml/didSwitchImplIntf"
```
rgrinberg, line 8: Let's call this `ocamllsp/` rather than `ocaml/`. It's not only for ocaml, and it's server specific so we should name it after the server.

src/switchImplIntf.ml
```diff
@@ -0,0 +1,60 @@
...
+        ~data:(document : TextDocument.t)
+        ()
+    in
+    targetFileName
+    |> Promise.catch (fun (_ : Promise.error) -> Promise.return None)
```
rgrinberg, line 13: It seems wrong to swallow this error. What's wrong with just letting it bubble up?

mnxn: Making it return None allows the Fallback version to be used in `tryFallback`

rgrinberg: Yeah, but it seems wrong to add code to work around buggy behavior in the server.

src/Extension.ml
```diff
@@ -94,6 +96,11 @@ module Instance = struct
         "ocamllsp in this toolchain is out of date, functionality will not be \
          available in mli sources. Please update to a recent version and \
          restart the server.";
+    if not (OcamlLsp.handleSwitchImplIntf ocamlLsp) then
+      message `Warn
+        "ocamllsp in this toolchain is out of date, switching between \
+         implementation and interface files will use a less accurate fallback \
+         mechanism. Please update to a recent version and restart the server.";
```
rgrinberg, line 103: We should combine these the two messages above into:

```
ocamllsp in this toolchain is out of date. The following features might be unavailable or degraded in functionality:
- switching between implementation and interface files
- functionality in mli sources
```

rgrinberg:
By the way, I just realized that we need to register `mly`, `mll` files with the server for this to work on the server side. We should probably do that first in a separate PR.

ulugbekna:
Do I understand correctly that it replaces the current file with the opposite? What do people think about also about "open in a new tab" with, say, alt+shift+o? (maybe in another PR if this one is good to go)

mnxn:
> Do I understand correctly that it replaces the current file with the opposite? What do people think about also about "open in a new tab" with, say, alt+shift+o? (maybe in another PR if this one is good to go)

The current state of this PR is that it opens the opposite file in a new tab.

src/Extension.ml
mnxn: The VSCode message boxes strip newlines, so I replaced them with commas.
> ocamllsp in this toolchain is out of date. The following features might be unavailable or degraded in functionality: switching between implementation and interface files, functionality in mli sources

ulugbekna:
Is there something stopping this from being merged?

rgrinberg:
I think we still an implementation on the server side. Would you like to try doing it?

ulugbekna:
I think the PR includes a fallback file switching mechanism, which could be used before the feature lands in ocaml-lsp. 

Sidenote: I would like to work on ocaml-lsp more-or-less full time in the second half of August after my exams. I tried to work on the side on cancellation and concurrent request handling, but those seem quite non-trivial. This is my first time dealing with async code in OCaml, especially using new libs such as Fiber, and manually doing scheduling.

rgrinberg:
> I think the PR includes a fallback file switching mechanism, which could be used before the feature lands in ocaml-lsp.

Indeed, but I don't think we should rely on it. There's a warning that introduces that the user can't act upon.

> I would like to work on ocaml-lsp more-or-less full time in the second half of August after my exams. I tried to work on the side on cancellation and concurrent request handling, but those seem quite non-trivial. This is my first time dealing with async code in OCaml, especially using new libs such as Fiber, and manually doing scheduling.

That's good. Let's talk about that over slack.

src/SwitchImplIntf.ml
ulugbekna: Since `TextDocument.t`, which is now passed to ocaml-lsp in the request, is a type specific to vs code, wouldn't it make sense to pass just the filepath, a string or an alias to a string, to keep it as general as possible?

mnxn: Yes, I think you're right.

ulugbekna: Unless we establish a certain type that all clients of ocaml-lsp should follow; something like 

```ocaml
type params =
  { uri : string
  ; languageId : string
  }
```

src/SwitchImplIntf.ml
ulugbekna: I use a similar approach by picking the first existing file to switch to. However, I'm not sure how to do this asynchronously in lsp: create a helper thread in the scheduler that we could use for small tasks such as this one or try to come up with an async io based on fibers if that's possible at all?

rgrinberg:
> However, I'm not sure how to do this asynchronously in lsp: create a helper thread in the scheduler that we could use for small tasks such as this one or try to come up with an async io based on fibers if that's possible at all?

You can do it synchronously in LSP.

rgrinberg:
> Do I understand correctly that it replaces the current file with the opposite? What do people think about also about "open in a new tab" with, say, alt+shift+o? (maybe in another PR if this one is good to go)

We should just copy how it works in C/C++. Hopefully there will be a standard LSP command for this one day.

rgrinberg:
type params =
 { uri : string; languageId : string
 }

Why do we need the language Id on the server side?

mnxn:
Is there an implementation of the LSP switching yet?

I think it would be useful to have that first before deciding on the types that are going to be passed between the client and server.

rgrinberg:
Sure, we can wait for it first. It just seems suspicious to obtain the languageId of a document this way. There’s already a mechanism for obtaining languageId’s once the document is opened.

Or is this language Id supposed to signify something else?

mnxn:
> There’s already a mechanism for obtaining languageId’s once the document is opened.

I wasn't aware of that. In that case, it's not necessary.

mnxn:
Closed in favor of #327

PR #279, opened 2020-07-06
Add ocaml.debounce.server configuration option

### LINKED ISSUE #273: Add debounce config option
https://github.com/ocaml/ocaml-lsp/pull/181/ adds debouncing support. The next step is to make this support configurable. We should add a float valued option `ocaml.server.debounce` to do this.

mnxn:
Closes #273

I think all of the logic can be implemented on the language server side with  [`workspace/configuration`](https://microsoft.github.io/language-server-protocol/specification.html#workspace_configuration) and  [`didChangeConfiguration`](https://microsoft.github.io/language-server-protocol/specification.html#workspace_didChangeConfiguration).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Stop automatically highlighting suffixed META files (#565)
 
+- Add `ocaml.debounce.server` configuration option to control the amount of time
+  the language server waits before recognizing requests
+
 ## 1.8.1
 
 - Revert automatic installation of platform tools
--- a/README.md
+++ b/README.md
@@ -91,20 +91,21 @@ esy
 This extension provides options in VSCode's configuration settings. You can find
 the settings under `File > Preferences > Settings`.
 
-| Name                               | Description                                                                                             | Default |
-| ---------------------------------- | ------------------------------------------------------------------------------------------------------- | ------- |
-| `ocaml.sandbox`                    | Determines where to find the sandbox for a given project                                                | `null`  |
-| `ocaml.dune.autoDetect`            | Controls whether dune tasks should be automatically detected.                                           | `true`  |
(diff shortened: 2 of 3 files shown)


mnxn:
I think `didChangeConfiguration` can be handled here:
https://github.com/ocaml/ocaml-lsp/blob/3b105e6ecac6f7a0df50953ce188681050236fea/ocaml-lsp-server/src/ocaml_lsp_server.ml#L690

mnxn:
It turns out that the client doesn't automatically send the didChangeConfiguration notifications, so this can't be merged until I figure that out.

mnxn:
Ok, my initial guess that all the logic can be implemented on the server side was correct, except that there are a few more steps.

The client doesn't send out `workspaces/didChangeConfiguration` notifications unless that capability is dynamically registered from the server.

So what the language server needs to do is to send a [`client/registerCapability`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#client_registerCapability) request that registers `workspaces/didChangeConfiguration`. If the server was written in JS, it would be something like this:

```js
sendRequest("client/registerCapability", {
  registrations: [{
    id: "some id that should probably be a GUID",
    method: "workspace/didChangeConfiguration",
    registerOptions: { section: "ocaml" }
  }]
});
```
I will see if I can implement this in the server myself.

rgrinberg:
I remember looking into this a little earlier and it seems like configuration change notifications are no longer recommended.

At this point I think it’s fine if the server just fetches the configuration on start up. I don’t think it’s so bad if the user just bounces the server if they change the config.

smorimoto:
This still looks helpful. Could you resolve the conflict? @mnxn

tmattio:
It doesn't seem like there's much interest in this, so I think we can close it until someone manifests an interest in having it.

PR #283, opened 2020-07-09
Remove inconsistent Esy tests

mnxn:
The `There should only be one diagnostic in the Reason source file` test seem to fail in the CI randomly. 

I don't believe we've received any issues from Reason users about problems with diagnostics so we should remove the test for now.

### DIFF
--- a/test/suite/esy.test.js
+++ b/test/suite/esy.test.js
@@ -45,26 +45,6 @@ suite("Basic tests", () => {
     );
     if (process.platform != "win32" && process.platform != "win64") {
       assert.equal(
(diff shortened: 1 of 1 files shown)


rgrinberg:
CI still failing:

```
  1) Basic tests
       Esy:

      AssertionError [ERR_ASSERTION]: There should only be one diagnostic in the OCaml source file
      + expected - actual

      -0
      +1
```

mnxn:
I guess both the Reason and OCaml tests would have failed, but the failing Reason test prevented it from continuing to to the OCaml test.

I suppose I will remove both diagnostics tests.

PR #301, opened 2020-07-29
fix esy snippet to include proper dune version

fakenickels:
without pinning the right dune version users get the OCaml version error in the end as esy tries to install dune v2.6.2

```
    The ocamlfind's secondary toolchain does not seem to be correctly
    installed.
    Dune requires OCaml 4.07 or later to compile.
    Please either upgrade your compile or configure a secondary OCaml compiler
    (in opam, this can be done by installing the ocamlfind-secondary package).
    error: command failed: 'ocaml' 'bootstrap.ml' '-j' '4' (exited with 2)
    esy-build-package: exiting with errors above...
    
  building @opam/dune@opam:2.6.2
esy: exiting due to errors above
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ npm install esy --global
     "@opam/ocaml-lsp-server": "ocaml/ocaml-lsp:ocaml-lsp-server.opam",
     "@opam/reason": "*",
     "ocaml": "4.6.x"
+  },
+  "resolutions": {
+    "@opam/dune": "2.5.1"
   }
 }
 ```


prometheansacrifice:
@fakenickels The reason you ran into this is because 2.6.2 was released recently (making it the latest) for which overrides weren't present - I have just merged a PR that adds them. This shouldn't be a problem anymore.

fakenickels:
I see! Thanks for answering

PR #322, opened 2020-08-15
Add the option to use a custom sandbox with a configurable command template

### LINKED ISSUE #201: config ocamllsp in Windows WSL
I cannot specify `ocaml.lsp.path` in WSL
```
"ocaml.lsp.path": "wsl -d Ubuntu-20.04 bash -c ocamllsp"
```
I am not sure whether I don't figure out how to do it or currently it's not supported

For comparasion, I could use OCaml tools inside WSL for the another vscode extention OCaml and Reason IDE
```
"reason.path.ocamlfind": "bash -ic ocamlfind",
"reason.path.ocamlmerlin": "bash -ic ocamlmerlin",
```

mnxn:
Closes #201

Can be used like:
```json
{
	"ocaml.sandbox": {
		"kind": "custom",
		"ocamllsp": "bash -ic ocamllsp",
		"dune": "bash -ic dune"
	}
}
```

(I haven't actually tested it with WSL, but I don't see why it wouldn't work)


rgrinberg:
The idea is good, but I'm wondering if we should go with an implementation that doesn't depend on enumerating on all the tools. For example:

```
"ocaml.sandbox": {
  "kind": "custom",
  "run": "bash -ic $prog $args"
}
```

Or, we can just assume that `$prog $args` always come last.

mnxn:
Alright, I will change it to that tomorrow.

mnxn:
Just changed it to use `$prog` and `$args`.

src/Toolchain.ml
```diff
@@ -335,15 +316,15 @@ let getCommand (t : PackageManager.t) bin args : Path.t * string array =
...
-        | _ -> [ bin ]
-      in
-      match command with
-      | bin :: extraArgs -> (Path.ofString bin, Array.of_list (extraArgs @ args))
+    | Custom run -> (
+      match
+        run
+        |> Js.String.replace "$prog" bin
+        |> Js.String.replace "$args" (String.concat "" args)
+        |> String.split_on_char ' '
```
rgrinberg, line 324: I fear this will not work well with quoting :/

What about interpreting this as a shell command? So just replacing `$prog` and `$args` and then passing the entire thing to `sh` (or however you get this binary in vscode).

mnxn: I think that it's not easy to make that solution portable across operating systems. The unix shells are pretty consistent with using `-c` for executing commands, but the same can't be said about Windows: cmd.exe uses `/C` and Powershell uses something like `Invoke-Command`. Figuring out the correct option for the user's selected shell seems like it will be too error-prone - especially considering that they could be using a shell that doesn't follow the platform conventions.

Using the shells as another layer will also make the error messages about the command fails harder to read.

Although it removes some abstraction, I think the best solution is to separate the `bin` and `args` in the configuration. The previous example would look like this:
```json
{
	"ocaml.sandbox": {
		"kind": "custom",
		"bin": "bash",
		"args": "-ic $prog $args"
	}
}
```

The `bin` field could have just the `$prog` substitution available, and the `args` field would have both the `$prog` and `$args` substitutions available.

mnxn: After some more investigation, passing the entire string from the args field as a single argument isn't satisfactory either. If I use my idea, `Window.createTerminal` breaks because it inserts quotes around args. Your idea does seem better.

I also found out that Node's child_process (what we use to execute all commands) has some [shell requirements](https://nodejs.org/api/child_process.html#child_process_shell_requirements) that we can take advantage of: 
> The shell should understand the -c switch. If the shell is 'cmd.exe', it should understand the /d /s /c switches and command line parsing should be compatible.

It turns out that powershell does understand `-c`, so the only special handling necessary is for cmd.exe

I will implement what you suggested.

mnxn: This has now been implemented with a slight change of terminology:
```json
"ocaml.sandbox": {
  "kind": "custom",
  "template": "bash -ic $prog $args"
}
```

mnxn:
Now that this feature makes it possible to run commands from different environments, I'm thinking that the different settings for `ocaml.terminal.shell` and `ocaml.terminal.shellArgs` on each operating system are unnecessary and potentially confusing.

For example, a user might be using WSL to run Linux on Windows; to set the shell to ubuntu's `bash`, they would have to edit `ocaml.terminal.shell.windows` despite the fact that it's running in Linux. 

I think the configuration options should no longer be distinguished by OS.

src/Toolchain.ml
mnxn: I just realized this doesn't handle spaces in the command name very well. Still needs more work...

src/Toolchain.ml
```diff
@@ -274,25 +317,37 @@ let getCommand (t : PackageManager.t) bin args : Path.t * string array =
...
+        template
+        |> Js.String.replace "$prog" bin
+        |> Js.String.replace "$args" (String.concat " " args)
+        |> String.trim
+      in
+      let args =
+        match Path.basename (Path.ofString shell) with
+        | "cmd.exe" -> [ "/d"; "/s"; "c"; command ]
+        | _ -> [ "-c"; command ]
+      in
```
rgrinberg, line 332: Instead of launching a process by binary path and argument list, I think we should just use a `ShellExecution` for a custom sandbox.

src/Toolchain.ml
```diff
@@ -254,12 +282,27 @@ let select () =
...
-  selectPackageManager candidates >>= fun { status; packageManager } ->
-  match status with
-  | Error s ->
-    message `Warn "This toolchain is invalid. Error: %s" s;
-    Promise.return None
-  | Ok () -> Promise.Option.return packageManager
+  selectPackageManager candidates >>= function
+  | { status = Ok (); packageManager = Custom _ } ->
+    let validateInput input =
+      if Js.String.includes "$prog" input && Js.String.includes "$args" input
```
rgrinberg, line 288: Actually, vscode already has a standard syntax for this stuff: `${prog}`, `${args}`. I suggest that we just reuse it.

I'm trying to find an API in vscode that would allow us to expand stuff like this, but without any luck so far.

src/Cmd.ml
```diff
@@ -45,23 +45,30 @@ let make ?(env = Process.env) ~cmd () =
       | None -> Error {j| Command "$cmd" not found |j}
       | Some cmd -> Ok { cmd; env } )
 
-let output ~args ?cwd ?stdin { cmd; env } =
-  (* TODO use ChildProcess.spawn to get rid of this pointless concatenation *)
-  let shellString =
-    Js.Array.joinWith " " (Js.Array.concat args [| Path.toString cmd |])
+let formatArgs args =
```
rgrinberg, line 48: We only use this function for writing log entries, right?

In that case, I wonder if we should just write the args as a json list and save ourselves the hassle. I know this effort at quoting is going to fail in some edge case regardless?

mnxn: Yes, the quoting is only for logs. I will just use the json list then.

src/Toolchain.ml
mnxn: I've been working on something similar to this. But instead of reusing the vscode type, I make a variant type like
```ocaml
type command = 
    | Exec of string
    | Spawn of Path.t * string array
```
Then the custom sandbox can return Exec while the other sandboxes use Spawn. The corresponding `ChildProcess.exec` or `ChildProcess.spawn` function is chosen. I think that should make the command execution simpler.

mnxn:
Slightly off-topic question, but it seems to me like the AzurePipelines, CheckBucklescriptCompat, and Setup modules are all dead code. Is there something I'm overlooking?

smorimoto:
You are right. It remains because it could be re-used in the future, but I would not expect it to be re-used.

mnxn:
If that's the case, I'm going to remove them, because I'm rewriting the Cmd api, and some of them rely on it.

src/Toolchain.ml
rgrinberg: Sounds reasonable to me. I'd call it `Shell` rather than `Exec`.

src/Cmd.mli
```diff
@@ -1,21 +1,27 @@
 (** Interface for running system commands *)
 
-type t
+open Import
 
-val make :
-  ?env:string Js.Dict.t -> cmd:Path.t -> unit -> (t, string) result Promise.t
+type shell = [ `Shell of string ]
+
+type spawn = [ `Spawn of Path.t * string list ]
```
rgrinberg, line 7: How about we just create a spawn type `{ prog : Path.t ; args : string list }` and then use that instead of the current `spawn`. I'm not a fan of polymorphic variants and try to avoid them as much as possible.

src/Extension.ml
```diff
@@ -29,14 +29,13 @@ end
...
-      Cmd.toSpawn (Toolchain.getLspCommand toolchain)
-    in
-    Cmd.log (Spawn command);
-    { command = Path.toString bin
-    ; args = Array.of_list args
-    ; options = { env = Process.env }
-    }
+    let command = Toolchain.getLspCommand toolchain in
+    Cmd.log command;
+    let options = { LanguageClient.env = Process.env; shell = true } in
```
rgrinberg, line 34: Shouldn't this be set only for `Shell`?

src/bindings/Vscode.mli
```diff
@@ -378,7 +378,10 @@ module LanguageClient : sig
     }
   [@@bs.deriving abstract]
 
-  type processOptions = { env : string Js.Dict.t }
+  type processOptions =
+    { env : string Js.Dict.t
+    ; shell : bool
```
rgrinberg, line 383: Let's leave a comment here that this field in fact accepts the shell path as an argument as well.

src/Cmd.mli
```diff
@@ -1,19 +1,31 @@
...
+val append : spawn -> string list -> spawn
+
+val checkSpawn : spawn -> spawn Or_error.t Promise.t
 
-val binPath : t -> Path.t
+val check : t -> t Or_error.t Promise.t
+
+val toSpawn : t -> spawn
+
+val log : ?cwd:string -> ?result:ChildProcess.return -> t -> unit
```
rgrinberg, line 28: I don't think `cwd` or `result` are used anywhere. Shall we just get rid of them?

mnxn: `result` is used here https://github.com/ocamllabs/vscode-ocaml-platform/blob/a180852b1fff330b19447dcdb3d0c70354c7e91a/src/Cmd.ml#L104

`cwd` can be removed

arbipher:
I see this commit #201 is closed and the commit is merged. But I don't figure out how to config vscode to use ocaml-lsp in WSL. There are several config snippets in this commit.
May I ask about (an example of) the final config structure for the `ocaml.sandbox` ?

mnxn:
> I see this commit #201 is closed and the commit is merged. But I don't figure out how to config vscode to use ocaml-lsp in WSL. There are several config snippets in this commit.
> May I ask about (an example of) the final config structure for the `ocaml.sandbox` ?

It is like 
```json
"ocaml.sandbox": {
  "kind": "custom",
  "template": "bash -ic $prog $args"
}
```
It is also possible to select a custom toolchain by clicking on the status bar item or running the `ocaml.select-sandbox` command.

arbipher:
I currently use WSL1 so I don't know whether it makes a difference.

The good news is the OCaml Platform in vscode finally start to work for me, but I need to change it a bit (adding a wraper **`**)

```
    "ocaml.sandbox": {
        "kind": "custom",
        "template": "wsl bash -c '$prog $args'"
    }
```

then in vscode I can see the Output for OCaml Platform Extension

```
external command {
  "shell": "wsl bash -c 'ocamllsp --version'",
  "result": {
    "exitCode": 0,
    "stdout": "1.0.0-7-g0d8bd3b\n",
    "stderr": ""
  }
}

external command {
  "shell": "wsl bash -c 'ocamllsp '"
}
```

If I use the command without **`** as you posted, vscode has no output.

I think it also matches the observation I did in powershell

```
PS C:\Users\deep> wsl bash -c 'ocamllsp --version'
1.0.0-7-g0d8bd3b

PS C:\Users\deep> wsl bash -c ocamllsp --version
^C
```

The first command returns a result while the second command gets stuck.

arbipher:
I can see a error message in vscode from OCaml Language Server.

```
Thread 2 killed on uncaught exception Unix.Unix_error(12, "sleep", "")
Raised by primitive operation at Lsp__Scheduler.wake_loop.loop in file "lsp/src/scheduler.ml", line 235, characters 28-59
Called from Thread.create.(fun) in file "thread.ml", line 39, characters 8-14
```

I guess it's related to WSL1.

mnxn:
>  I need to change it a bit

The custom sandbox configuration is intended to be flexible so you can specify the format. I don't use WSL, so that was just my guess of the format.

> I can see a error message in vscode from OCaml Language Server.

> I guess it's related to WSL1.

That's a Unix_error, so that's my guess too. If you ever get the chance, you should try it on WSL2 - which should have better compatibility with Unix/Linux APIs.

rgrinberg:
This bug should be reported in the LSP repo

PR #327, opened 2020-08-18
Implement switch impl/intf with file creation support

ulugbekna:
Implements #270, based on @mnxn's work (I hope he doesn't mind me hijacking his work 🙏)

The difference between @mnxn's PR is that this PR offers the user to create a file (a dialog box with several file options) if there is no file to switch to on `Alt + o`.

Two things to consider:

1. Do we want to keep a fallback in case the lsp doesn't implement this feature? (I would need to fix it slightly)
2. Before using `Fs.writeFile`, I tried to use vs code API `WorkspaceEdit.createFile` and wrote bindings, which didn't work. We can either remove it or keep it if someone more knowledgeable about bucklescript can help fix the binding. (I would prefer keeping) 

switch_impl_intf


rgrinberg:
> Before using Fs.writeFile, I tried to use vs code API WorkspaceEdit.createFile and wrote bindings, which didn't work

The difference is that the former always saves the file on disk, right? If `WorkspaceEdit.createFile` allows the user to save the file manually, then it's much preferable. We should refrain from writing stuff to disk as much as possible.

rgrinberg:
I think the gif shows some undesirable behavior. We're offered a prompt where only a single options makes sense `a_lonely.mli`. Why don't we just select that option automatically for the user?

src/switchImplIntf.ml
```diff
@@ -0,0 +1,98 @@
...
+      when OcamlLsp.handleSwitchImplIntf capabilities ->
+      Lsp.switch client document
+    | _ ->
+      log "using fallback mechanism to switch implementation/interface";
+      Fallback.switch document.TextDocument.fileName
+  in
+
+  let open Promise.O in
+  filesToSwitchTo >>= fun arr -> match Array.to_list arr with
+  | [] -> assert false
```
mnxn, line 69: Is it not possible for the language server to return an empty array?

ulugbekna: At the moment of writing this code, it wasn't possible, but since we're [discussing](https://github.com/ocaml/ocaml-lsp/pull/233#discussion_r472937245) some behavior change on ocaml-lsp side, this is subject to change.

src/bindings/Vscode.ml
```diff
@@ -502,8 +536,25 @@ module LanguageClient = struct
...
+  external sendRequest_token :
+    t -> string -> 'a -> Workspace.cancellationToken -> 'b Promise.t
+    = "sendRequest"
+    [@@bs.send]
+
+  let sendRequest t ~meth ~data ?token () =
+    match token with
+    | None -> sendRequest t meth data
+    | Some token -> sendRequest_token t meth data token
+
```
mnxn, line 557: This section is necessary to avoid sending extra arguments to the sendRequest function which uses variable arguments.
See https://github.com/ocaml/ocaml-lsp/pull/233#discussion_r474338641

ulugbekna:
I am quite happy with this implementation: 
- if no opposing file exists, the most likely candidate is opened but not saved to disk (user has to save to keep the file) - as @rgrinberg wanted and which makes a lot of sense, of course :-)
- if a single opposing (ml vs mli, re vs rei, etc) exists, it is opened
- if several exist, the user is given a choice to pick one to open (want to make several files pickable, wdyt?)

ulugbekna:
Should I remove the fallback mechanism?

rgrinberg:
Yeah, I don't think it's so important & it's additional maintenance. Might as well.

ulugbekna:
I don't really want to deal with handling a case when the lsp server misses the capability to switch between impl and intf because 

1) we have a single impl of ocamllsp that does implement this capability
2) if the user is on prev version of ocamllsp, then they are warned that they are on outdated ocamllsp and the switch command will not work
3) if we want to really handle this case, we would need to dynamically put 'Switch implementation/interface' if the lsp has this capability only, etc.

Otherwise, I think it's ready for the final review and merging.

rgrinberg:
> I don't really want to deal with handling a case when the lsp server misses the capability to switch between impl and intf because

As I said, the fallback is indeed optional and we should focus our resources on more important features. Let's just make sure that our error messages are always clear and the user always knows when he needs to upgrade.

ulugbekna:
We can then merge and release perhaps.

src/switchImplIntf.ml
```diff
@@ -0,0 +1,66 @@
...
+    Vscode.Window.showTextDocument (`Document doc) >>| fun _ -> ()
+  in
+
+  let filesToSwitchTo =
+    match (client, capabilities) with
+    | Some client, Some capabilities
+      when OcamlLsp.handleSwitchImplIntf capabilities ->
+      switch client document
+    | _ -> failwith "NOT IMPLEMENTED"
+    (* TODO handle more gracefully *)
```
mnxn, line 31: Are we going to keep this unimplemented?

ulugbekna: I think so unless failwith causes hard crash for the user.

mnxn (approved):
Looks good to me.

src/switchImplIntf.ml
ulugbekna: another option is to remind the user to upgrade ocamllsp with a possibility to "Do not show this again" but I won't be able to get to work on this until the weekend (I would really prefer to have this merged :)))

rgrinberg: Okay, but we can't release until we hide errors like this. I don't think we necessarily need "a do not show this again" message, but every error message should be clear and make it obvious for how to fix the problem.

src/switchImplIntf.ml
```diff
@@ -0,0 +1,66 @@
...
+  in
+
+  let open Promise.O in
+  filesToSwitchTo >>= fun arr ->
+  match Array.to_list arr with
+  | [] ->
+    failwith
+      "all files must be mapped (can be switched to) to at least one file"
```
rgrinberg, line 39: Is this case possible at all?

ulugbekna: No, see the spec for the command [link](https://github.com/ulugbekna/ocaml-lsp/blob/5ba658835e2171da19d5053cca814075a0e79d13/ocaml-lsp-server/docs/ocamllsp/switchImplIntf-spec.md). How could I make this invariant more declarative in the code? I could put a comment saying such a case is impossible and `assert(false)` for example.

rgrinberg: Yes, that is preferable. `assert false` with an explanation sounds good to me.

ulugbekna:
It should be in a better shape now. Sorry for pushing its merging. Let me know what you think and where I could improve.

(I would want to clean up `Extension.ml` and `ocamlLsp.ml` but it's better be in another PR)

ulugbekna:
rebased this as well @rgrinberg

ulugbekna:
@rgrinberg sorry, I force pushed and removed latest commits; I'll fix it now -- juggling between work & personal laptops 🤷‍♂️

rgrinberg:
Yeah, you also need to `git config` your `user.name` and `user.email` for this repo. If you want all of your contributions to visible under your  account.

ulugbekna:
Thanks, config'ed. 

I don't mind my credentials wrong on these commits (I don't know how to retrofix them in commits) if that's fine with you.

ulugbekna:
would be great to merge 

@mnxn Rudi mentioned that he wants your review before merging

rgrinberg:
If Max is busy, I can review & merge tomorrow.

mnxn:
I can review in a couple of hours.

mnxn (approved):

ulugbekna:
Thanks Rudi! :)

PR #350, opened 2020-09-02
Make .ocamlformat syntax highlighting more distinct

### LINKED ISSUE #324: Separate color of keyword.operator.ocamlformat from the options for ocamlformat

mnxn:
Closes #324

The equals operator is now `punctuation.separator.key-value`, which is the same value as equal signs in `.gitconfig` or `.conf` files. Whether or not this makes the equals sign a different color from the options depends on the user's theme.

The option values are now highlighting as strings, which makes them a different color from keywords in the Dark+ theme or numbers in the file.

---

VSCode Dark+
| Before | After |
| - | - |
| before-dark | after-dark |

Monokai
| Before | After |
| - | - |
| before-monokai | after-monokai |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Highlight dashes in PKG names for .merlin files (#349)
+- Make .ocamlformat syntax highlighting more distinct (#350)
 
 ## 1.1.1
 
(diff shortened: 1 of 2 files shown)


smorimoto (approved):

smorimoto:
Awesome 👏

PR #360, opened 2020-09-05
Fix error when ocaml.sandbox is unset

### LINKED ISSUE #357: Installation on Windows works, but can't find language server
I installed OCaml and OPAM via the recommended https://fdopen.github.io/opam-repository-mingw/installation/

I installed ocaml-lsp using OPAM, which worked fine.

I installed the plugin in vscode, which appeared to work. However, when the plugin prompts and gives options for which OCaml installation to use "Select package manager", the OCaml versions don't show (unlike on OS X, where it shows "4.11.1" and you just pick it). 

How can I tell the ocaml-platform plugin where to find what it needs? Has anyone successfully used the ocaml-platform plugin on Windows?

I see there is an option "Input a custom command template", but I have no idea what to put in it.

Also, I get the additional error "setting sandbox is invalid, expected object, got null" upon first loading a .ml file.

mnxn:
Closes #357

Currently when extension activates and `ocaml.sandbox` is unset, the editor shows 2 error messages:
old

`Settings.get` was supposed to handle null and undefined gracefully, but bucklescript doesn't seem to consider `null` as `None`.

This PR changes `WorkspaceConfiguration.get` to return `Js.Nullable.t`, which can be safely converted into an `option` in order to handle nulls correctly. The extra error message no longer appears.

### DIFF
--- a/src/bindings/Vscode.mli
+++ b/src/bindings/Vscode.mli
@@ -101,7 +101,7 @@ end
 module WorkspaceConfiguration : sig
   type t
 
-  val get : t -> string -> Js.Json.t option
+  val get : t -> string -> Js.Json.t Js.Nullable.t
 
   type configurationTarget =
     | Global
--- a/src/bindings/Vscode.mli
+++ b/src/bindings/Vscode.mli
@@ -101,7 +101,7 @@ end
 module WorkspaceConfiguration : sig
   type t
 
-  val get : t -> string -> Js.Json.t Js.Nullable.t
+  val get : t -> string -> Js.Json.t option
 
   type configurationTarget =
     | Global
--- a/src/settings.ml
+++ b/src/settings.ml
@@ -11,7 +11,7 @@ let create ~scope ~key ~ofJson ~toJson = { scope; key; toJson; ofJson }
 
 let get ?section t =
   let section = Workspace.getConfiguration ?section () in
-  match Js.Nullable.toOption (WorkspaceConfiguration.get section t.key) with
+  match WorkspaceConfiguration.get section t.key with
   | None -> None
   | Some v -> (
     match t.ofJson v with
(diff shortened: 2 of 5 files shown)


src/bindings/Vscode.mli
```diff
@@ -101,7 +101,7 @@ end
 module WorkspaceConfiguration : sig
   type t
 
-  val get : t -> string -> Js.Json.t option
+  val get : t -> string -> Js.Json.t Js.Nullable.t
```
rgrinberg, line 104: How about we keep the signature but change the implementation to do the unwrapping?

I'd prefer not to leak bucklescript implementation details whenever it's possible

rgrinberg (approved):

src/settings.ml
```diff
@@ -17,9 +17,7 @@ let get ?section t =
     match t.ofJson v with
     | s -> Some s
     | exception Json.Decode.DecodeError msg ->
-      let (_ : unit Promise.t) =
-        Window.showErrorMessage (sprintf "Setting %s is invalid: %s" t.key msg)
-      in
+      message `Error "Setting %s is invalid: %s" t.key msg;
```
rgrinberg, line 20: Actually, I think this change is too drastic. How would the user know if he accidentally typod the setting for example?

How about we just ignore nulls and show the error when a legit parsing error occurs.

mnxn: It does ignore nulls. `None` (`null`) is returned from the function without showing an error message here:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/17100c6531beb97a117c45654962e09d6c596f84/src/settings.ml#L15

If the user typod the setting and it does not exist, `Settings.get` returns `None` without showing an error message, as it was originally intended. This PR is just an implementation fix because the previous version expected `null` to turn into ` None` `option` automatically, but it didn't.

All of the other functions that receive a value from `Settings.get` are already set up to handle the situation if it returns `None`, which happens when the setting doesn't exist.

rgrinberg: Oh ok. For some reason I thought `message` just displays it in the log.

rgrinberg (approved):

PR #364, opened 2020-09-09
Fix highlighting of rec keyword in type-annotated functions

mnxn:
The last function's highlighting was fixed:
| Before | After |
| - | - |
| before | after |


smorimoto (approved):

rgrinberg:
Perhaps it's worth to release `1.1.2` with all these highlighting fixes.

mnxn:
Alright, I can do that.

PR #366, opened 2020-09-14
fix missing indentation for 'val's in snippet for 'sig'

ulugbekna:


mnxn (approved):

smorimoto (approved):

ulugbekna:
I don't have merging rights, so consider merging when you have time :)

PR #370, opened 2020-09-20
Prettifies sandbox info in status bar and created terminal

ulugbekna:
Shortens the sandbox info displayed in the status bar (at the bottom of the editor) and the created terminal. This is a "prettify" PR but I'm very excited about it! :-)

It now looks like this:

For global sandbox:

image

For named opam switch (located in ~/.opam):

image

For local opam switch:

image

For esy sandbox:

image

For custom:

image

-------------

For the `Create Terminal` command:

Global:

image

Sandbox: 

image

PR #371, opened 2020-09-20
Prettify sandbox names to pick

ulugbekna:
Now it looks like this:

image
image

### DIFF
--- a/src/Toolchain.ml
+++ b/src/Toolchain.ml
@@ -200,7 +200,7 @@ module Candidate = struct
 
   let toQuickPick { packageManager; status } =
     let create = Window.QuickPickItem.create in
-    let detail =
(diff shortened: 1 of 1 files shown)


ulugbekna:
The branch is checked out from #370 so start review from [68092f6](https://github.com/ocamllabs/vscode-ocaml-platform/pull/371/commits/68092f68b46e6a8184185757f310901329d38b35). 

The two PRs also share literally a couple of lines of code but moving those out seemed like overengineering -- I would rather rewrite some of the stuff later when we have a more precise understanding of things we want from the extension.

rgrinberg:
@ulugbekna could you rebase?

ulugbekna:
@rgrinberg sure

PR #372, opened 2020-09-20
Fix esy and ocamlformat version in esy.json

ulugbekna:


Makefile
```diff
@@ -6,7 +6,7 @@ build:
 .PHONY: build
 
 fmt:
-	ocamlformat --inplace --enable-outside-detected-project $(OCAML_SRCFILES)
+	esy x ocamlformat --inplace --enable-outside-detected-project $(OCAML_SRCFILES)
```
rgrinberg, line 9: I don't use esy to develop the extension so let's leave this target useful for everyone.

It's just easy to run make inside of esy x.

ulugbekna:
I don't know how to keep track of many files changes that esy.lock causes, especially in PRs by different people. Maybe to fix dependencies to fixed versions. 

Anyway, removed esy sandbox for running ocamlformat. 

@rgrinberg but how do you develop without esy? pulling dependencies, building, etc.

rgrinberg:
> I don't know how to keep track of many file changes that esy.lock causes, especially in PRs by different people. Maybe to fix dependencies to particular versions.

I'm not exactly sure why is this a problem? The lock file changes only when esy.json changes. This is expected.

> but how do you develop without esy? pulling dependencies, building, etc.

Well the only thing you need is bucklescript to build & run. I have ocamlformat installed in a global switch.

PR #375, opened 2020-09-28
Switch to esbuild bundler

mnxn:
Switches to the esbuild bundler in preparation for the switch to Js_of_ocaml. 

JSOO produces much larger JS files, which drastically increase the time taken by `ncc` to bundle the files (took 15 seconds on my computer). Switching to Esbuild speeds up the bundling time significantly, and it can also minify the generated file.


rgrinberg (approved):

package.json
```diff
@@ -559,9 +559,11 @@
     "@glennsl/bs-json": "5.0.2",
     "@vercel/ncc": "0.24.1",
     "bs-platform": "8.2.0",
+    "esbuild": "^0.7.7",
```
smorimoto, line 562:
```suggestion
    "esbuild": "0.7.7",
```

package.json
```diff
@@ -559,9 +559,11 @@
...
     "fs-extra": "9.0.1",
     "jest-cli": "26.4.2",
     "mocha": "7.2.0",
+    "nodemon": "^2.0.4",
```
smorimoto, line 566:
```suggestion
    "nodemon": "2.0.4",
```

smorimoto (commented):
This project uses the exact version.

mnxn:
We actually don't need a bundler since the only bundled library is `vscode-languageclient`. What would be useful is a decently fast minifier with dead code elimination. Closing this for now.

smorimoto:
Why not just use esbuild's minify feature?

mnxn:
I wasn't aware that esbuild's minify did tree-shaking/dead-code elimination, but it does look like a good option if we revisit this.

Removing the bundler alleviates the problem I had with long bundle times. I am going to hold off on adding a minifier until we have a complete JSOO version of the extension with decent source maps.

smorimoto:
Sounds reasonable.

PR #383, opened 2020-10-02
add custom icons for switchImplIntf

ulugbekna:
fixes #379

PR #384, opened 2020-10-03
add keyboard shortcut for destruct code action

ulugbekna:
On `alt+d`, user gets what they would otherwise get by `cmd+.` to open quick fix options and picking the only option that we have "destruct". 

image

User still can open quick fix options and pick "destruct" from there or just click on `alt+d`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,13 @@
         "command": "ocaml.switch-impl-intf",
         "key": "Alt+O",
         "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
+      },
+      {
+        "command": "editor.action.codeAction",
+        "key": "Alt+D",
+        "args": {
+          "kind": "destruct"
+        }
       }
     ],
     "menus": {
--- a/package.json
+++ b/package.json
@@ -73,7 +73,8 @@
         "key": "Alt+D",
         "args": {
           "kind": "destruct"
-        }
+        },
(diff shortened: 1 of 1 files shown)


mnxn:
Looks good. Needs change log entry.

ulugbekna:
I'll rebase and update when we merge #383

package.json
```diff
@@ -67,6 +67,13 @@
         "command": "ocaml.switch-impl-intf",
         "key": "Alt+O",
         "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
+      },
+      {
+        "command": "editor.action.codeAction",
+        "key": "Alt+D",
```
rgrinberg, line 73: Should this binding be available in mli files?

mnxn: Doesn't the `destruct` code action only expand to `match ... with` expressions? That wouldn't be valid in a mli. The language server doesn't even provide a `destruct` code action in an mli.

rgrinberg: Indeed. As far as I can see we aren't checking for `editorLangId == ocaml || editorLangId == reason` in this binding. Does that mean it will be available in mli, ocamllex, and menhir files?

mnxn: Oh, I see your point. That is indeed a problem. Using `Alt+D` in any file - even non-OCaml files  will error with `No code actions for destruct available`.

mnxn: I pushed a commit that fixes this with the condition you mentioned.

rgrinberg (approved):

smorimoto:
Oops, I forgot the changelog for this. I will open another PR.

PR #385, opened 2020-10-03
 Port to Js_of_ocaml

mnxn:
I have completed the port from BuckleScript to Js_of_ocaml. From my testing, all of the features continue to be functional under both opam and esy sandboxes. 

All bucklescript-specific code and configuration files have been removed. The VSCode npm dependency has been bumped to 1.49 because the VSCode bindings are written around that version.

Although the size of the generated JS has increased to about 16MB with Js_of_ocaml, the installation vsix package has only increased to 3.5MB from BuckleScript's 1.0MB.

A few opam dependencies have been introduced:

- [js_of_ocaml](https://opam.ocaml.org/packages/js_of_ocaml/) and [gen_js_api](https://opam.ocaml.org/packages/gen_js_api/) for creating JS bindings.

- [core_kernel](https://opam.ocaml.org/packages/core_kernel/) for a few instances where the OCaml stdlib does not provide certain functionality. We could also opt to use Core_kernel as a complete stdlib replacement.

- [promise_jsoo](https://opam.ocaml.org/packages/promise_jsoo/) for gen_js_api-compatible bindings to JS Promises. This library includes variations of the functions that were previously in `Promise.ml`.

- [jsonoo](https://opam.ocaml.org/packages/jsonoo/) for gen_js_api-compatible JSON type and convertors. This library replaces the functionality of bs-json.


mnxn:
It took me a while, but I finally figured out the CI. 

We will have to continue to use OCaml 4.06 in the CI for running the bsb and opam tests, while OCaml 4.10 will be used for building the extension and running esy tests. 

- OCaml 4.06 can't be used to build the extension itself because some of the dependencies need at least 4.08. 
- OCaml 4.11 can't be used to build the extension because the latest published ReasonML package doesn't support it yet.
- OCaml 4.10 can't be used in the CI because BuckleScript requires 4.06.

The JS file is built by `esy build` in the CI and `yarn package` no longer calls the build command. It's a bit confusing having this mix of build systems in a single project, so there's definitely room for improvement. I think it would be useful to move the remaining commands from the `package.json` to the Makefile to make it simpler.

smorimoto:
Amazing work! I just skimmed it on the move but it looks good overall. What does everyone else think?

smorimoto (approved):

esy.json
```diff
@@ -1,8 +1,20 @@
 {
+  "esy": {
+    "build": "make build",
```
rgrinberg, line 3: Can we call dune directly?

esy.json
```diff
@@ -1,8 +1,20 @@
...
+    "buildsInSource": "_build"
```
rgrinberg, line 4: Why do we need this line?

src/DuneTaskProvider.ml
```diff
@@ -4,38 +4,42 @@ type t = Disposable.t option ref
...
 let source = task_type
 
-let problemMatchers = [| "$ocamlc" |]
+let problem_matchers = [ "$ocamlc" ]
 
 (* the ocamlc matcher is not able to parse ocaml compiler errors unless they
    follow the short style. *)
-let env = Js.Dict.fromList [ ("OCAML_ERROR_STYLE", "short") ]
+let env =
+  Core_kernel.Hashtbl.of_alist_exn
```
rgrinberg, line 16: Can this be a map? I don't think it's right to modify this.

dune-project
```diff
@@ -0,0 +1 @@
+(lang dune 2.7)
```
rgrinberg, line 1: Let's add a `(name ..)` field here.

esy.json
mnxn: That line is so the esy build location is consistent with normal dune. The `main` is specified to be `_build/default/src/extension.bc.js` in the package.json. If that line wasn't there, it would be built under the `_esy` folder.

src/Cmd.ml
```diff
@@ -29,12 +29,12 @@ let candidateFns =
 
 let which path fn =
   let candidates = candidateFns fn in
-  Js.String.split envSep path
-  |> Promise.Array.findMap (fun p ->
+  Core_kernel.String.split ~on:envSep path
```
rgrinberg, line 32: You should `open Core_kernel` in import. will make this a bit simpler to read.

src/Cmd.ml
```diff
@@ -55,39 +55,48 @@ let check t =
...
+      :: ("args", list string args)
       :: message
-    | Shell commandLine -> ("shell", Log.field commandLine) :: message
+    | Shell commandLine -> ("shell", string commandLine) :: message
   in
   logJson "external command" message
 
 let output ?stdin (t : t) =
-  let open Promise.O in
+  let open Promise.Syntax in
```
rgrinberg, line 93: Why was this renamed?

esy.json
mnxn: I suppose we could, but that would be duplicating `dune build src/extension.bc.js` in two places.

Is there was a way for `dune build` to build that JS file automatically?

src/Output.ml
```diff
@@ -1,5 +1,5 @@
 let languageServerOutputChannel =
-  lazy (Vscode.Window.createOutputChannel "OCaml Language Server")
+  lazy (Vscode.Window.create_output_channel ~name:"OCaml Language Server")
```
rgrinberg, line 2: Why was this renamed? When creating bindings we should match original names whenever possible. Makes our bindings a lot easier to use.

src/Toolchain.ml
```diff
@@ -240,25 +241,25 @@ let selectPackageManager (choices : Candidate.t list) =
         (quickPick, pm))
       choices
   in
-  let options =
-    Window.QuickPickOptions.make ~canPickMany:false ~placeHolder ()
-  in
-  Window.showQuickPickItems choices options
+  let options = QuickPickOptions.create ~can_pick_many:false ~place_holder () in
+  Window.show_quick_pick_items ~choices ~options ()
```
rgrinberg, line 245: Ditto. The old name made it obvious how to look up the docs.

src/bindings/dune
```diff
@@ -0,0 +1,27 @@
+(library
+ (name interop)
+ (modules interop)
+ (preprocess
+  (pps gen_js_api.ppx))
+ (libraries core_kernel js_of_ocaml gen_js_api)
+ (modes byte))
```
rgrinberg, line 7: Could we put these libraries in separate subdirs?

src/bindings/dune
```diff
@@ -0,0 +1,27 @@
...
+ (preprocess
+  (pps gen_js_api.ppx))
+ (js_of_ocaml
+  (javascript_files node_stub.js))
+ (libraries interop core_kernel promise_jsoo jsonoo gen_js_api)
+ (modes byte))
+
+(library
+ (name vscode)
+ (modules vscode vscode__core vscode__languageclient)
```
rgrinberg, line 21: Why are you manually using the `__` convention here? this is just a convention chosen by dune, we shouldn't use it.

src/dune
```diff
@@ -0,0 +1,5 @@
+(executable
+ (name extension)
```
rgrinberg, line 2: I'd rename this executable to `vscode-ocaml-platform`. The name `extension` is a bit too generic.

src/path.ml
```diff
@@ -2,19 +2,19 @@ type t = string
 
 let ofString s = s
 
-let isAbsolute t = Node.Path.isAbsolute t
+let isAbsolute t = not (Filename.is_relative t)
```
rgrinberg, line 5: We can now use the standard naming convention here.

esy.json
rgrinberg: We could introduce an alias. E.g. `dune build @vscode` or even just `$ dune build @all`.

src/Cmd.ml
rgrinberg: The `O` convention is what we use in dune to mean the same thing. I like it a little more than `Syntax` which is quite verbose. In any case, I don't think it matters here. I don't mind following core's convention.

src/opam.ml
```diff
@@ -21,24 +21,25 @@ let binary = Path.ofString "opam"
...
   | Ok cmd -> Some cmd
 
 let parseSwitchList out =
   let lines = String.split_on_char '\n' out in
   let result =
-    Belt.List.keepMap lines (function
-      | "" -> None
-      | s -> Some (Switch.make s))
+    lines
+    |> List.filter_map (function
```
rgrinberg, line 33: This module should also use Core_kernel

mnxn: So is it all right to completely transition to using Core_kernel?
I didn't do that when I made the PR because I didn't want to make too many unnecessary changes.

rgrinberg:
CI needs to be updated to 4.10, I think?

src/opam.ml
rgrinberg: Okay, it can  be done later. I saw other files ported to Core_kernel, so I thought you were doing it here too.

mnxn:
> CI needs to be updated to 4.10, I think?

I mentioned in [this comment](https://github.com/ocamllabs/vscode-ocaml-platform/pull/385#issuecomment-703183258) that we can't do that as long as we're still testing for BuckleScript, which requires the 4.06 version of the tools.

rgrinberg:
> I mentioned in this comment that we can't do that as long as we're still testing for BuckleScript, which requires the 4.06 version of the tools.

Oops sorry. Missed that.

I think we can just drop the bucklescript tests. We don't officially support bucklescript anymore. We'll still fix bucklescript bugs in the short term (~6 month), but by then, I hope that most people would transition to the official rescript extension.

rgrinberg:
Perhaps we should split the vscode core and language client bindings into separate libraries `Vscode` and `Vscode_languageclient`? It seems simpler to me to just mirror the upstream structure as much as possible.

src/bindings/node/dune
```diff
@@ -0,0 +1,9 @@
+(library
+ (name node)
+ (modules node)
```
rgrinberg, line 3: Now `modules` can be omitted.

src/bindings/node/node.ml
```diff
@@ -0,0 +1,139 @@
+open Interop
+
+module Process = struct
+  val cwd : unit -> string [@@js.global "process.cwd"]
+
+  val platform : string [@@js.global "process.platform"]
+
+  val env : unit -> string Dict.t [@@js.get "process.env"]
```
rgrinberg, line 8: Is it possible to keep this type completely opaque?

```ocaml
module Env : sig
  type t
  val get : t -> string -> string option
  val set : t -> string -> string -> unit
end
```
and that's it.

ulugbekna:
I would agree with splitting vscode and languageclient; less so about prefixing it with vscode

rgrinberg:
> less so about prefixing it with vscode

We could always unprefix in our import.ml file.

@mnxn I'm curious, why didn't you decide to upstream `Promise_jsoo` to jsoo itself? The promise API exists on every browser, correct? Usually such bindings just live directly in Jsoo.

mnxn:
> @mnxn I'm curious, why didn't you decide to upstream `Promise_jsoo` to jsoo itself? The promise API exists on every browser, correct? Usually such bindings just live directly in Jsoo.

I don't think I'm prepared to contribute such a project to Js_of_ocaml yet. Especially since some of the language/library features used would raise the Js_of_ocaml's minimum OCaml version up from 4.02 (let operators, `List.find_map`, `List.filter_map`). I could try slimming it down and upstreaming in the future.

rgrinberg:
Do consider upstreaming what's possible in the future. Jsoo's own browser bindings are fairly complete because there's a large group of contributors sending PR's.

src/Extension.ml
```diff
@@ -11,37 +11,39 @@ let openTerminalSelectCommandId = "ocaml.open-terminal-select"
...
-      { command = Path.toString bin
-      ; args = Array.of_list args
-      ; options = { env; shell = false }
-      }
+      let command = Path.toString bin in
+      let options =
+        LanguageClient.ExecutableOptions.create ~env ~shell:false ()
+      in
+      LanguageClient.Executable.create ~command ~args ~options ()
 end
```
ulugbekna, line 47: What do you think if we replace these single function modules with just functions? `createServerOptions ()` or `serverOptions ()`

src/dune
mnxn: If I change the name to `vscode-ocaml-platform`, dune looks for the `Vscode-ocaml-platform` module as an entry point. That doesn't seem like a very optimal change.

Is there a way to have separate names for the executable and entry-point module?

rgrinberg:
> Is there a way to have separate names for the executable and entry-point module?

Unfortunately not. You can replace the dashes with underscores however. That seems good enough.

Khady:
> Although the size of the generated JS has increased to about 16MB with Js_of_ocaml, the installation vsix package has only increased to 3.5MB from BuckleScript's 1.0MB.

Have you tried to build in release profile to see if the JS generated by jsoo is smaller? From what I see in the makefile you always build with the default profile.

mnxn:
> Have you tried to build in release profile to see if the JS generated by jsoo is smaller? From what I see in the makefile you always build with the default profile.

I can't believe I forgot about this. Much smaller now - only 2.1MB!

Unfortunately It looks like the sourcemap no longer works, even with the `-source-map` flag. I'm not sure if there's another combination of flags which would keep the sourcemap at a small size.

rgrinberg:
Let's compare in both cases to decide? IMO, if we get better stacktraces with larger binaries, then so be it.

rgrinberg:
@mnxn I pushed a few tweaks. This is ready to go.

One small request: can we generate some opam files from the dune-project? We aren't planning to publish this extension on opam, but it would make it easier to install the dependencies using opam.

rgrinberg (approved):

mnxn:
> One small request: can we generate some opam files from the dune-project?

Sure, should I specify version constraints for any of the libraries?

What values should I use for the `maintainers` and `authors`?

rgrinberg:
> Sure, should I specify version constraints for any of the libraries?

Let's just set the lower bounds to be the versions of dependencies that you've tested.

> What values should I use for the maintainers and authors?

You can list me as the maintainer. For authors, we can put anyone with over 10 commits as listed by `git shortlog -s -n -e`.

rgrinberg:
Btw, you can create a separate package for your bindings as well. Those can actually be published to opam (if you choose to).

mnxn:
> Btw, you can create a separate package for your bindings as well. Those can actually be published to opam (if you choose to).

If you're referring to the VSCode and Node bindings, I don't feel that those bindings are complete enough to justify publishing to opam.

ulugbekna:
Congrats everyone and huge huge thanks to Max and Rudi!

yminsky:
A minor point: if you want to reduce the size of the generated code, you could consider using Base instead of Core_kernel.

idkjs:
Do the docs need to change to reflect this? They still refer to [Bucklescript here](https://github.com/ocamllabs/vscode-ocaml-platform#bucklescript).

mnxn:
> Do the docs need to change to reflect this? They still refer to [Bucklescript here](https://github.com/ocamllabs/vscode-ocaml-platform#bucklescript).

No, those instructions are about how to _use_ the extension for BuckleScript projects. This PR only changed how the extension is built. BuckleScript will be supported with those instructions for the time being.

PR #391, opened 2020-10-07
Do not open Core_kernel everywhere

rgrinberg:
It should be available via the Import module

### DIFF
--- a/src/Cmd.ml
+++ b/src/Cmd.ml
@@ -25,14 +25,14 @@ let candidate_fns =
   [| x |]
   else
     fun x ->
-  [| ".exe"; ".cmd" |] |> Array.map (fun ext -> Path.with_ext x ~ext)
+  [| ".exe"; ".cmd" |] |> Array.map ~f:(fun ext -> Path.with_ext x ~ext)
 
 let which path fn =
   let candidates = candidate_fns fn in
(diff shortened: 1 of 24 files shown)


rgrinberg:
I did not touch the bindings because I'm wondering if we should just get rid of core in the bindings completely? I don't think it's used much there and it would make the bindings a lot more useful for people

mnxn:
> I did not touch the bindings because I'm wondering if we should just get rid of core in the bindings completely? I don't think it's used much there and it would make the bindings a lot more useful for people

I can rewrite the JsDict module so it doesn't need to reference the Core_kernel Map type.

mnxn:
I renamed `JsDict` to `Dict` since JsDict implies that it has the same representation as the JS value, which it doesn't.

I also included the StringMap module inside of Dict because I thought it would be useful to have access to those functions outside of the module.
https://github.com/ocamllabs/vscode-ocaml-platform/blob/38bb9516700231b3a98c01975da994a96d053a3b/src/bindings/interop/interop.ml#L57
https://github.com/ocamllabs/vscode-ocaml-platform/blob/38bb9516700231b3a98c01975da994a96d053a3b/src/bindings/interop/interop.mli#L24

mnxn (approved):

PR #392, opened 2020-10-07
Add vscode pacakge & move to own dir

rgrinberg:
One directory per package is the recommended way to organize things with
dune.


rgrinberg:
Note that this doesn't mean we need to publish this package. It just makes it easier to pin the bindings.

mnxn:
Not sure if this changes anything, but shouldn't `src/dune` reference `vscode.languageclient` instead of `vscode_languageclient` now?

rgrinberg:
> Not sure if this changes anything, but shouldn't src/dune reference vscode.languageclient instead of vscode_languageclient now?

There used to be a bug in dune that made this matter. I've fixed it in 2.7.0 or 2.7.1 so that it no longer makes a difference.

mnxn (approved):

rgrinberg:
It would also be nice if we could move the extension package into its own subdirectory. But that seems like a much larger change. I'll refrain for now.

PR #397, opened 2020-10-07
Move extension to dist directory

rgrinberg:
This is easier to remember than _build/default

### DIFF
--- a/dune
+++ b/dune
@@ -1,1 +1,11 @@
 (dirs :standard \ node_modules)
+
+(subdir
+ dist
+ (rule
+  (alias vscode)
+  (mode
+   (promote (until-clean)))
+  (action
+   (copy %{project_root}/src/vscode_ocaml_platform.bc.js
+     ./vscode_ocaml_platform.js))))
--- a/dune
+++ b/dune
@@ -1,13 +1,1 @@
 (dirs :standard \ node_modules)
-
-(subdir
- dist
- (rule
-  (alias vscode)
-  (mode
-   (promote (until-clean)))
-  (action
-   (progn
-    (system "mkdir -p dist")
-    (copy %{project_root}/src/vscode_ocaml_platform.bc.js
-      ./vscode_ocaml_platform.js)))))
--- a/dune
+++ b/dune
@@ -7,5 +7,7 @@
   (mode
    (promote (until-clean)))
   (action
-   (copy %{project_root}/src/vscode_ocaml_platform.bc.js
-     ./vscode_ocaml_platform.js))))
+   (progn
+    (system "mkdir -p dist")
+    (copy %{project_root}/src/vscode_ocaml_platform.bc.js
+      ./vscode_ocaml_platform.js)))))
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -71,12 +71,16 @@ jobs:
         run: |
           opam pin ocaml-lsp-server https://github.com/ocaml/ocaml-lsp.git --no-action
           opam install ocaml-lsp-server ocamlformat --yes
+          opam install --deps-only .
 
       - name: Build esy packages
         # Cleanup build cache in case dependencies have changed
         run: |
           esy build
           esy cleanup .
+
+      - name: Build extension
+        run: opam exec -- dune build @vscode
 
       - name: Install npm packages
         run: yarn --frozen-lockfile
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -49,15 +49,15 @@ jobs:
         id: esy-cache-dir-path
         run: node .github/workflows/print_esy_cache.js
 
-      - name: Restore opam cache
-        id: cache-opam
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.opam-root-dir-path.outputs.dir }}
-          # opam cache is coupled to esy lock files, so both caches will be wiped whenever esy deps change
-          key: opam-${{ matrix.os }}-${{ matrix.ocaml-version }}-${{ hashFiles('esy.lock/index.json') }}
-          restore-keys: |
-            opam-${{ matrix.os }}-${{ matrix.ocaml-version }}-
+      # - name: Restore opam cache
+      #   id: cache-opam
+      #   uses: actions/cache@v2
+      #   with:
+      #     path: ${{ steps.opam-root-dir-path.outputs.dir }}
+      #     # opam cache is coupled to esy lock files, so both caches will be wiped whenever esy deps change
+      #     key: opam-${{ matrix.os }}-${{ matrix.ocaml-version }}-${{ hashFiles('esy.lock/index.json') }}
+      #     restore-keys: |
+      #       opam-${{ matrix.os }}-${{ matrix.ocaml-version }}-
 
       - name: Restore esy cache
         uses: actions/cache@v2
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ _build/
 _esy/
 _opam/
 .vscode-test/
-dist/
+dist/*.js
 lib/
 node_modules/
 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,7 +11,7 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "outFiles": [
-        "${workspaceRoot}/_build/default/src/vscode_ocaml_platform.bc.js"
+        "${workspaceRoot}/dist/vscode_ocaml_platform.js"
       ],
       "args": [
         "--disable-extensions",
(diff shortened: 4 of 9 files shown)


vscode/interop/interop.mli
```diff
@@ -20,12 +20,10 @@ module Regexp : sig
...
+  type 'a t = (string, 'a) Core_kernel.Map.Poly.t
 
-  val t_to_js : ('a -> Ojs.t) -> 'a t -> Ojs.t
+  val t_to_js : ('a -> Ojs.t) -> (string, 'a) Core_kernel.Map.Poly.t -> Ojs.t
 
-  val t_of_js : (Ojs.t -> 'a) -> Ojs.t -> 'a t
-
-  val of_alist : (string * 'a) list -> 'a t
+  val t_of_js : (Ojs.t -> 'a) -> Ojs.t -> (string, 'a) Core_kernel.Map.Poly.t
 end
```
mnxn, line 29: I think you have some leftover changes here.

rgrinberg: Got rid of them.

dune
```diff
@@ -1 +1,11 @@
...
+
+(subdir
+ dist
+ (rule
+  (alias vscode)
+  (mode
+   (promote (until-clean)))
+  (action
+   (copy %{project_root}/src/vscode_ocaml_platform.bc.js
+     ./vscode_ocaml_platform.js))))
```
mnxn, line 11: This rule fails if the `dist` directory doesn't exist:
```
max /Users/max/Projects/vscode-ocaml-platform [0] $ dune build @vscode                                                                                                 promote-vscode-ext
Error: dist/vscode_ocaml_platform.js: No such file or directory
Done: 0/0 (jobs: 0)%                
```

rgrinberg: That's a dune bug actually :/

mnxn: I'm pretty sure it's going to cause the CI to fail, so it needs a workaround.

rgrinberg: Okay, I just created the dune file in dist. I would have preferred to ignore all of `dist` but we need to fix the dune bug first.

rgrinberg:
@mnxn I accidentally overwrote your commit. Could you push it again?

mnxn:
I think you pushed the wrong commit or something, because there's no more `vscode` rule at all.

rgrinberg:
Yes, I forgot to stage the dune file in the dist directory

mnxn:
The CI is failing from an Esy permissions error:
```
Error: dist/vscode_ocaml_platform.js: Operation not permitted
```

This seems to be an existing problem: https://github.com/esy/esy/issues/1085.

rgrinberg:
@ManasJayanth is there any workaround for this?

ManasJayanth:
Setting `esy.buildsInSource`, in the esy.json, to true is the workaround you are looking for.

rgrinberg:
> Setting esy.buildsInSource, in the esy.json, to true is the workaround you are looking for.

I think I tried that and it didn't seem to work. Let me try again to post the error message.

ManasJayanth:
Caveat: `buildsInSource = true`: esy has to run an additionally step of copying sources to a build directory.

rgrinberg:
There you go: `Error: mkdir: _build: Operation not permitted`.

@mnxn shall we just experiment with opam's lock files? I don't mind keep an esy file around for those who insist on using it, but it doesn't seem to fully support dune builds. This adds needless friction for us.

mnxn:
> There you go: `Error: mkdir: _build: Operation not permitted`.
> 
> @mnxn shall we just experiment with opam's lock files? I don't mind keep an esy file around for those who insist on using it, but it doesn't seem to fully support dune builds. This adds needless friction for us.

Using dune through esy directly (`esy dune build @vscode`) seems to work fine on my computer. I just pushed that version.

I'm okay with switching to opam for building the extension in CI, but we also use the `esy.json` file to install esy dependencies into cache which are used for tests. The tests only need a few packages so I'd like to avoid the situation where we're installing dependencies that go unused.

If this CI check succeeds, I suggest that we move the CI to opam in a separate PR.

ManasJayanth:
> but it doesn't seem to fully support dune builds

I notice build artifacts being moved around the source tree. This is definitely going to be unpleasant with esy. 

What do we gain from moving the `.js` to `dist/`?

rgrinberg:
Modifying the source directory is perfectly reasonable with dune via its promotion mechanism. What do we gain from restricting it? I don't use esy at all, so I definitely gain nothing.In exchange, I only lose from having to remember the specifics of maintaining esy support.

ManasJayanth:
> What do we gain from restricting it? 

This issue probably not the best place to discuss this. Happy to discuss this elsewhere. The way the repo is setup, `esy dune build @vscode` seems like a nice workaround. For anyone else reading this, it is not usually recommended doing this. Regarding whether, esy fits this repo's use case, I don't have a strong opinion.

mnxn:
The esy build workaround works on macos and linux but is failing on windows:
https://github.com/ocamllabs/vscode-ocaml-platform/runs/1223463930

I've decided to just switch to opam. The CI caching behavior still needs to be updated to accommodate that.

ManasJayanth:
As long as esy is supported by the plugin and ocamllsp behind the scenes, it's fine. If you wish to look for a workaround for the Windows issue, make the command `esy b dune build @vscode` only for Windows (Windows doesn't have a bwrap/sandbox-exec equivalent support right now)

rgrinberg:
If `esy b dune build` the correct way to invoke dune, then I'm fine testing with that. I'd rather just avoid adding hacks just so things work with esy. Is the behavior of `esy b` vs `esy` wrt to source dir permissions documented? I'd like to understand that a little better.

mnxn:
@rgrinberg what should we use to get lock files for opam?

ManasJayanth:
@rgrinberg Any build operation is recommended to be run in an isolated sandboxed environment where the sources are considered immutable. Esy uses sandbox-exec on mac to enforce a sandbox (Linux and Windows are pending). It is always recommended that build commands are run with `esy b ..` prefix. For this to work, esy assumes that there is not inplace editing of source tree - any promotions (as they are called) that take place in the isolated phase make it hard for esy to bet on anything and hence it is not written to handle it well at the moment (hence the cryptic error messages)

Any command that does not generate build artifacts (`dune subst`, launching lightweight editors etc) are recommended to be run with just `esy ...` prefix (We call it the command environment to distinguish it from build environment)

In this case, `esy build dune build` could run in the build environment and `esy dune build @vscode` (or any other way to run just the copying alias) in the command environment where in source editing is allowed (and the environment is not isolated)

This is as far as using esy for moving around generated assets is concerned. IMO, this should be outside the build tool and into the dev scripts (I can see a Makefile in the repo and think its a good place to place the cp/mv command there). I feel this PR would have been much simpler. By making such compromises (placing the mv/cp command outside Dune), tools that can leverage immutability of sources can provide interesting benefits.

I can understand if such benefits are not the goal of this repo (development-wise). I'm already very happy that the plugin supports esy projects first class and have been recommending this plugin to everyone.

rgrinberg:
> Any command that does not generate build artifacts (dune subst, launching lightweight editors etc) are recommended to be run with just esy ... prefix (We call it the command environment to distinguish it from build environment)

`dune subst` edits the source files in place however. Are you sure that works?

> This is as far as using esy for moving around generated assets is concerned. IMO, this should be outside the build tool and into the dev scripts (I can see a Makefile in the repo and think its a good place to place the cp/mv command there).

I don't get the motivation. Why would a developer prefer to cobble together dev scripts using cp/mv instead of relying on a streamlined process that can be relied on in every single repository? Your cp/mv solution is not going to work well with with watch mode for example. 

> I can understand if such benefits are not the goal of this repo (development-wise).

There's plenty of other dune projects out there that are using this feature. So these restrictions are really hurting the usability of the tool.

ManasJayanth:
It's hard to answer why developers would prefer (or not) a separate copy step outside. If they opt-in, they'll benefit from immutability features tools like esy offer - which do support inplace editing of sourcetree but do so by copying the sources which can be a slow process. Esy does this too, but doesn't handle this well enough (hence the unexplained `operation not permitted` messages).

> There's plenty of other dune projects out there that are using this feature. So these restrictions are really hurting the usability of the tool.

In the long run, Esy will handle such mutations of the source tree better and provide a slower alternative.

ManasJayanth:
@rgrinberg This kind of moving built artifacts supported better in the install step - as part of @install alias of Dune IIUC. I think it's fair to expect projects to do this

mnxn:
Alright, the CI with opam shouldn't fail now. I didn't know what to use for opam lockfiles, so I just disabled the opam package cache. It takes a little while longer for the CI to run, but we can add the opam cache back in another PR.

mnxn (approved):

rgrinberg:
Sounds good to me. Thanks Max.

PR #399, opened 2020-10-08
Fix naming convention

rgrinberg:
Prefer snake_case for our identifiers


src/ocaml_lsp.ml
```diff
@@ -40,4 +40,4 @@ let of_initialize_result (t : LanguageClient.InitializeResult.t) =
 
 let interfaceSpecificLangId t = t.interfaceSpecificLangId
 
-let canHandleSwitchImplIntf t = t.handleSwitchImplIntf
+let can_handle_switch_impl_intf t = t.handleSwitchImplIntf
```
mnxn, line 43: Both of these should be snake_case

rgrinberg: For the field names, perahps we should keep them to match the identifiers from capabilities payload?

mnxn: I mean that `let interfaceSpecificLangId` should be snake_case if `let can_handle_switch_impl_intf` is. The field names can stay the same.

rgrinberg: OK. I added a `has` prefix to match the `can` prefix of the other function.

mnxn (approved):

PR #401, opened 2020-10-09
Fix path handling on Windows

mnxn:
Under the Js_of_ocaml build of the extension, `Sys.unix` is always true, which prevents the `Filename` module and `candidates` function from handling Windows paths properly. 

This PR switches to using the Node `path` module for filename functions, and `Platform.t` for the candidates function.

### DIFF
--- a/src/bindings/node/node.mli
+++ b/src/bindings/node/node.mli
@@ -49,7 +49,7 @@ module Stream : sig
 end
 
 module Path : sig
-  val delimiter : string
+  val delimiter : char
 
   val basename : string -> string
 
--- a/src/bindings/node/node.mli
+++ b/src/bindings/node/node.mli
@@ -48,6 +48,20 @@ module Stream : sig
   val end_ : t -> unit
 end
 
+module Path : sig
+  val delimiter : string
+
+  val basename : string -> string
+
+  val dirname : string -> string
+
+  val extname : string -> string
+
+  val isAbsolute : string -> bool
+
+  val join : string list -> string
+end
+
 module Fs : sig
   val read_dir : string -> (string list, string) result Promise.t
 
--- a/src/path.ml
+++ b/src/path.ml
@@ -8,7 +8,7 @@ let of_string s = s
 
 let to_string s = s
 
-let delimiter = Node.Path.delimiter.[0]
+let delimiter = Node.Path.delimiter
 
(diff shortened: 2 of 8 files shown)


rgrinberg:
@hhugo why is `Sys.unix` always true in jsoo?

mnxn:
It's hardcoded to unix in https://github.com/ocsigen/js_of_ocaml/blob/master/runtime/sys.js.

I'm guessing that there's no good way to determine the operating system across browser and node runtimes.

src/path.ml
```diff
@@ -8,35 +8,34 @@ let of_string s = s
 
 let to_string s = s
 
-let is_absolute t = Filename.is_absolute t
+let delimiter = Node.Path.delimiter.[0]
```
rgrinberg, line 11: Do we need to assume it's a single character? Can't we just split on a string?

rgrinberg: Otherwise, let's add `assert (String.length delimiter = 1)`.

mnxn: [path.delimiter](https://nodejs.org/api/path.html#path_path_delimiter) will always return `:` or `;` as a JS string and `String.split` expects a `char`.

rgrinberg:
> I'm guessing that there's no good way to determine the operating system across browser and node runtimes.

For browsers, there's user agent strings. Anyway, it just seemed a little harsh to break anything using `Sys.unix`.

src/bindings/node/node.mli
```diff
@@ -48,6 +48,20 @@ module Stream : sig
   val end_ : t -> unit
 end
 
+module Path : sig
+  val delimiter : string
```
rgrinberg, line 52: Or better yet, we could try binding `delimiter` as a `char`.

mnxn: `char` is not a supported [gen_js_api type](https://github.com/LexiFi/gen_js_api/blob/master/TYPES.md#js-able-types). Do you mean I should move the assertion and indexing to the node module?

rgrinberg: I meant using `char` directly. But if it's not supported, then I suppose moving our cast to the bindings also makes sense.

rgrinberg:
Shall we publish 1.3.2 with this fix?

mnxn:
Yes, that's what I was planning.

PR #406, opened 2020-10-13
Update VSCode bindings for 1.50

mnxn:
TaskProvider cancellation tokens are no longer optional in VSCode 1.50:

https://github.com/microsoft/vscode/blob/93c2f0fbf16c5a4b10e4d5f89737d9c2c25488a3/src/vs/vscode.d.ts#L6301

https://github.com/microsoft/vscode/blob/93c2f0fbf16c5a4b10e4d5f89737d9c2c25488a3/src/vs/vscode.d.ts#L6316

### DIFF
--- a/vscode/vscode/vscode.mli
+++ b/vscode/vscode/vscode.mli
@@ -1718,23 +1718,15 @@ module TaskProvider : sig
   val t_to_js : t -> Ojs.t
 
   val provideTasks :
-    t -> ?token:CancellationToken.t -> unit -> Task.t list ProviderResult.t
+    t -> token:CancellationToken.t -> Task.t list ProviderResult.t
 
   val resolveTasks :
-       t
-    -> task:Task.t
-    -> ?token:CancellationToken.t
-    -> unit
-    -> Task.t ProviderResult.t
+    t -> task:Task.t -> token:CancellationToken.t -> Task.t ProviderResult.t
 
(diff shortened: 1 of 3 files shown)


vscode/vscode/vscode.mli
```diff
@@ -1718,23 +1718,15 @@ module TaskProvider : sig
...
-    -> task:Task.t
-    -> ?token:CancellationToken.t
-    -> unit
-    -> Task.t ProviderResult.t
+    t -> task:Task.t -> token:CancellationToken.t -> Task.t ProviderResult.t
 
   val create :
-       provideTasks:
-         (?token:CancellationToken.t -> unit -> Task.t list ProviderResult.t)
+       provideTasks:(token:CancellationToken.t -> Task.t list ProviderResult.t)
```
rgrinberg, line 1727: Not entirely related to this PR, but do we need the `token` label here?

rgrinberg (approved):

vscode/vscode/vscode.mli
mnxn: Not necessarily, but I'd like to keep the labels consistent across the bindings.

ulugbekna: Agreed with Rudi: since we have type `CancellationToken.t`, don't really need label but up to you

PR #408, opened 2020-10-16
Use monadic let operators

### LINKED ISSUE #394: Change >>| and >>= to let+ and let*
We can do this since we're no longer stuck with an old version of OCaml.

mnxn:
Closes #394

### DIFF
--- a/vscode/vscode_languageclient/vscode_languageclient.ml
+++ b/vscode/vscode_languageclient/vscode_languageclient.ml
@@ -154,7 +154,8 @@ module LanguageClient = struct
 
   let readyInitializeResult t =
     let open Promise.Syntax in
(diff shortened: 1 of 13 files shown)


vscode/vscode_languageclient/vscode_languageclient.ml
```diff
@@ -154,7 +154,8 @@ module LanguageClient = struct
 
   let readyInitializeResult t =
     let open Promise.Syntax in
-    onReady t >>= fun () -> Promise.return (initializeResult t)
+    let* () = onReady t in
```
rgrinberg, line 157: This could be `let+`.

rgrinberg (approved):

PR #409, opened 2020-10-18
Handle empty workspaceFolders list

mnxn:
`workspaceFolders` is empty if no folder in vscode is open.

`maybe_list` turns `undefined` results into an empty list.


smorimoto (approved):
Indeed.

PR #410, opened 2020-10-19
Update README instructions to use ocaml-lsp-server from the opam repository

mbacarella:
It took me a few seconds of thinking to realize I was supposed to install `ocaml-lsp-server` and not `lsp`.  This is a minor thing for experienced users but I could see it tripping newer users up for awhile with false starts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ npm install esy --global
 ```json
 {
   "dependencies": {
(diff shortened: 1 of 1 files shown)


mnxn (approved):
I wasn't aware the language server was already published to the opam repository.

I changed the BuckleScript instructions to use the published version too.

smorimoto (approved):

PR #413, opened 2020-10-19
Remove duplicate esy sandboxes from package manager selection

mnxn:
| Before | After |
| - | - |
| before | after |

This won't show an option for invalid toolchains though:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/a1e21d0f95304cdf3e3606c296d029eba01aeb7f/src/esy.ml#L34

Should the extension even show an option to select if the package.json is invalid?

### DIFF
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -26,10 +26,10 @@ module Discover = struct
     | n -> n
 
   let invalid_json file json_file =
-    Some
-      { file
-      ; status = Error ("unable to parse " ^ json_file ^ " file for Esy")
-      }
+    let message =
+      Printf.sprintf "Esy manifest file '%s' is not a valid json file" json_file
+    in
+    Some { file; status = Error message }
 
   let is_esy_compatible filename json =
     String.equal filename "esy.json"
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -18,6 +18,8 @@ let make () =
 module Discover = struct
   let valid file = Some { file; status = Ok () }
 
+  let compare l r = Path.compare l.file r.file
+
   let invalid_json file = Some { file; status = Error "unable to parse json" }
 
   let parse_file project_root = function
@@ -56,6 +58,7 @@ module Discover = struct
               dir;
             [])
     >>= Promise.List.filter_map (parse_file dir)
+    >>| List.dedup_and_sort ~compare
 
   let parse_dirs_up dir =
     let rec loop parsed_dirs dir =
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -25,30 +25,27 @@ module Discover = struct
     | 0 -> compare_status l.status r.status
     | n -> n
 
-  let invalid_package_json file =
-    Some { file; status = Error "unable to parse package.json" }
+  let invalid_json file json_file =
+    Some { file; status = Error ("unable to parse " ^ json_file) }
+
+  let is_esy_compatible filename json =
+    String.equal filename "esy.json"
+    || ( property_exists json "dependencies"
+       || property_exists json "devDependencies" )
+       && property_exists json "esy"
 
   let parse_file project_root = function
-    | "esy.json"
-    | "opam" ->
-      Promise.return (valid project_root)
+    | "opam" -> Promise.return (valid project_root)
     | s when String.equal (Caml.Filename.extension s) ".opam" ->
       Promise.return (valid project_root)
-    | "package.json" as fname -> (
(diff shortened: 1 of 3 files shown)


rgrinberg:
> Should the extension even show an option to select if the package.json is invalid?

I think that's a better user experience. When a package.json file is silently omitted, the user might be confused as to why his project isn't being detected.

src/esy.ml
```diff
@@ -69,6 +71,7 @@ module Discover = struct
   let run ~dir : discover list Promise.t =
     let open Promise.Syntax in
     dir |> parse_dirs_up |> Promise.all_list >>| List.concat
+    >>| List.dedup_and_sort ~compare
```
rgrinberg, line 74: Shouldn't the de-duplication be on `parse_dir`? This is the function that is returning duplicates as far as I see.

mnxn: Yeah that makes more sense...

rgrinberg (approved):

rgrinberg:
Another point release with this fix would be nice.

mnxn:
I'm going to simplify the discovery so it doesn't show invalid package.json toolchains, and will release after that.

rgrinberg:
> I'm going to simplify the discovery so it doesn't show invalid package.json toolchains, and will release after that.

I think my comment above was confusing. I think it does make sense to show invalid .json files. Since that clearly shows the user why a toolchain might not be picked up.

rgrinberg:
In the future, I actually wanted to extend this to opam toolchains and improve the error detection to check if lsp/ocamlformat are installed as well.

mnxn:
> > I'm going to simplify the discovery so it doesn't show invalid package.json toolchains, and will release after that.
> 
> I think my comment above was confusing. I think it does make sense to show invalid .json files. Since that clearly shows the user why a toolchain might not be picked up.

My bad. I misunderstood.

mnxn:
Added a little bit more logic to show the message for invalid package.json:
image



Before this PR, it wasn't showing those messages at all.

mnxn:
Should the toolchain selection show both the valid and invalid options? Even if you try to select the valid option, it will fail on initialization since Esy checks the package.json anyways:

image

ulugbekna:
"invalid: unable to parse package.json" looks a little like the error is
internal to the extension rather than there is an error in a user-provided
config file.

Maybe something like:

`jsonoo Esy [cannot parse package.json - make sure your package.json is a
valid json file]` or something more concise but explicit?

rgrinberg:
> Should the toolchain selection show both the valid and invalid options? Even if you try to select the valid option, it will fail on initialization since Esy checks the package.json anyways:

Do you mean "if you try to select the invalid option"?

Yes, we should valid and invalid options with explanations as to why invalid options don't work. 

If there is a way to grey out invalid sandboxes, that would be even better. If there's no way to grey them out, instead of trying to initialize them, we should just jump the json file when the user selects the sandbox.

mnxn:
> Do you mean "if you try to select the invalid option"?

No, I mean the valid option. Even if there is a valid `opam` file in the directory, Esy will try to use the `package.json` since it exists, and will fail. I found out that this is also true for `esy.json`.

That makes the "valid" option misleading, since the presence of invalid `package.json` or `esy.json` files stops the valid configuration from working.

rgrinberg:
> No, I mean the valid option. Even if there is a valid opam file in the directory, Esy will try to use the package.json since it exists, and will fail. I found out that this is also true for esy.json.

That seems like a bug in esy? Don't think there's much we can do at this point.

mnxn:
Maybe this was the wrong approach.

Currently the way Esy handling works is that it passes the directory of the project to `esy -P`. For example, if a folder has `package.json`, `esy.json` and `lib.opam`, all 3 options will do `esy -P DIR`.

From my testing, esy accepts paths to those files directly. If we change the implementation we can have options that do
`esy -P DIR/package.json`
`esy -P DIR/esy.json`
`esy -P DIR/lib.opam`

That would stop incorrect `esy.json` or `package.json` files from preventing the opam file from working. They also wouldn't be considered duplicates because they point to different files. 

I can implement that instead. What do you think? @rgrinberg @ulugbekna

rgrinberg:
Yeah, that seems like the correct approach.

mnxn:
Closed in favor of #415

mnxn:
Merging this for now because it is a simpler change than #415. Once we have better integration with esy for sandbox discovery, #415 can be revisited.

PR #415, opened 2020-10-20
Allow selecting specific configuration files for Esy sandboxes

mnxn:
The settings.json serialization/deserialization stay backwards compatible to not break users' current settings. 

With this configuration, the extension will run `esy -P /some/dir/lib.opam`
```json
"ocaml.sandbox": {
	"kind": "esy",
	"root": "/some/dir",
	"file": "lib.opam"
}
```

With this (old) configuration, the extension will continue to run `esy -P /some/dir`
```json
"ocaml.sandbox": {
	"kind": "esy",
	"root": "/some/dir"
}
```

@ulugbekna: how's the formatting for the item labels and descriptions?

| Before |
| - |
| before |

| After w/ Errors | After w/o Errors |
| - | - |
| after | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fix sandbox functionality when a folder is not opened (#409)
 - Remove duplicate esy sandboxes from package manager selection (#412)
+- Allow selecting specific configuration files (esy.json, package.json, .opam)
+  for Esy sandboxes (#415)
 
 ## 1.3.2
 
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -38,7 +38,7 @@ module Discover = struct
 
   let invalid_json root json_file =
     let message =
-      Printf.sprintf "Esy manifest file '%s' is not a valid json file" json_file
+      Printf.sprintf "manifest file '%s' is not a valid json file" json_file
     in
     Some { manifest = { root; file = Some json_file }; status = Error message }
 
--- a/src/esy.ml
+++ b/src/esy.ml
@@ -4,8 +4,25 @@ type t = Cmd.spawn
 
 let binary = Path.of_string "esy"
 
+module Manifest = struct
+  type t =
+    { root : Path.t
+    ; file : string option
+    }
+
+  let path = function
+    | { root; file = None } -> root
+    | { root; file = Some file } -> Path.(root / file)
+
+  let to_string manifest = Path.to_string (path manifest)
+
+  let to_pretty_string = function
+    | { root; file = None } -> Path.basename root
+    | { root; file = Some file } -> Path.basename root ^ "/" ^ file
+end
+
 type discover =
-  { file : Path.t
+  { manifest : Manifest.t
   ; status : (unit, string) result
   }
 
@@ -16,20 +33,14 @@ let make () =
   | Ok cmd -> Some cmd
 
 module Discover = struct
-  let valid file = Some { file; status = Ok () }
+  let valid root file =
+    Some { manifest = { root; file = Some file }; status = Ok () }
 
-  let compare l r =
-    let compare_file = Path.compare in
-    let compare_status = Result.compare Unit.compare String.compare in
-    match compare_file l.file r.file with
-    | 0 -> compare_status l.status r.status
-    | n -> n
-
-  let invalid_json file json_file =
+  let invalid_json root json_file =
     let message =
       Printf.sprintf "Esy manifest file '%s' is not a valid json file" json_file
     in
-    Some { file; status = Error message }
+    Some { manifest = { root; file = Some json_file }; status = Error message }
 
   let is_esy_compatible filename json =
     String.equal filename "esy.json"
@@ -38,16 +49,17 @@ module Discover = struct
        && property_exists json "esy"
 
   let parse_file project_root = function
-    | "opam" -> Promise.return (valid project_root)
-    | s when String.equal (Caml.Filename.extension s) ".opam" ->
-      Promise.return (valid project_root)
-    | ("esy.json" | "package.json") as fname -> (
-      let manifest_file = Path.(project_root / fname) |> Path.to_string in
+    | "opam" -> Promise.return (valid project_root "opam")
+    | opam_file when String.equal (Caml.Filename.extension opam_file) ".opam" ->
+      Promise.return (valid project_root opam_file)
+    | ("esy.json" | "package.json") as filename -> (
+      let manifest_file = Path.(project_root / filename) |> Path.to_string in
       let open Promise.Syntax in
       Fs.readFile manifest_file >>| fun manifest ->
       match Jsonoo.try_parse_opt manifest with
-      | None -> invalid_json project_root fname
-      | Some json when is_esy_compatible fname json -> valid project_root
+      | None -> invalid_json project_root filename
+      | Some json when is_esy_compatible filename json ->
+        valid project_root filename
       | _ -> None )
     | _ -> Promise.return None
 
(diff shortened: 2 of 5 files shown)


ulugbekna:
First of all, thank you for taking time to work on this!

1) Am I understanding the problem correctly that Esy has three types of "manifest" files : package.json, esy.json, and *.opam, which means that if at least one of them is found in the folder, that folder is (was?) reported as an Esy sandbox? 

2) My opinion is that we shouldn't have duplicate project titles "jsonoo/*" for Esy. Moveover, I think mixing filenames with sandbox titles as "jsonoo" for opam sandbox and "jsonoo/package.json" for Esy is a little confusing and not the best way to go. 

I think that both opam and esy sandboxes should have single selectable entry of form:

[sandbox title]  [package manager][some instructions if necessary/additional info]
[path]

For Esy case:

- when we have a certain file user indicated:

jsonoo   Esy [using package.json] 
path/to/jsonooo

- when we don't know which file to use:

jsonoo  Esy [select to choose manifest file: package.json | esy.json | ...]
path/to/jsonoo

In this case user picks the entry, which causes another input box to appear asking which file to use. I don't know whether the second box should appear if the user selects this entry when we already know which file Esy should use. Perhaps, making user to choose it every time would prevent them from opening their settings and manually editing the "file" entry, which is better. Having to pick all the time file is not the best experience, but we don't expect the user to change their sandbox all the time imo. 

If there something unclear, I can elaborate.

ulugbekna:
Message "invalid: unable to parse esy.json for Esy" is not best because:
1. "invalid" doesn't identify _what_ is invalid
2. "unable to parse" signals that _the extension_ cannot parse not that the given config file is incorrect. 

A message such as 

`[Invalid sandbox: manifest file 'package.json' is not a valid json file]` signals what is broken and what needs to be done to fix. Picking an invalid sandbox then can trigger an error popup message in right bottom corner with more details. (Such message occurs only when the user have previously chosen `package.json` as their manifest file for the esy sandbox, right?)

rdavison:
I think @ulugbekna is on the right track here, I think it sounds like a pretty good suggestion. Some thoughts I've also had that I think might be worth considering:

1) .opam/opam files don't really have anything to do with esy, so they shouldn't be labelled as relating to Esy.  Afaik, they exist in Esy projects in order to make dune happy.

2) Esy allows you have have multiple sandboxes, so it's probably not sufficient to just check `package.json` and `esy.json` (not currently sure if you check for other .json files?) (perhaps check all json files, or have some option to check a custom json file).

3) If you're able to determine through some means that a json file is _not_ an Esy configuration, then it should be filtered out automatically, and leave a message in the README about it not auto-suggesting files that are for whatever reason "invalid".

rgrinberg:
> .opam/opam files don't really have anything to do with esy, so they shouldn't be labelled as relating to Esy. Afaik, they exist in Esy projects in order to make dune happy.

I use esy to test some opam projects that don't contain any esy.json or package.json files.

> Esy allows you have have multiple sandboxes, so it's probably not sufficient to just check package.json and esy.json (not currently sure if you check for other .json files?) (perhaps check all json files, or have some option to check a custom json file).

Indeed. @EduardoRFS what do you think we should do here? Would it be possible for esy to give us a list of possible toolchains perhaps?

> If you're able to determine through some means that a json file is not an Esy configuration, then it should be filtered out automatically, and leave a message in the README about it not auto-suggesting files that are for whatever reason "invalid".

I don't think that's a good idea. I'd much rather show the user why sandbox discovery failed explicitly. However, we shouldn't allow users to select these as toolchains.

mnxn:
> Am I understanding the problem correctly that Esy has three types of "manifest" files : package.json, esy.json, and *.opam, which means that if at least one of them is found in the folder, that folder is (was?) reported as an Esy sandbox?

@ulugbekna: That's correct. It was finding the different "manifest" files before but they all just ran esy from the folder. This PR makes it run the files that it found directly.  


Esy works with opam files as documented here: https://esy.sh/docs/en/opam-workflow.html

rdavison:
> I don't think that's a good idea. I'd much rather show the user why sandbox discovery failed explicitly. However, we shouldn't allow users to select these as toolchains.

Are you perhaps thinking of a solution that I'm not imagining?  My thinking on this is as follows: Since esy config files can be named anything (not limited to `esy.json` or `package.json`) then wouldn't it be the case that you have to test every .json file when populating the suggestion list? Since most json files at the project root might not be about esy, then I think it seems unnecessary to raise parse errors (at auto-suggestion time).

However, if a user explicitly selected a .json file through a drop-down list and the file failed to parse, then I think it would make plenty of sense to show the error.

Essentially what I'm trying to avoid is bombing the user with parse errors when being presented with an auto-suggestion list.  Is there some way of logging that parse error in a way that could be discovered if some higher logging level is selected?

> I use esy to test some opam projects that don't contain any esy.json or package.json files.

and

> Esy works with opam files as documented here: https://esy.sh/docs/en/opam-workflow.html

TIL

rgrinberg:
It's a bit nuanced and I didn't appreciate this fact before this issue:

> My thinking on this is as follows: Since esy config files can be named anything (not limited to esy.json or package.json) then wouldn't it be the case that you have to test every .json file when populating the suggestion list

In this case, I would expect things to be as following:
* if the esy.json was invalid, I would certainly like to notify the user. Since this information is highly likely to be relevant
* if the package.json is invalid, I'm not confident if this should be suggested to the user.

Essentially, this is the problem I'd like to solve (or at least not make worse):

Many issues reported in this repo are of the form: "why doesn't the plugin use X?", where X is some external tool like esy, opam, lsp, etc. The answer is usually that there is some configuration error in the user's environment. Consequently, I want the plugin to as insensitive as possible in to the presence of bad configuration. In the context of the toolchain discovery process, I think that means we should display what we've discovered regardless of whether it's functional or not. We already violate this, and the plugin suffers from some unintuitive behavior like the list of sandboxes silently expanding when esy or opam become available in PATH.

> Essentially what I'm trying to avoid is bombing the user with parse errors when being presented with an auto-suggestion list. 

Yeah, this is the wrong level of detail. We should just say that the toolchain is invalid and when the user selects it, display a full error.

rdavison:
I think a reasonable middle ground would be to consider files named `esy.json` or `package.json` as defaults when trying to discover Esy projects, since Esy does in fact look at those two files by default.  However, the drop down list should also allow someone to select a specific file if they want to try to configure a different sandbox.

EduardoRFS:
I'm also using esy to test .opam projects.

## detection

The best detection that came to my mind is not detecting the manifest, but actually detecting the lock, inside the lock you will have the lock manifest(`index.json`), and with this you can know which command to run based on the `root` field in the lock.

Also you should always follows the lock, if the lock says you should run `esy -P .` you shouldn't select a specific manifest by default, unless the user manually selects it, `esy -P ./mirage-crypto.opam`  and `esy -P .` are different commands and will generate different locks, even if it is using the same manifest.

### examples

package.json: `hello-reason@link-dev:./package.json`
command: `esy -P ./package.json`

linux.musl.x86_64.json: `linux.musl.x86_64@link-dev:./linux.musl.x86_64.json`
command: `esy -P ./linux.musl.x86_64.json`

single opam: `mirage-crypto@link-dev:./mirage-crypto.opam`
command: `esy  -P ./mirage-crypto.opam`

folder opam: `mirage-crypto@no-source:@d41d8cd9`
command: `esy  -P .`

### problems

This will not detect manifests not initialized, so if you remove the `esy.lock` it will not show to you as an option until the user run's `esy`, but you can always default for detection on `esy.json` and `package.json` if no lock is available

There is some magic involved in `esy @` as it will look for manifest in the parent directory, so you would also want to look recursively on parent directories for locks

## Default

If there is a single lock in the current folder and no setting was selected, using `esy -P .` seems reasonable. Then if it fails showing the list for the user to select from opam or a parent lock is the best option.

Also the `root` in .vscode should be relative by default, like `${workspaceFolder}/package.json`, otherwise you cannot commit your .vscode folder

rgrinberg:
Closing as this PR is out of date. @mnxn feel free to re-open once you update it.

PR #416, opened 2020-10-20
Remove styling tag from language

cmxa:
The `strong` tag is intended for markup languages (e.g., AsciiDoc, Markdown) in which the language explicitly says to be bold. Whether or not a unit literal, empty array, or empty list should be bold should instead be determined by the theme, not the OCaml syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## Unreleased
+## 1.3.3
 
 - Fix sandbox functionality when a folder is not opened (#409)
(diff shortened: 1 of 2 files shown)


mnxn:
Needs a change log entry.

> Whether or not a unit literal, empty array, or empty list should be bold should instead be determined by the theme.

The literals needs distinct names like `constant.language.unit.ocaml` and `constant.language.list.ocaml` for a user to adjust their theme for these literals.

mnxn:
@cmxa: Sorry, I forced pushed the wrong commit by accident and lost push access.

PR #424, opened 2020-11-01
Allow using workspaceFolder placeholder variable for more portable sandbox configurations

### LINKED ISSUE #374: settings.json::ocaml.sandbox.root written with absolute path vs relative path breaks project portability
# Problem

Absolute paths in settings.json did not work with project on a different computer.

E.g.

```
{
  "ocaml.sandbox": {
    "kind": "esy",
    "root": "/Users/cdaringe/src/freshawair"
  }
}
```

# Context

The only difference between my two computers  is `/Users` and `/home`--pretty similar!--so it took me a few to figure out what was up :). I'm _pretty sure_ I didn't enter in the `root` field in settings.json, but I can't quite pin down where it gets written in this extension.

# Discussion

Perhahps the `"${workspaceFolder}"` placeholder var should be used instead when creating the sandbox path?

mnxn:
Closes #374

Allows using `${workspaceFolder}` for the sandbox setting's manifest/switch/template. Since this only replaces the variable, it's done during the JSON serialization/deserialization. Relative paths are still not supported, so no command execution logic was changed.

In a workspace editor, it resolves `${workspaceFolder}` with the path of the first workspace folder. If no workspace folders are open, `${workspaceFolder}` is resolved with the value of `Process.cwd`.

When selecting a sandbox, if the path of the workspace folder appears in a discovered configuration, it will be substituted with `${workspaceFolder}`. This makes the sandbox configurations portable by default.

### DIFF
--- a/src/settings.mli
+++ b/src/settings.mli
@@ -14,3 +14,9 @@ val create :
   -> 'a t
 
 val string : scope:ConfigurationTarget.t -> key:string -> string t
+
+(** replace ${workspaceFolder} with the first workspace folder that is open *)
+val resolve_workspace_var : string -> string
+
+(** replace the path of the first open workspace folder with ${workspaceFolder} *)
+val substitute_workspace_var : string -> string
--- a/src/settings.mli
+++ b/src/settings.mli
@@ -15,8 +15,8 @@ val create :
 
 val string : scope:ConfigurationTarget.t -> key:string -> string t
 
-(** replace ${workspaceFolder} with the first workspace folder that is open *)
-val resolve_workspace_var : string -> string
+(** replace ${workspaceFolder:folder_name} variables with workspace folder paths *)
+val resolve_workspace_vars : string -> string
 
-(** replace the path of the first open workspace folder with ${workspaceFolder} *)
-val substitute_workspace_var : string -> string
+(** replace workspace folder paths with ${workspaceFolder:folder_name} variables *)
+val substitute_workspace_vars : string -> string
--- a/vscode/interop/interop.mli
+++ b/vscode/interop/interop.mli
@@ -18,6 +18,10 @@ module Regexp : sig
   val t_to_js : Js_of_ocaml.Regexp.regexp -> Ojs.t
 
   val t_of_js : Ojs.t -> Js_of_ocaml.Regexp.regexp
+
+  type replacer = string -> string list -> string
+
+  val replace : string -> regexp:t -> replacer:replacer -> string
 end
 
 module Dict : sig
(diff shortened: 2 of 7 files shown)


src/settings.mli
```diff
@@ -14,3 +14,9 @@ val create :
   -> 'a t
 
 val string : scope:ConfigurationTarget.t -> key:string -> string t
+
+(** replace ${workspaceFolder} with the first workspace folder that is open *)
```
rgrinberg, line 18: That does not match the semantics described [here](https://code.visualstudio.com/docs/editor/multi-root-workspaces):

> used in a configuration (for example ${workspaceFolder} or the now deprecated ${workspaceRoot}) are resolved relative to the folder they belong to. It is possible to scope a variable per workspace folder by appending the root folder's name to a variable (separated by a colon).

mnxn: This is really annoying since the VSCode API doesn't provide a way to get the workspace folder of the configuration file.

Since the sandbox configurations are primarily changed through commands, we could do what the C++ and CMake extension do and only support named workspace folder variables (`${workspaceFolder:name}`):

https://github.com/microsoft/vscode-cpptools/blob/839eefa0325a718b7feefc737eaf9d3bdae18a6a/Extension/src/common.ts#L339

https://github.com/microsoft/vscode-cmake-tools/blob/6b30c8f0812c5a5fcc2458544991a200c05d9491/src/expand.ts#L112

Eventually I would hope that VSCode gets native support for settings variables: https://github.com/microsoft/vscode/issues/2809.

rgrinberg: Indeed, the api doesn't seem well thought out.

> Since the sandbox configurations are primarily changed through commands, we could do what the C++ and CMake extension do and only support named workspace folder variables (${workspaceFolder:name}):

This workaround sounds good to me.

rgrinberg (approved):

vscode/interop/interop.mli
```diff
@@ -18,6 +18,10 @@ module Regexp : sig
   val t_to_js : Js_of_ocaml.Regexp.regexp -> Ojs.t
 
   val t_of_js : Ojs.t -> Js_of_ocaml.Regexp.regexp
+
+  type replacer = string -> string list -> string
+
+  val replace : string -> regexp:t -> replacer:replacer -> string
```
rgrinberg, line 24: Surprised that this binding doesn't exist in jsoo. You should consider contributing it.

mnxn: This function is difficult to type correctly. It includes a variable number of string parameters with different parameter types after that: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter

Now that I've revisited it, I think this binding is unsafe due to the trailing parameters that were omitted.

I'll try to contribute it to Js_of_ocaml if I can get a gen_js_api version working first.

PR #426, opened 2020-11-02
Only restart language server for ocaml.server.restart

mnxn:
Suggested by @ulugbekna

Previously `ocaml.server.restart` would needlessly dispose and recreate the status bar item, dune formatter, and dune task provider when all it needs to do is restart the language server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Stop highlighting ocaml unit/array/list literals with bold (#416)
 - Add a snippet `struct end` with prefix `struct` (#420)
+- Only restart the language server for the `ocaml.server.restart` command (#426)
 
 ## 1.3.3
 
--- a/src/vscode_ocaml_platform.ml
+++ b/src/vscode_ocaml_platform.ml
@@ -57,6 +57,11 @@ module Instance = struct
     ; dune_task_provider = Dune_task_provider.create ()
     }
 
+  let stop_language_server t =
+    Option.iter t.client ~f:(fun client ->
+        LanguageClient.stop client;
+        t.client <- None)
+
   let stop t =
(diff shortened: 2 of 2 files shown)


ulugbekna:
Thanks for your PR! 

The idea behind my proposal was to simplify the extension's state: if we don't need to restart the extension, we can make all providers (such as `DuneTaskProvider` and `DuneFormatter`) to be instantiated when the extension is first run, which means we don't need to have them in the extension state (so we remove some record fields) and can put their `Disposable`s in the execution context, so that we don't dispose them manually. So, this is a great opportunity to simplify other things: remove state from `DuneTaskProvider` and `DuneFormatter` modules (and maybe other similar modules).

On the other hand, since my proposal, I thought more about whole extension restart. Could it be useful for when the extension is in an "incorrect/corrupted" state and the user wants to restart it? The other way to restart would be to restart vscode.

mnxn:
All three features (status bar item, dune formatter, and dune task provider) depend on the toolchain value to execute commands or display information. The toolchain value changes every time the user changes the sandbox, which causes the three fields to be recreated for the new toolchain.

In other words, although this PR reduces the amount of mutation happening for [`restart_instance`](https://github.com/ocamllabs/vscode-ocaml-platform/blob/b74950281c7876ec21aa6da42e823b0adf7be09f/src/vscode_ocaml_platform.ml#L158), mutable state is still necessary for [`select_sandbox`](https://github.com/ocamllabs/vscode-ocaml-platform/blob/b74950281c7876ec21aa6da42e823b0adf7be09f/src/vscode_ocaml_platform.ml#L142).


I don't see how it's possible to remove mutable state given this fact.

rgrinberg (approved):
Looks good. Don't forget the CHANGES entry.

PR #429, opened 2020-11-04
Support error messages without characters in problem matcher

### LINKED ISSUE #425: Linting mli Files
One issue I run into a lot while using the plugin is that the linter doesn't catch mli vs. ml type mismatches.
Just for an example:
consts.ml:
```ocaml 
let g = 9.81
```
consts.mli:
```ocaml 
val g : string
```

The linter shows no error in consts.mli, but when I run dune build, it throws:

File "virtuality2d/src/consts.ml", line 1:
Error: The implementation virtuality2d/src/consts.pp.ml
       does not match the interface virtuality2d/src/.virtuality2d.objs/byte/virtuality2d__Consts.cmi:
       Values do not match: val g : float is not included in val g : string
       File "virtuality2d/src/consts.mli", line 1, characters 0-14:
         Expected declaration
       File "virtuality2d/src/consts.ml", line 1, characters 4-5:
         Actual declaration

Is there any way that Merlin VS code plugin could catch this in the linter?

Khady:
fix #425 

The error message about an implementation file not matching its signature
does not contain characters. The characters group of the problem match
must hence be optional.

```
File "virtuality2d/src/consts.ml", line 1:
Error: The implementation virtuality2d/src/consts.pp.ml
does not match the interface virtuality2d/src/.virtuality2d.objs/byte/virtuality2d__Consts.cmi:
Values do not match: val g : float is not included in val g : string
File "virtuality2d/src/consts.mli", line 1, characters 0-14:
Expected declaration
File "virtuality2d/src/consts.ml", line 1, characters 4-5:
Actual declaration
```

Note that I have only tested the regex in https://regex101.com/r/13ldQG/1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   languages in VS Code (#428)
 - Allow using `${workspaceFolder:folder_name}` placeholder variables in sandbox
   configurations for portable settings.json files (#424)
+- Fix OCaml problem matcher for warning codes and error messages without
+  characters (#429)
 
 ## 1.3.3
 
--- a/package.json
+++ b/package.json
@@ -197,7 +197,7 @@
         ],
         "pattern": [
           {
-            "regexp": "^\\s*\\bFile\\b\\s*\"(.*)\",\\s*\\blines?\\b\\s*(\\d+)(?:-(\\d+))?,\\s*\\bcharacters\\b\\s*(\\d+)-(\\d+)\\s*:\\s*$",
+            "regexp": "^\\s*\\bFile\\b\\s*\"(.*)\",\\s*\\blines?\\b\\s*(\\d+)(?:-(\\d+))?(?:,\\s*\\bcharacters\\b\\s*(\\d+)-(\\d+)\\s*)?:\\s*$",
             "file": 1,
             "line": 2,
             "endLine": 3,
--- a/package.json
+++ b/package.json
@@ -205,7 +205,7 @@
             "endColumn": 5
           },
           {
-            "regexp": "^(?:\\s*\\bParse\\b\\s*)?\\s*\\b([Ee]rror|Warning)\\b\\s*(?:\\(\\s*\\bwarning\\b\\s*(\\d+)\\))?\\s*:\\s*(.*)$",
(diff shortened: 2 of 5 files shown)


rgrinberg:
Looks good. We should add automatic tests for this stuff though. @mnxn do you mind tackling this?

mnxn:
> Looks good. We should add automatic tests for this stuff though. @mnxn do you mind tackling this?

Sure. I can add tests in a couple of days when I have time to work on it.

It should be pretty straightforward as all it requires is adding a few fixtures and checking for matched problems with `vscode.languages.getDiagnostics`.

rgrinberg:
> It should be pretty straightforward as all it requires is adding a few fixtures and checking for matched problems with vscode.languages.getDiagnostics.

Sounds even better than what I had in mind. My idea was just to run the regular expressions and see if they match the expected output.

mnxn:
The tests are failing for some reason although they work locally. I'm not sure what the problem is, but I think it's related to #283.

Tests that inconsistently fail don't seem very useful for me. @rgrinberg, what do you suggest we do?

rgrinberg:
I guess we should just avoid using the vscode api.

Write the test in OCaml and compile it to JS. Turn the problem matcher into a regular expression, and then make sure it matches some the error from the issue. Clearly, this isn't great, but it's better than nothing. At least we can start building a regression test bed of the various errors that didn't match.

Khady:
https://github.com/ocaml/tuareg/blob/c1d245a62455a78a0da96b5f79da50237aeb2474/compilation.txt

This file can be used as a source of inspiration for the messages to test.

mnxn:
I updated the problem matcher regex because the old regex wasn't capturing warning codes from warnings like
```
Warning 26: unused variable y.
```

The tests should hopefully be passing now.

rgrinberg (approved):
Looks good. I'm curious as to why you didn't just write the tests in OCaml. As far as I can see, you aren't using the vscode api, and it would have been just as easy to write the tests in jsoo.

mnxn:
It would require writing more bindings and I'm not a fan of porting the tests in pieces. I'd much rather rewrite the tests in OCaml all at once.

PR #437, opened 2020-11-18
Infer interface when switching to an non-existing interface

tmattio:
This PR changes the behavior of the switch implementation-interface feature to initialize the interface with an inferred one when switching to a non-existing file.

It depends on https://github.com/ocaml/ocaml-lsp/pull/308.

I implemented a custom request `ocamllsp/inferIntf` in `ocaml-lsp`, because running a codeAction automatically did not seem like a good way to implement this. The code action is still in place though: when switching to an existing interface, we can select some text and replace it with the inferred interface.

### DIFF
--- a/src/extension_instance.ml
+++ b/src/extension_instance.ml
@@ -57,7 +57,8 @@ let start_language_server toolchain =
   let ocaml_lsp = Ocaml_lsp.of_initialize_result initialize_result in
   if
     (not (Ocaml_lsp.has_interface_specific_lang_id ocaml_lsp))
-    || not (Ocaml_lsp.can_handle_switch_impl_intf ocaml_lsp)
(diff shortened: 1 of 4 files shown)


rgrinberg:
This is no longer relevant I believe.

tmattio:
I experimented with `workspace/applyEdit` but did not find a conclusive solution with it, so I believe a custom request remains our best option for this feature. The PR is here: https://github.com/ocaml/ocaml-lsp/pull/314. 

I'll reopen until we decide to merge it, or we have a better solution.

tmattio:
ocaml/ocaml-lsp#314 is merged, so this is good to go if you don't have any concerns.

mnxn (approved):

PR #439, opened 2020-11-19
Add workflow step to publish the extension to Open VSX

### LINKED ISSUE #438: Publish to open-vsx in CI
open-vsx is used by editors like VSCodium instead of the VSCode extension marketplace.

The process is similar to using the `vsce` tool, but requires another token:
https://github.com/eclipse/openvsx/wiki/Publishing-Extensions

smorimoto:
Closes #438

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -111,7 +111,9 @@ jobs:
 
       - name: Publish extension to Open VSX
         if: success() && runner.os == 'Linux' && contains(github.ref, 'refs/tags')
-        run: yarn deploy:ovsx ocaml-platform-*.vsix --pat ${{ secrets.OPEN_VSX_TOKEN }}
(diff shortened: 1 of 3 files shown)


smorimoto:
Note that we need to add a secret named OPEN_VSX_TOKEN to this repository.

mnxn (approved):

rgrinberg:
> Note that we need to add a secret named OPEN_VSX_TOKEN to this repository.

Where do I get this secret from?

mnxn:
> Where do I get this secret from?

You can get it from [open-vsx.org](https://open-vsx.org) - sign in with GitHub and create an access token in settings.

The process is also described here: https://github.com/eclipse/openvsx/wiki/Publishing-Extensions#1-create-an-access-token

mnxn:
I edited the workflow so it reads the token from an environmental variables. I don't think it changes anything, but it is the recommended way in the documentation: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#example-using-bash

PR #440, opened 2020-11-19
Redesign extension state (Instance)

ulugbekna:
This is a slightly large change to how to store and manage extension state (was called `Instance` and is called`Extension_instance` now). It removes some mutable state, avoids restarting the instance when we can just update it, and does several other small things. 

For reviewing: I would recommend to go commit by commit from oldest to youngest, and the code should make sense.

### DIFF
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -23,7 +23,7 @@ let get_formatter toolchain ~document ~options:_ ~token:_ =
     match output with
     | Ok newText -> Some [ TextEdit.replace ~range ~newText ]
     | Error msg ->
-      message `Error "Dune formatting failed: %s" msg;
+      show_message `Error "Dune formatting failed: %s" msg;
       Some []
   in
   `Promise promise
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -14,7 +14,7 @@ let get_formatter instance ~document ~options:_ ~token:_ =
   let command =
     let toolchain =
       (* we're gonna remove [value_exn] use later (we refactor instance creation) *)
-      Option.value_exn instance.Extension_instance.toolchain
+      instance.Extension_instance.toolchain
     in
     Toolchain.get_dune_command toolchain [ "format-dune-file" ]
   in
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -14,7 +14,7 @@ let get_formatter instance ~document ~options:_ ~token:_ =
   let command =
     let toolchain =
       (* we're gonna remove [value_exn] use later (we refactor instance creation) *)
-      instance.Extension_instance.toolchain
+      Extension_instance.toolchain instance
     in
     Toolchain.get_dune_command toolchain [ "format-dune-file" ]
   in
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -13,7 +13,6 @@ let get_formatter instance ~document ~options:_ ~token:_ =
   let document_text = TextDocument.getText document ~range () in
   let command =
     let toolchain =
-      (* we're gonna remove [value_exn] use later (we refactor instance creation) *)
       Extension_instance.toolchain instance
     in
     Toolchain.get_dune_command toolchain [ "format-dune-file" ]
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -12,9 +12,7 @@ let get_formatter instance ~document ~options:_ ~token:_ =
   (* text of entire document *)
   let document_text = TextDocument.getText document ~range () in
   let command =
-    let toolchain =
-      Extension_instance.toolchain instance
-    in
+    let toolchain = Extension_instance.toolchain instance in
     Toolchain.get_dune_command toolchain [ "format-dune-file" ]
   in
   let output =
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -1,6 +1,6 @@
 open Import
 
-let get_formatter toolchain ~document ~options:_ ~token:_ =
+let get_formatter instance ~document ~options:_ ~token:_ =
   let endLine = TextDocument.lineCount document - 1 in
   let endCharacter =
     TextDocument.lineAt document ~line:endLine |> TextLine.text |> String.length
@@ -11,7 +11,13 @@ let get_formatter toolchain ~document ~options:_ ~token:_ =
   in
   (* text of entire document *)
   let document_text = TextDocument.getText document ~range () in
-  let command = Toolchain.get_dune_command toolchain [ "format-dune-file" ] in
+  let command =
+    let toolchain =
+      (* we're gonna remove [value_exn] use later (we refactor instance creation) *)
+      Option.value_exn instance.Extension_instance.toolchain
+    in
+    Toolchain.get_dune_command toolchain [ "format-dune-file" ]
+  in
   let output =
     let open Promise.Result.Syntax in
     let* command = Cmd.check command in
@@ -28,23 +34,14 @@ let get_formatter toolchain ~document ~options:_ ~token:_ =
   in
   `Promise promise
 
-type t = Disposable.t list ref
-
-let create () = ref []
-
-let register t toolchain =
-  t :=
-    [ "dune"; "dune-project"; "dune-workspace" ]
-    |> List.map ~f:(fun language ->
-           let selector =
-             `Filter (DocumentFilter.create ~scheme:"file" ~language ())
-           in
-           let provider =
-             DocumentFormattingEditProvider.create
-               ~provideDocumentFormattingEdits:(get_formatter toolchain)
-           in
-           Languages.registerDocumentFormattingEditProvider ~selector ~provider)
-
-let dispose t =
-  List.iter ~f:Disposable.dispose !t;
-  t := []
+let register instance =
+  [ "dune"; "dune-project"; "dune-workspace" ]
+  |> List.map ~f:(fun language ->
+         let selector =
+           `Filter (DocumentFilter.create ~scheme:"file" ~language ())
+         in
+         let provider =
+           DocumentFormattingEditProvider.create
+             ~provideDocumentFormattingEdits:(get_formatter instance)
+         in
+         Languages.registerDocumentFormattingEditProvider ~selector ~provider)
--- a/src/dune_formatter.ml
+++ b/src/dune_formatter.ml
@@ -34,7 +34,7 @@ let get_formatter instance ~document ~options:_ ~token:_ =
   in
   `Promise promise
 
-let register instance =
+let register extension instance =
   [ "dune"; "dune-project"; "dune-workspace" ]
   |> List.map ~f:(fun language ->
          let selector =
@@ -45,3 +45,5 @@ let register instance =
              ~provideDocumentFormattingEdits:(get_formatter instance)
          in
          Languages.registerDocumentFormattingEditProvider ~selector ~provider)
+  |> List.iter ~f:(fun disposable ->
+         ExtensionContext.subscribe extension ~disposable)
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -72,7 +72,7 @@ let switch_impl_intf =
       let document = TextEditor.document editor in
       let* client = instance.client in
       (* extension needs to be activated; otherwise, just ignore the switch try *)
-      let+ ocaml_lsp = instance.ocaml_lsp_capabilities in
+      let+ ocaml_lsp = instance.ocaml_lsp in
       (* same as for instance.client; ignore the try if it's None *)
       if Ocaml_lsp.can_handle_switch_impl_intf ocaml_lsp then
         Switch_impl_intf.request_switch client document
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -1,0 +1,104 @@
+open Import
+
+type command =
+  { id : string
+  ; handler : Extension_instance.t -> unit -> unit
+  }
+
+let command id handler = { id; handler }
+
+let select_sandbox =
+  let handler (instance : Extension_instance.t) () =
+    let set_toolchain =
+      let open Promise.Syntax in
+      let* package_manager = Toolchain.select_and_save () in
+      match package_manager with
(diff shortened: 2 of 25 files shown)


src/toolchain.ml
```diff
@@ -137,9 +137,11 @@ module Package_manager = struct
     | Custom _ -> "Custom OCaml"
 end
 
-type resources = Package_manager.t
+type t = { package_manager : Package_manager.t }
```
rgrinberg, line 148: The new name is good, but we don't need this extra record.

src/vscode_ocaml_platform.ml
```diff
@@ -138,72 +130,110 @@ module Instance = struct
...
-    | Some pm ->
-      Instance.stop instance;
-      let t = Toolchain.make pm in
-      Instance.start instance t
-  in
-  let (_ : unit Promise.t) =
-    Promise.Result.iter set_toolchain ~error:(message `Error "%s")
-  in
-  ()
+module ExtensionCommands = struct
```
rgrinberg: We don't use CamelCase for our own code. Only when wrapping external code.

src/dune_formatter.ml
```diff
@@ -23,7 +23,7 @@ let get_formatter toolchain ~document ~options:_ ~token:_ =
     match output with
     | Ok newText -> Some [ TextEdit.replace ~range ~newText ]
     | Error msg ->
-      message `Error "Dune formatting failed: %s" msg;
+      show_message `Error "Dune formatting failed: %s" msg;
```
rgrinberg, line 32: Please revert this. No additional clarity is gained here.

src/toolchain.ml
```diff
@@ -110,6 +110,14 @@ module Package_manager = struct
     let t = Settings.create ~scope:Workspace ~key:"sandbox" ~of_json ~to_json
   end
 
+  let equal pm1 pm2 =
```
rgrinberg, line 113: This is not a general purpose equality function, so it's quite confusing to call it that.

src/vscode_ocaml_platform.ml
```diff
@@ -1,263 +1,69 @@
...
+         in
+         let _register_dune_formatter : unit =
+           Dune_formatter.register instance
+           |> List.iter ~f:(fun disposable ->
+                  ExtensionContext.subscribe extension ~disposable)
+         in
+         let _register_dune_task_provider : unit =
+           let disposable = Dune_task_provider.register instance in
+           ExtensionContext.subscribe extension ~disposable
+         in
```
mnxn, line 50: I don't see the point of giving names to these bindings if they're going to be unused. I would just separate the statements with semicolons.

src/extension_commands.ml
```diff
@@ -0,0 +1,108 @@
+open Import
```
rgrinberg, line 1: This module has no signature.

src/extension_instance.mli
```diff
@@ -0,0 +1,20 @@
+open Import
+
+type t =
```
rgrinberg, line 3: Is it possible to make this type abstract?  I'd prefer if mutability wasn't part of the public api.

src/vscode_ocaml_platform.ml
```diff
@@ -1,263 +1,69 @@
...
+           Extension_commands.register_all_commands extension instance
+         in
+         let _register_extension_instance : unit =
+           ExtensionContext.subscribe extension
+             ~disposable:(Extension_instance.disposable instance)
+         in
+         let _register_dune_formatter : unit =
+           Dune_formatter.register instance
+           |> List.iter ~f:(fun disposable ->
+                  ExtensionContext.subscribe extension ~disposable)
```
mnxn, line 45: Let's not expose the list of formatters outside of the Dune_formatter module. We can abstract it away as
```ocaml
let dune_formatter = Dune_formatter.register instance in
ExtensionContext.subscribe extension
  ~disposable:(Dune_formatter.disposable dune_formatter)
```

src/toolchain.ml
ulugbekna: I wanted to avoid type aliasing for reasons:
- to avoid lower-quality type-on-hover as is the case with `Cmd` module `type stderr = string` and `type stdout = string`, which ends up making all strings `stderr` (this is less acute here, but I'd love to keep it as it is
- having `package_manager` as a part of `Toolchain.t` makes it (1) easily extensible and (2) shows that a `package_manager` is only a _part_ of what a toolchain is

Please let's keep it :-)

src/extension_instance.ml
```diff
@@ -0,0 +1,130 @@
...
+  let make toolchain =
+    let status_bar_item =
+      Window.createStatusBarItem ~alignment:StatusBarAlignment.Left ()
+    in
+    let status_bar_item_text =
+      make_status_bar_item_text @@ Toolchain.package_manager toolchain
+    in
+    StatusBarItem.set_text status_bar_item status_bar_item_text;
+    StatusBarItem.set_command status_bar_item
+      (`String select_sandbox_command_id);
```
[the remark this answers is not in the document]
ulugbekna: No, that causes depedency cycle: commands use instance and instance would be using `command.id`

src/extension_instance.ml
```diff
@@ -0,0 +1,130 @@
...
+    StatusBarItem.show status_bar_item;
+    status_bar_item
+
+  let _update sandbox_info ~new_toolchain =
```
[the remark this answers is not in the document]
ulugbekna: oh, thanks for noting, I'll remove it; it was there because `update` wasn't being used at the time

src/toolchain.ml
rgrinberg: Okay, you can keep because of the hovers.

I don't think it's worth it to complicate current code to make it easier to *possibly* add more code in the future. So making things extensible should have a very concrete extension in mind. OCaml already makes it easy to refactor, there's no need to leave unused hooks for extension everywhere.

src/dune_formatter.ml
ulugbekna: I can for sure revert this but would love to keep it because at the time `message` was confusing to me and may be confusing to others because it doesn't say what kind of message that is:

- lsp and extension use messages to communicate; is that the `message`? 
- does `message` simply log somewhere?

Having either `popup_message` or `show_message` hints much more about the fact that it's being displayed.

(don't like taking your time to argue about naming; will revert if you think that's better)

src/toolchain.ml
mnxn: We might want to remove the distinction between the two types entirely. I think there's a historical reason why there are two different types, but currently I feel like it complicates the code by requiring conversions between them.

src/toolchain.ml
ulugbekna: Didn't quite get what you mean. Isn't `M.equal: M.t -> M.t -> bool` a common function defined for modules?

src/vscode_ocaml_platform.ml
ulugbekna: you're right, but with names it looks a little more structured and declarative; if not bindings, I would at least comment these, but again with bindings it looks more structured

```ocaml
         let _register_extension_commands : unit =
           Extension_commands.register_all_commands extension instance
         in
         let _register_extension_instance : unit =
           ExtensionContext.subscribe extension
             ~disposable:(Extension_instance.disposable instance)
         in
         let _register_dune_formatter : unit =
           Dune_formatter.register instance
           |> List.iter ~f:(fun disposable ->
                  ExtensionContext.subscribe extension ~disposable)
         in
         let _register_dune_task_provider : unit =
           let disposable = Dune_task_provider.register instance in
           ExtensionContext.subscribe extension ~disposable
```

vs 

```ocaml
           Extension_commands.register_all_commands extension instance;
           ExtensionContext.subscribe extension
             ~disposable:(Extension_instance.disposable instance);
           Dune_formatter.register instance
           |> List.iter ~f:(fun disposable ->
                  ExtensionContext.subscribe extension ~disposable);
           let disposable = Dune_task_provider.register instance in
           ExtensionContext.subscribe extension ~disposable;
```

src/toolchain.ml
rgrinberg: `equal` is the common name for structural equality (or something that is morally equivalent). What you're implementing here can give us equality for:

```ocaml
let x = Opam (Opam.make (), Named "foo"))
let y = Opam (Opam.make (), Named "bar"))
```

Even if the `Opam.t` returned isn't equal.

There's nothing wrong with having a function like that, but it shouldn't be called `equal`.

src/vscode_ocaml_platform.ml
mnxn: With some API changes, you can write
```ocaml
Extension_commands.register_all_commands extension instance;

ExtensionContext.subscribe extension
    ~disposable:(Extension_instance.disposable instance);
    
let dune_formatter = Dune_formatter.register instance in
ExtensionContext.subscribe extension ~disposable:dune_formatter in

let dune_task_provider = Dune_task_provider.register instance in
ExtensionContext.subscribe extension ~disposable:dune_task_provider;
```

Which is perfectly readable in my opinion.

src/toolchain.ml
ulugbekna: > We might want to remove the distinction between the two types entirely. I think there's a historical reason why there are two different types, but currently I feel like it complicates the code by requiring conversions between them.

I thought about it but I'm yet to come up with proper abstractions, and I didn't want to change _too_ much.

mnxn: In that case, lets revert the Toolchain changes and handle it in a separate PR.

src/dune_formatter.ml
rgrinberg: > does message simply log somewhere?

If it logged, it would be called "log".

The purpose of this alias was to avoid the annoyingly verbose `show{Information,Error,Debug}Message` functions. If our alias is just as verbose, then I don't see the point of it.

Indeed, I don't have much time to argue about this. Feel free to keep whatever name you like

src/extension_instance.ml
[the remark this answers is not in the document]
rgrinberg: With, or without the cycle, this identifier can't live in `Import`. `Import` is for non application specific things.

Perhaps, you could split `Extension_commands` into two modules:

* Vscode_command : which just defines the type + helpers
* Extension_commands : which defines the commands in our extension.

ulugbekna:
I think I addressed all of the comments.

tmattio:
I'm starting work on the detection of the sandbox and it seems sensible to do it on top of this PR.

Is there anything blocking to merge it? The changes make complete sense IMHO and the remaining work from the review (i.e. merge the toolchain and package manager types) can be done in a separate PR I think.

src/extension_consts.ml
```diff
@@ -0,0 +1,15 @@
...
+  let restart_language_server = ocaml_prefixed "server.restart"
+
+  let select_sandbox_and_open_terminal = ocaml_prefixed "open-terminal-select"
+
+  let open_terminal = ocaml_prefixed "open-terminal"
+
+  let switch_impl_intf = ocaml_prefixed "switch-impl-intf"
+end
+
+(* TODO: add other constants such as settings keywords etc *)
```
rgrinberg, line 15: I have a strong dislike for modules that just group unrelated constants. I prefer to encapsulate constants in logical modules as much as possible. I realize that we haven't come up with a good alternative here yet, so this shouldn't block the PR from being merged. We definitely don't want to make the situation worse in the future.

rgrinberg (approved):
I didn't do a full review, but it seems to be good if it addresses all the comments. If @mnxn approves, then it's good to merge.

src/extension_instance.ml
```diff
@@ -0,0 +1,143 @@
...
+let disposable t =
+  Disposable.make ~dispose:(fun () ->
+      (* [stop] is not defined in the toplevel module because
+         we don't want anyone to call it, only vscode when the extension is no longer needed;
+         a user calling [stop] would leave the extension in a corrupt state *)
+      let stop instance =
+        StatusBarItem.dispose instance.sandbox_info;
+        LanguageClient.stop instance.client
+      in
+      stop t)
```
mnxn, line 143: I think it's unnecessary to define a local function inside of a lambda that's only used once.

```suggestion
  Disposable.make ~dispose:(fun () ->
      StatusBarItem.dispose instance.sandbox_info;
      LanguageClient.stop instance.client)
```

ulugbekna: For readability :-) Feel free to commit suggestion if you want

src/toolchain.ml
mnxn: I'd still like this change to be reverted, since the `Toolchain` module will be revisited in another PR.

src/extension_consts.ml
ulugbekna: I also have mixed feelings about non-logically related constants, but 
1. This may be a good thing, e.g., almost all android applications extract out strings in a separate file, so that user-visible stuff such as warning messages can be easily changed (+ i18n but not our case :-))
2. I don't have alternatives either as you say :-)

src/toolchain.ml
ulugbekna: I provided specific arguments why I wouldn't like to have a type alias with which Rudi, I think, is okay. 

I don't want to revert and redo things later just for the sake of having another PR. If you want to do that, go ahead.

src/dune_formatter.ml
```diff
@@ -11,7 +11,13 @@ let get_formatter toolchain ~document ~options:_ ~token:_ =
   in
   (* text of entire document *)
   let document_text = TextDocument.getText document ~range () in
-  let command = Toolchain.get_dune_command toolchain [ "format-dune-file" ] in
+  let command =
+    let toolchain =
+      (* we're gonna remove [value_exn] use later (we refactor instance creation) *)
```
ulugbekna, line 16:
```suggestion
```

mnxn (approved):
I think this is ready to merge now.

> I'm starting work on the detection of the sandbox and it seems sensible to do it on top of this PR.

@tmattio: I suggest you base your work on top of the next PR I submit, which will simplify the Toolchain/Package_manager types.

tmattio:
@mnxn Sure - I already started working on it here: https://github.com/ulugbekna/vscode-ocaml-platform/compare/redesign-extension-state...tmattio:sandbox-detection but I can wait for your PR to submit it if you are planning on doing it soon 👍

PR #444, opened 2020-11-23
Unify toolchain and package manager types

mnxn:
This PR removes the `Package_manager` module and directly integrates its components in the `Toolchain` module.

The separate types (`Toolchain.t` and `Package_manager.t`) have only caused unnecessary conversions between each other, without providing additional functionality.

This PR also cleans up the use of language internally to refer to "toolchains", not "package managers" or "sandboxes". The previous language didn't make sense because only the `Opam`/`Esy` variants could be considered "package managers" and the `Custom`/`Global` variants aren't necessarily sandboxed. 

I didn't change any exposed names, so the commands and settings are still the same (`ocaml.select-sandbox` and `ocaml.sandbox`).

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -20,11 +20,10 @@ let select_sandbox =
     let open Promise.Syntax in
     let current_toolchain = Extension_instance.toolchain instance in
     let (_ : unit Promise.t) =
-      let* package_manager = Toolchain.select_sandbox () in
-      match package_manager with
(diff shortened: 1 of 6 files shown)


rgrinberg (approved):
Looks good. The distinction between toolchains and package managers was inherited from early versions of the plugin. I agree that nothing is gained from it now.

mnxn:
@tmattio: you should be able to rebase your work on top of this now.

PR #445, opened 2020-11-24
Implement detection of sandbox

### LINKED ISSUE #296: auto-detect switch; support relative path
Is there something speaking against automatically selecting the opam switch by default?
Taking the result of `opam switch show` for the currently opened vscode workspace should be the right choice 99% of the time and if it's not, it's shown in the status bar and easy to change.

Another issue: I wanted to share my `.vscode/settings.json` but `ocaml.sandbox.switch` does not seem to support relative paths.

### LINKED ISSUE #321: Auto-selecting opam switch doesn't work with opam switch link
When I'm in a directory with a local opam switch, vscode + vscode-ocaml-platform detects it automatically.  However, if I `opam switch link path/to/directory/with/local/switch/` in another path the switch is not detected automatically.

I can manually select the intended local switch but it would be nice if this could be detected automatically, using the information `opam switch link` leaves behind.

### LINKED ISSUE #446: Minimizing sandbox storage in settings.json
I think that we want user experience with the extension be smooth as possible. Generating a new file `settings.json` simply when one opens an OCaml project doesn't seem very smooth to me, e.g., having to add it to .gitignore, or genuinely having to care there is a new folder `.vscode` with a file it. 

One thing we could achieve with automatic sandbox detection is that we can minimize use of settings.json. When the automatic sandbox detection is merged, we don't really need to store the detected sandbox in the settings unless the user explicitly picks another sandbox. Even in that case, if user picks another sandbox to be used, we can `opam switch link` and avoid saving a file. Saving the sandbox in the settings causing additional noise because the same sandbox will be picked next time by the detection algorithm unless user changes their environment, in which case they are likely to want to change their sandbox as well.  

We should think of backward compatibility and, if possible, remove `settings.json` in machines where it is already present: If the automatically detected sandbox based on opam switch show or esy sandbox detection is the same as the one in settings, we could remove that setting at all if there is one.

tmattio:
This PR implements the detection of the sandboxing environment for projects with a single workspace folder.

**TODO**

- [x] Add Esy detection

Fixes #446
Fixes #321
Fixes #296


tmattio:
The commit 83383c8 renames the `Global` toolchain to `System`. After discussing with @ulugbekna, we agreed that `Global` was confusing and did not clearly explain what the toolchain was. `System` seems like a better name, but the goal is more to start the discussion, so very open to other suggestions (and I can split it into a different PR if needed)

ulugbekna:
Thanks for the PR! :-)

1. I would define what `System` represents and when it's used.
2. Think of backward compatibility when dealing with persistence in
settings. Someone may already have `{"kind: "global"}` in their
`.vscode/settings.json`, which the new code I think wouldn't be able to
account for.

mnxn:
For backwards compatibility, we could continue to read "global" kinds from user settings, but warn the user that it has been renamed. Maybe even add a button in the message that lets the user rewrite it from "global" to "system".

src/toolchain.ml
```diff
@@ -181,14 +181,73 @@ let of_settings () : t option Promise.t =
...
       ) )
-  | Some Global -> Promise.return (Some Global)
+  | Some System -> Promise.return (Some System)
   | Some (Custom template) -> Promise.return (Some (Custom template))
 
+let detect_esy_sandbox ~project_root esy () =
+  let open Promise.Syntax in
+  let* esy = esy in
+  let* esy_build_dir_exists = Fs.exists (Filename.concat project_root "_esy") in
+  let* manifest = Esy.find_manifest_in_dir project_root in
```
mnxn, line 191: This can be a `let+` if you remove the following `Promise.return`s.

src/toolchain.ml
```diff
@@ -181,14 +181,73 @@ let of_settings () : t option Promise.t =
...
+let detect_opam_sandbox ~project_root opam () =
+  let open Promise.Syntax in
+  let* opam = opam in
+  match opam with
+  | None -> Promise.return None
+  | Some opam -> (
+    let+ switch = Opam.switch_show ~cwd:project_root opam in
+    match switch with
+    | Some switch -> Some (Opam (opam, switch))
+    | None -> None )
```
mnxn, line 214: Using `Promise.Option.Syntax` would make this code cleaner:
```suggestion
  let open Promise.Option.Syntax in
  let* opam = opam in
  let+ switch = Opam.switch_show ~cwd:project_root opam in
  Opam (opam, switch)
```

src/cmd.mli
```diff
@@ -25,6 +25,7 @@ val check : t -> t Or_error.t Promise.t
 
 val log : ?result:ChildProcess.return -> t -> unit
 
-val output : ?stdin:string -> t -> (stdout, stderr) result Promise.t
+val output :
+  ?cwd:string -> ?stdin:string -> t -> (stdout, stderr) result Promise.t
```
rgrinberg, line 29: Is it possible to change `cwd` to `Path.t`?

src/toolchain.ml
```diff
@@ -181,14 +181,73 @@ let of_settings () : t option Promise.t =
...
+
+let detect () =
+  let rec aux = function
+    | [] -> Promise.return None
+    | el :: rest -> (
+      let open Promise.Syntax in
+      let* sandbox_opt = el () in
+      match sandbox_opt with
+      | Some x -> Promise.return (Some x)
+      | None -> aux rest )
```
rgrinberg, line 224: My preference is against polluting application code with generalized functions. So I would move this into `Promise.List.find_map` or some such.

rgrinberg:
> The commit 83383c8 renames the Global toolchain to System

The name Global was chosen on purpose to avoid collision with how opam uses the word System.

src/opam.ml
```diff
@@ -54,6 +53,16 @@ let switch_list t =
...
 
+let switch_show ?cwd t =
+  let command = Cmd.append t [ "switch"; "show" ] in
+  let open Promise.Syntax in
+  let+ output = Cmd.output ?cwd (Spawn command) in
+  match output with
+  | Error _ ->
+    show_message `Warn "Unable to read the current switch.";
+    None
+  | Ok out -> parse_switch out
```
rgrinberg, line 64: The shorter clause goes on top

src/esy.ml
```diff
@@ -92,6 +92,19 @@ module State = struct
     | Pending
 end
 
+let find_manifest_in_dir dir =
+  let open Promise.Syntax in
+  let esy_file = Filename.concat dir "esy.json" in
```
rgrinberg, line 97: Path module please

src/esy.ml
```diff
@@ -92,6 +92,19 @@ module State = struct
...
+  let package_file = Filename.concat dir "package.json" in
+  let* esy_file_exists = Fs.exists esy_file in
```
rgrinberg, line 99: If you have the `Promise.List.find_map` primitive you could reuse it here:

```ocaml
["package.json"; "esy.json"] |> Promise.List.find_map ~f:(fun ...)
```

It would also save you a disk access when one file is found immediately.

ulugbekna:
re: Global -> System

`Global` _collides_ with opam's named switches, which are also global, and _creates confusion_. We need to be clear about these things, IMO. So I proposed the term `System` to account for a case when we have ocaml installed outside opam, which, I think, is how opam uses this term. 

With this PR in, we'll likely use `Global` (in the sense before this PR) rarely because we're going to get the currently active environment by `opam switch show` (which is always the case if the user has opam installed?) and/or esy sandbox detection. 

----

re: backward compat 

1. Since dealing with settings is the extension's job, and users aren't expected to be editing them manually (we don't have info how to do so publicly, for example), I believe we should rewrite "global" setting automatically but just `showMessageInfo` the user about that just once or, preferably, with a button "never show again". Such persistent state (for never show again button) can be saved to `globalState` (see vscode API).

2. I also proposed Thibaut to minimize use of settings to save toolchain, as I discussed once on slack. I'll make an issue for discussion. 

-----

re: opam vs esy when both detected

I would prioritize opam if both are possible, especially if the detected switch is local.

tmattio:
I reverted the change of `Global` -> `System`. It seems there are multiple concerns regarding the backward compatibility, so I prefer to discuss it in a separate PR.

I addressed the PR reviews and tested with the following:

- Esy sandbox
- Opam local switch
- Opam global switch

It seems to work reliably so ready to merge on my side.

src/toolchain.ml
```diff
@@ -181,14 +181,73 @@ let of_settings () : t option Promise.t =
...
+  let* esy_build_dir_exists = Fs.exists (Filename.concat project_root "_esy") in
+  let* manifest = Esy.find_manifest_in_dir project_root in
+  match (esy, esy_build_dir_exists, manifest) with
+  | Some esy, true, Some manifest ->
+    (* Some projects have both use both Esy and Opam, we can't assume that the user
+       wants to use Esy just by looking for an "esy.json" file.
+
+       Instead, we assume the user wants to use esy if these three conditions are met:
+        - The esy command is in the user's system
+        - There is an _esy build directory in the project root
```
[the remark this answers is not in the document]
tmattio: Condition 2 is actually the only one that would allow us to guess that the user probably wants to use Esy:

- Esy can be used with `package.json`, `esy.json`, or even without any of those and just use the local `.opam` files.
- Lots of projects (e.g. `vscode-ocaml-platform`) support using both Opam and Esy (so we can't assume the user uses Esy simply by checking if the project has an `esy.json` file).

Since it seems difficult to know what the user wants to use by just looking at the files in the project, I think a reliable way of guessing it is to check if the user compiled the project with Esy (which will always generate an `_esy` folder).

A drawback of this approach is that the Esy sandbox won't be selected if the user just open the project for the first time, before compiling it. But in this scenario, it does not seem possible to predict that the user wants to use Esy, and I think we want to prioritize Opam.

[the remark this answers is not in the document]
rgrinberg: Yeah, that's how I viewed things initially as well, but now I'm thinking that it's a mistake to be this conservative.

As you've pointed, the main utility of this feature is when users initially clone a project to get started. This is precisely when condition 2 will not hold. After they've built their projects, it's likely that they've saved their choice in in settings file and no longer care for this detection.

Second of all, we need to consider two pools of users that have installed both esy & opam:
* The group with a strong preference for opam
* The group with a strong preference for esy

I would guess that the 2nd group is much longer, but our extension is mostly optimizing for the 1st group.

To me what seems reasonable is to auto detect and use esy whenever we know it's quite likely the first choice (existence of esy binary + esy.json). And if we hear complaints, we can always add a package manager preference option.

In any case, I don't use esy don't have the motivation to improve things for esy users. So I will not let this block the PR.

src/toolchain.ml
```diff
@@ -184,6 +184,58 @@ let of_settings () : t option Promise.t =
...
+      workspace_folder |> WorkspaceFolder.uri |> Uri.path |> Path.of_string
+    in
+    let available = available_toolchains () in
+    try
+      Promise.List.find_map
+        (fun f -> f ())
+        [ detect_esy_sandbox ~project_root available.esy
+        ; detect_opam_sandbox ~project_root available.opam
+        ]
+    with Assert_failure _ -> Promise.return None )
```
rgrinberg, line 227: Looks like catching `Assert_failure _` is only required because of a bug in promise_jsoo. I've sent a PR to fix it here https://github.com/mnxn/promise_jsoo/pull/1

So let's wait until @mnxn fixes and sends a release.

src/toolchain.ml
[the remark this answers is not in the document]
tmattio: I agree that the scenario to optimize for is when the user opens the project for the first time, and with the proposed approach, we completely fail Esy users for this.

I don't use Esy, so I don't have a strong opinion on when we should choose to use it, as long as Esy users are happy.

However, in order to keep the first group (Opam users) happy, I don't think simply looking for an `esy.json` file is conservative enough: I don't use Esy, but have it in my system because I use it for Spin's integration tests. So with the approach you propose, the sandbox detection wouldn't detect my preference correctly for `vscode-ocaml-platform`, for instance. And I don't think my case is isolated: I figure lots of Opam users have installed Esy, used it once or twice, but will prefer their detected package manager to be Opam.

So it's really a question of where we put the threshold: either we introduce friction for an (unknown) part of Opam users, or we introduce friction for the entirety of Esy userbase. I would lean for the latter, but I'm obviously biased. Will let you make the call: that was my last argument 😄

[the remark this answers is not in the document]
rgrinberg: Let's leave it as is. If an esy user will care enough they will send a PR.

ulugbekna:
These questions were blockers for me, when I worked on sandbox detection. I
came up with this algorithm at the time, which should satisfy both opam and
esy users. Sorry for pseudo-ocaml code

```
let opam_switch = detect_opam_sandbox
if opam_switch is local then (* user 100% wants opam *)
  opam_switch
else
  match detect_esy_sandbox with
  | `Underscore_esy_dir_present
    (* user has initiated esy sandbox and has _esy, which means they want
esy *)
  | `Esy_json_present -> esy
    (* there are projects with esy.json that are used with both esy & opam
       but they should be _quite_ rare; ordinary OCaml user is unaffected *)
  | `Package_json_present
  | `Nothing_detected -> opam
```

ulugbekna:
There is also `esy status` iirc that gives some useful info about the
project such if the deps are pulled etc., but creates _esy if not already
present, so requires some clean up after.

src/toolchain.ml
mnxn: Next release should be available soon: https://github.com/ocaml/opam-repository/pull/17717

rgrinberg: We'll need a constraint to reflect this as well.

src/toolchain.ml
```diff
@@ -184,6 +184,58 @@ let of_settings () : t option Promise.t =
...
+  | [ workspace_folder ] -> (
+    let project_root =
+      workspace_folder |> WorkspaceFolder.uri |> Uri.path |> Path.of_string
+    in
+    let available = available_toolchains () in
+    try
+      Promise.List.find_map
+        (fun f -> f ())
+        [ detect_esy_sandbox ~project_root available.esy
+        ; detect_opam_sandbox ~project_root available.opam
```
rgrinberg, line 225: Actually, I would expect a local opam switch to have the highest precedence in this case. That is, when `$ opam switch show` is set to a local switch in this dir, or it is linked to this dir.

tmattio:
I updated the detection logic to prioritize opam local switches. With this change, I don't think it is necessary to be so conservative in the detection of an Esy sandbox. The logic is not exactly the same as the one @ulugbekna described, and the one you wanted @rgrinberg.

I believe I addressed all of the change requests and everyone is aligned with the detection logic, so it's good to merge I think.

src/esy.ml
```diff
@@ -92,6 +92,20 @@ module State = struct
...
+  let package_file = Path.(dir / "package.json") in
+  try
+    [ esy_file; package_file ]
+    |> Promise.List.find_map (fun path ->
+           let+ file_exists = path |> Path.to_string |> Fs.exists in
+           if file_exists then
+             Some path
+           else
+             None)
+  with Assert_failure _ -> Promise.return None
```
ulugbekna, line 107: Rudi said something about this exception earlier; ignore if resolved

src/toolchain.ml
```diff
@@ -184,6 +184,63 @@ let of_settings () : t option Promise.t =
   | Some Global -> Promise.return (Some Global)
   | Some (Custom template) -> Promise.return (Some (Custom template))
 
+let detect_esy_sandbox ~project_root esy () =
+  let open Promise.Syntax in
+  let* esy = esy in
+  let* esy_build_dir_exists =
+    Fs.exists Path.(project_root / "_esy" |> Path.to_string)
+  in
+  let+ manifest = Esy.find_manifest_in_dir project_root in
```
ulugbekna, line 193: if these operations are independent, they can be run concurrently with `Promise.all`. Or we're running them sequentially, we can avoid the computation of `esy_build_dir_exists` if `esy` is `None`, etc.

src/toolchain.ml
```diff
@@ -184,6 +184,63 @@ let of_settings () : t option Promise.t =
...
+  match (esy, esy_build_dir_exists, manifest) with
+  | Some esy, true, _
+  | Some esy, _, Some _ ->
+    (* Esy can be used with [esy.json], [package.json], or without any of those.
+        So we check if we find an [_esy] directory, which means the user created an Esy sandbox.
+
+       If we don't, but there is an [esy.json] file, we can assume the user wants to use Esy.
+    *)
+    Some (Esy (esy, project_root))
+  | _ -> None
```
ulugbekna, line 203: By implementing the previous comment, we can avoid this large pattern match & wildcard in it.

src/toolchain.ml
```diff
@@ -184,6 +184,63 @@ let of_settings () : t option Promise.t =
...
+
+let detect_opam_local_switch ~project_root opam () =
+  let open Promise.Option.Syntax in
+  let* opam = opam in
+  let* switch = Opam.switch_show ~cwd:project_root opam in
+  match switch with
+  | Local _ as switch -> Promise.Option.return (Opam (opam, switch))
+  | _ -> Promise.return None
```
ulugbekna, line 211: I'd avoid wildcard pattern: line count doesn't change but refactoring is potentially easier.

src/toolchain.ml
```diff
@@ -184,6 +184,63 @@ let of_settings () : t option Promise.t =
...
+  | [ workspace_folder ] ->
+    let project_root =
+      workspace_folder |> WorkspaceFolder.uri |> Uri.path |> Path.of_string
+    in
+    let available = available_toolchains () in
+    Promise.List.find_map
+      (fun f -> f ())
+      [ detect_opam_local_switch ~project_root available.opam
+      ; detect_esy_sandbox ~project_root available.esy
+      ; detect_opam_sandbox ~project_root available.opam
```
ulugbekna, line 231: we're doing repeated work here by detecting the switch twice; we should reuse the detected switch and pattern match on it, imo.

tmattio: The list has types `(unit -> t option Promise.t) list`, and we select the first `Some`, so there should not be any repeated work.

src/esy.ml
tmattio: Ah my bad: I saw @mnxn pushed the upgrade of `jsoo_promise` and assumed he did it at the same time. Will update.

ulugbekna:
I meant that if the first two elements are `None`, you end up `opam switch
show`ing in both `detect_opam_local_switch` and `detect_opam_sandbox`,
right?

tmattio:
> I meant that if the first two elements are `None`, you end up `opam switch show`ing in both `detect_opam_local_switch` and `detect_opam_sandbox`, right?

Ah yes, I'd prefer keeping them as two separate functions nonetheless: the code clarity is worth the additional call to `opam switch` IMO

rgrinberg (approved):

PR #448, opened 2020-11-24
Add keyboard shortcut for infer interface code action

mnxn:
Adds an `Alt+P` keybinding to fill a file with an inferred interface.

The workflow is quite smooth from my testing:
- `Alt+O` to switch to  new .mli file.
- `CTRL+S`/ `CMD+S` to save the .mli file.
- `Alt+P` to fill the .mli file with an inferred interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix reason highlighting of let extensions (#447)
 - Improve highlighting of Menhir new syntax (#450)
 - Improve Menhir syntax highlighting (#455)
+- Add `Alt + P` keyboard shortcut for infer interface code action (#448)
 
 ## 1.4.0
 
--- a/package.json
+++ b/package.json
@@ -79,6 +79,14 @@
           "kind": "destruct"
         },
         "when": "editorLangId  == ocaml || editorLangId == reason"
+      },
+      {
+        "command": "editor.action.codeAction",
+        "key": "Alt+P",
+        "args": {
+          "kind": "inferred_intf"
+        },
+        "when": "editorLangId  == ocaml.interface || editorLangId == reason"
       }
     ],
     "menus": {


tmattio (approved):

ulugbekna:
Thank for the PR! :-) That's definitely useful but maybe a little more "mnemonic" shortcut such as `alt+i` or `alt+i+i` so that we don't reserve `alt+i` for something more frequent in future?

mnxn:
I tried `Alt+I` but it didn't work on my Mac keyboard since it's used to insert accents. `P` was the next closest key to `O`, which makes it comfortable to use after switching with `Alt+O`.

smelc:
Sounds nice :pray:! Is there a condition on the `mli` file for the action to be available? I tried triggering it, to no avail. I tried on an empty `light.mli` file next to https://gitlab.com/smelc/tezos/-/blob/ec8cf3aeaff465e20c1fbf66d0835cf97b461e6d/src/lib_proxy/light.ml.

tmattio:
@smelc The file has to be saved, or contain some content. It also depends on the latest `ocaml-server-lsp` version (1.3.0), is it the one you're using?

smelc:
I recompiled `ocaml-lsp-server` on tag `1.3.0` (I was indeed using 1.1.*), but it did not change anything. Here's what I'm doing: https://imgur.com/a/5mcsKlx (see the blue popup `No code actions for inferred_intf available` when I press `Alt-p` at the end).

tmattio:
That's strange. Just to double-check - you installed the new version of `ocaml-lsp-server` in your switch `tezos`, right?

ulugbekna:
Have you restarted vscode after recompilation? :-)

smelc:
@tmattio> no it's installed globally, but I think I got this right (this is from my `tezos` directory/switch):

```bash
> eval $(opam env)
> opam exec -- ocamllsp --version
dev
> opam exec -- which ocamllsp
/usr/local/bin/ocamllsp  # outside the sandbox
> ls -alh /usr/local/bin/ocamllsp
lrwxrwxrwx 1 root root 76 nov.  30 17:48 /usr/local/bin/ocamllsp -> /home/churlin/tools/ocaml-lsp/./_build/default/ocaml-lsp-server/src/main.exe
```

This directory is a fresh clone of ocaml-lsp (on changeset 3d2e533). In the past I had `ocaml-lsp-server` pinned in my `tezos` switch (I don't do that anymore, it doesn't work anymore), but I could not find the old version myself (`find . -name ocamllsp` yields nothing from within the switch), but maybe the extension is still finding it somewhere?

@ulugbekna> yes I did :-)

PR #455, opened 2020-11-26
Drastically improve Menhir syntax highlighting

ALANVF:
Menhir's syntax highlighting was bad so I decided to fix it.

Possible future improvements:
1) Fix highlighting for attributes (not possible without `#2` below).
2) Generate grammars using OCaml because JSON is terrible.

A lengthy example:
image

Another example (new merlin syntax):
image

### DIFF
--- a/syntaxes/menhir.json
+++ b/syntaxes/menhir.json
@@ -58,6 +58,7 @@
             { "include": "#type-annotation" },
             { "include": "#token-name" },
             { "include": "source.ocaml#strings" },
+            { "include": "source.ocaml#attributes" },
             { "include": "#comments" }
           ]
         },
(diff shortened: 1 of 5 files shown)


mnxn (approved):

syntaxes/menhir.json
```diff
@@ -66,9 +66,21 @@
...
         {
-          "comment": "builtin declaration",
+          "comment": "parameter declaration",
           "name": "keyword.other.menhir",
-          "match": "%(?:parameter|attribute)\\b"
+          "match": "%parameter\\b"
+        },
+        {
+          "comment": "attribute declaration",
+          "begin": "%(?:attribute\\b)?",
```
ALANVF, line 93: This will break stuff

ALANVF (commented):
Pls fix

syntaxes/menhir.json
mnxn: Please be more descriptive.

Can you give an example?

ALANVF: This might eat up `%%` when matched, and in general the `?` seems to be unnecessary.

mnxn:
Thank you for your contribution!

Issue #457, opened 2020-11-30
Improve discoverability of the extension


ulugbekna:
Problem: There are quite many OCaml extensions, some of which have more downloads than OCaml Platform and are higher in the search (see screenshot below). Users are not sure which extension to install, especially because @hackwaly's extension is installed 10x more and has official OCaml logo. 

![image](https://user-images.githubusercontent.com/16353531/100584126-7d04f380-330d-11eb-89fd-10a90e41eec9.png)

2 things we can do about this:

1)  **Get recommended by vscode:** If a user opens a golang file, they are recommended to install golang extension (see screenshot below). If vscode allows such recommendations for language extensions developed outside microsoft, it would be nice to have this. 

![image](https://user-images.githubusercontent.com/16353531/100583937-3a431b80-330d-11eb-8b3e-e801bb7c1ca2.png)

2) **Try contacting (again?)** other extensions' authors to add a pointer to the official OCaml vscode extension or deprecate as in case for Darrin Morrison's `OCaml and Reason IDE`, which is a great extension but is not maintained afaik.

hackwaly:
>Try contacting (again?) other extensions' authors to add a pointer to the official OCaml vscode extension or deprecate as in case for Darrin Morrison's OCaml and Reason IDE, which is a great extension but is not maintained afaik.

I would like to accept the deprecation PR like [this](https://github.com/hackwaly/vscode-ocaml/pull/115).

tmattio:
Thanks a lot @hackwaly! I'll send out a PR to this 🙂

hackwaly:
Just published new version to make new deprecation to take effect to vscode marketplace.

tmattio:
That's awesome 🎉 

Here's a screenshot of the extension explorer now:

![image](https://user-images.githubusercontent.com/6162008/100624755-0e7f6000-3324-11eb-9270-033a28269c1e.png)

I'll try to get in touch with the maintainers of "OCaml Formatter" and "OCaml and Reason IDE"  as well.

As for OCaml Debugger, we've been discussing about it and it would be interesting to explore an integration of `earlybird` in `vscode-ocaml-platform` Not sure what your plans are for it @hackwaly, but if it sounds like something you'd be onboard with, feel free to reach out (thibaut.mattio@gmail.com) 🙂

hackwaly:
[...]

PR #458, opened 2020-11-30
Change logo and name

tmattio:
This PR addresses part of the concerns raised in #457:

- Change the logo to the official OCaml logo
- Change the name of the extension to OCaml

For the second point, the name currently collides with @hackwaly's extension, but he agreed to deprecate it in favor of this one, so if https://github.com/hackwaly/vscode-ocaml/pull/124 is merged with the name update, that won't be a problem anymore.

Merge blocked by the PR above, and a new release of hackwaly/vscode-ocaml.


mnxn:
Is changing the extension ID from `ocamllabs.ocaml-platform` something that we want to do right now? I think it would be listed as a different extension on the marketplace and make it harder for existing users to update. 

If this repo ever moves to the https://github.com/ocaml organization, maybe we could change the ID to `ocaml.ocaml` then?

tmattio:
> I think it would be listed as a different extension on the marketplace

I don't think that's the case. The PR I mentioned renamed the extension ID from `ocaml` to `ocaml-deprecated` and the old extension is not listed in the market place anymore: https://marketplace.visualstudio.com/search?term=ocaml&target=VSCode&category=All%20categories&sortBy=Relevance

mnxn:
That extension is still listed under `hackwaly.ocaml`:
https://marketplace.visualstudio.com/items?itemName=hackwaly.ocaml

`hackwaly.ocaml-deprecated` is a 404:
https://marketplace.visualstudio.com/items?itemName=hackwaly.ocaml-deprecated

I don't see how it would be possible for the `vsce` tool to detect an extension ID change since it reads from the `package.json` to decide what extension it's publishing in the first place.

If we're planning to move to the `ocaml` organization/publisher (are we?), I'd prefer to wait until then to change the extension ID in order to minimize changes. Just changing the extension title and icon should be sufficient to improve discoverability.

tmattio:
> That extension is still listed under hackwaly.ocaml:
That's very strange: https://github.com/hackwaly/vscode-ocaml/pull/124/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R3

> I don't see how it would be possible for the vsce tool to detect an extension ID 
Maybe it cross-checks with the VCS URI?

> I'd prefer to wait until then to change the extension ID in order to minimize changes
Sure that sounds reasonable

hackwaly:
> That extension is still listed under `hackwaly.ocaml`:
> https://marketplace.visualstudio.com/items?itemName=hackwaly.ocaml
> 
> `hackwaly.ocaml-deprecated` is a 404:
> https://marketplace.visualstudio.com/items?itemName=hackwaly.ocaml-deprecated
> 
> I don't see how it would be possible for the `vsce` tool to detect an extension ID change since it reads from the `package.json` to decide what extension it's publishing in the first place.
> 
> If we're planning to move to the `ocaml` organization/publisher (are we?), I'd prefer to wait until then to change the extension ID in order to minimize changes. Just changing the extension title and icon should be sufficient to improve discoverability.

I  worry about directly change ocaml to ocaml-deprecated in package.json just results two published extensions - one is ocaml the other is ocaml-deprecated. So I didn't use ocaml-deprecated to publish.

AFAIK. There's no name conflict policy in vscode marketplace. So you can use ocamllabs/ocaml without issue.

mnxn:
> I worry about directly change ocaml to ocaml-deprecated in package.json just results two published extensions - one is ocaml the other is ocaml-deprecated. So I didn't use ocaml-deprecated to publish.

That is my guess too. That is also why I don't want to rename `ocamllabs.ocaml-platform` to `ocamllabs.ocaml` right now, because it might confuse users.


> AFAIK. There's no name conflict policy in vscode marketplace. So you can use ocamllabs/ocaml without issue.

Yeah, I don't think there's any name conflict right now. My worry is about the user experience of switching the ID: if it causes two published extension or stops the extension from updating.

tmattio:
I don't think we should wait for the move of the repository to merge this anymore, so I reverted the name change, and this PR now only updates the displayed name and the logo.

Ready to merge when approved.

mnxn:
The extension is still being featured on the extensions page: https://marketplace.visualstudio.com/vscode
Should we wait to change the display name/logo until it stops being featured?

@smorimoto: any thoughts?

smorimoto:
There should be no problem to change the logo, but I don't know the name. If it can be handled as another extension, you should be very careful about that. If something goes wrong, there's nothing to gain and everything to lose.

smorimoto:
Honestly, I don't really think icon and name are blocking adoption right now. In the case of F#, Ionide is the most popular one, but I have never decided not to use it by name, and no one around me has made such a decision. Maybe the best way is to promote it enough in a diffusive medium like OCaml's Twitter account.

tmattio:
Closing for now, as it seems a bit stale. I guess we can revisit when (if) we move the repo to the OCaml org.

src/extension_consts.ml
```diff
@@ -56,8 +56,8 @@ module Commands = struct
 
   let open_ocamllsp_output = ocaml_prefixed "open-ocamllsp-output"
 
-  let open_ocaml_platform_ext_output =
-    ocaml_prefixed "open-ocaml-platform-ext-output"
```
rgrinberg, line 60: isn't this a breaking change?

tmattio: I guess if users have associated a keybinding to it, it is yes. It's a pretty niche use case I feel, but if you feel we shouldn't change the ID I'll revert this

PR #462, opened 2020-12-01
Highlight format indication characters in OCaml strings

mnxn:
Highlights indication characters that begin with `@` (https://ocaml.org/releases/4.11/htmlman/libref/Format.html#fpp).

Example: 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+# Unreleased
+
+- Highlight `@` format indication characters in OCaml strings (#462)
+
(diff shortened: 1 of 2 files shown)


rgrinberg:
Do we highlight this stuff only in formatting strings? It seems wrong to highlight these chars everywhere.

mnxn:
No, it highlights every string. There's no way to determine if it's a format string from just the syntax.

Should this wait until we get proper semantic highlighting?

rgrinberg:
I'd say let's wait.

PR #463, opened 2020-12-01
Automatic installation of missing binaries

### LINKED ISSUE #294: automatic install of ocaml-lsp
See https://github.com/ocaml/ocaml-lsp/issues/227, it would be great if the extension could install ocaml-lsp automatically.
Since the extension already detects the type of sandbox, it would be very simple to install the server using either `opam install` or `esy add`.

### LINKED ISSUE #353: Handling missing ocamlformat
If ocamlformat is missing, the extension displays an error on every file save. What do you think of making this not an error but a box with, say, two buttons: [Install ocamlformat] [Don't show again].

image

Install ocamlformat in case of an opam switch should just `opam install ocamlformat`. Don't know about esy.

### LINKED ISSUE #361: global esy ocaml-lsp-server
## Feature Request

It's a problem to need to add ocaml-lsp-server to every package that you will develop as it's a not ideal solution, that is not a problem when using opam, as you use a global switch and things works, but for esy that's not true as we have no global sandbox, ideally after installing the extension the user would receive a progress bar and in a couple of minutes the environment would be working for any project, maybe that could even be cached on a CI using something like pesy.

If the ocaml-lsp-server is already available on esy or opam then we skip it.

## How

When esy is installed, the extension itself would create a hidden json file with ocaml-lsp-server added

Read the esy lock, get the ocaml version and install it + ocaml-lsp-server on a internal esy package, that could also be used with opam switches and global installations but we need to look on details

## Why not a override?

esy allow us to extend a `package.json` with a json like, that would use a different lock version which would need to reinstall everything with the most recent version, using only the ocaml package prevents that

tmattio:
This PR adds a popup to install missing tools (`ocamllsp` and `ocamlformat` for now) when they are not found in the user's sandbox.

It installs the tools from source for now, which is time-consuming, so we'll want to use something like `opam-tools`, but I figured it was good enough for a first version, and we most likely want to use it as a fallback anyways.

Not ready for review yet, but I wanted to get some early feedback on the approach.

**TODO**

- [ ] ~~Cancellation~~
- [X] Display errors
- [ ] ~~What to do when the install command wants to upgrade/downgrade some dependencies? Should we use "-y"?~~
- [X] Do not try to run ocamllsp if missing (and avoid the won't restart error)
- [x] Use the toolchain corresponding to the sandboxe's ocaml compiler version
- [ ] What to do if the user does not have Opam? Can we install it for them (🙈)?
- [ ] What to do if the user does not have Opam but has Esy? Should we create a private sandbox with it?
- [x] Should we use the tools in the user's sandbox when they are available?

Fixes #294
Fixes #361
Fixes #353

### DIFF
--- a/src/bindings/node/node.ml
+++ b/src/bindings/node/node.ml
@@ -40,6 +40,10 @@ module Process = struct
   end
 end
 
+module Os = struct
+  val homedir : unit -> string [@@js.global "os.homedir"]
+end
+
 module JsError = struct
   let message (error : Promise.error) =
     let js_error = [%js.of: Promise.error] error in
--- a/src/sandbox.ml
+++ b/src/sandbox.ml
@@ -230,6 +230,15 @@ let of_settings () : t option Promise.t =
   | Some Global -> Promise.return (Some Global)
   | Some (Custom template) -> Promise.return (Some (Custom template))
 
+let workspace_root () =
+  match Workspace.workspaceFolders () with
+  | [] -> None
+  | [ workspace_folder ] ->
+    Some (workspace_folder |> WorkspaceFolder.uri |> Uri.path |> Path.of_string)
+  | _ ->
+    (* We don't support multiple workspace roots at the moment *)
+    None
+
 let detect_esy_sandbox ~project_root esy () =
   let open Promise.Option.Syntax in
   let* esy = esy in
@@ -265,22 +274,15 @@ let detect_opam_sandbox ~project_root opam () =
   Opam (opam, switch)
 
 let detect () =
-  match Workspace.workspaceFolders () with
-  | [] -> Promise.return None
-  | [ workspace_folder ] ->
-    let project_root =
-      workspace_folder |> WorkspaceFolder.uri |> Uri.path |> Path.of_string
-    in
-    let available = available_sandboxes () in
-    Promise.List.find_map
-      (fun f -> f ())
-      [ detect_opam_local_switch ~project_root available.opam
-      ; detect_esy_sandbox ~project_root available.esy
-      ; detect_opam_sandbox ~project_root available.opam
-      ]
-  | _ ->
-    (* If there are several workspace folders, skip the detection entirely. *)
-    Promise.return None
+  let open Promise.Option.Syntax in
+  let* project_root = workspace_root () |> Promise.return in
+  let available = available_sandboxes () in
+  Promise.List.find_map
+    (fun f -> f ())
+    [ detect_opam_local_switch ~project_root available.opam
+    ; detect_esy_sandbox ~project_root available.esy
+    ; detect_opam_sandbox ~project_root available.opam
+    ]
 
 let of_settings_or_detect () =
   let open Promise.Syntax in
@@ -392,14 +394,6 @@ let sandbox_candidates ~workspace_folders =
   let+ esy, opam = Promise.all2 (esy, opam) in
   (global :: custom :: esy) @ opam
 
-let setup_sandbox (kind : t) =
-  match kind with
-  | Esy (esy, manifest) -> Esy.setup_sandbox esy manifest
-  | Opam _
-  | Global
-  | Custom _ ->
-    Promise.Result.return ()
-
 let select_sandbox () =
   let open Promise.Syntax in
   let workspace_folders = Workspace.workspaceFolders () in
@@ -459,27 +453,41 @@ let get_command sandbox bin args : Cmd.t =
     in
     Shell command
 
+let has_command sandbox bin =
+  let open Promise.Syntax in
+  let cmd =
+    match Platform.t with
+    | Win32 -> get_command sandbox "cmd" [ "/c"; "where"; bin ]
+    | _ -> get_command sandbox "which" [ bin ]
+  in
+  let+ result = Cmd.output cmd in
+  match result with
+  | Ok _ -> true
+  | Error _ -> false
+
 let get_lsp_command ?(args = []) sandbox : Cmd.t =
   get_command sandbox "ocamllsp" args
 
 let get_dune_command sandbox args : Cmd.t = get_command sandbox "dune" args
 
-let run_setup sandbox =
-  let open Promise.Syntax in
-  let+ output =
-    let open Promise.Result.Syntax in
-    let* () = setup_sandbox sandbox in
-    let args = [ "--version" ] in
-    let* command = Cmd.check (get_lsp_command sandbox ~args) in
-    Cmd.output command
-  in
-  match output with
-  | Ok _ -> Ok ()
-  | Error msg ->
-    (* TODO: if ocamllsp not found, suggest to install it on press of a button;
-       consider checking and suggesting installation for other tools: formatter,
-       etc. *)
-    Error (Printf.sprintf "Sandbox initialisation failed: %s" msg)
+let get_install_command sandbox tools =
+  match sandbox with
+  | Opam (opam, switch) -> Some (Opam.install opam switch ~packages:tools)
+  | Esy (esy, manifest) -> Some (Esy.install esy manifest ~packages:tools)
+  | _ -> None
+
+let get_exec_command sandbox tools =
+  match sandbox with
+  | Opam (opam, switch) -> Some (Opam.exec opam switch ~args:tools)
+  | Esy (esy, manifest) -> Some (Esy.exec esy manifest ~args:tools)
+  | _ -> None
+
+let ocaml_version sandbox =
+  let cmd = get_command sandbox "ocamlc" [ "--version" ] in
+  let open Promise.Result.Syntax in
+  let* cmd = Cmd.check cmd in
+  let+ output = Cmd.output cmd in
+  String.strip output
 
 let packages t =
   let open Promise.Result.Syntax in
@@ -573,8 +581,8 @@ let install_packages t packages =
       let open Promise.Syntax in
       let+ result =
         let open Promise.Result.Syntax in
-        let* _ = Opam.update opam |> Cmd.output in
-        let+ _ = Opam.install opam switch packages |> Cmd.output in
+        let* _ = Opam.update opam switch |> Cmd.output in
+        let+ _ = Opam.install opam switch ~packages |> Cmd.output in
         ()
       in
       match result with
@@ -611,7 +619,7 @@ let upgrade_packages t =
       let open Promise.Syntax in
       let+ result =
         let open Promise.Result.Syntax in
-        let* _ = Opam.update opam |> Cmd.output in
+        let* _ = Opam.update opam switch |> Cmd.output in
         let+ _ = Opam.upgrade opam switch |> Cmd.output in
         ()
       in
--- a/src/toolchain.ml
+++ b/src/toolchain.ml
@@ -1,0 +1,244 @@
+open Import
+
+let get_opamroot () =
+  let open Promise.Syntax in
+  let+ home =
+    if not (Ocaml_windows.use_ocaml_env ()) then
+      Promise.return @@ Node.Os.homedir ()
+    else
+      let+ output = Ocaml_windows.cygwin_home () in
+      match output with
+      | Ok cygwin_home -> cygwin_home
+      | Error e ->
+        show_message `Warn
+          "Could not determine Cygwin home directory, defaulting to Windows \
+           home directory: %s"
+          e;
+        Node.Os.homedir ()
+  in
+  Path.(of_string home / ".vscode-ocaml" / "opam")
+
+type t =
+  { project_sandbox : Sandbox.t option
+  ; opam : Opam.t * Opam.Switch.t
+  ; version : string
+  ; is_new : bool
+  }
+
+module Tool = struct
+  type t =
+    { opam_package : string
+    ; cmd : Cmd.t
+    }
+
+  let opam_package t = t.opam_package
+
+  let cmd t = t.cmd
+
+  let make ?(args = []) ~opam_package
+      { opam = opam, switch; project_sandbox; _ } bin =
+    match project_sandbox with
+    | Some sandbox ->
+      let open Promise.Syntax in
+      let+ has_command = Sandbox.has_command sandbox bin in
+      if has_command then
+        { opam_package; cmd = Sandbox.get_command sandbox bin args }
+      else
+        { opam_package; cmd = Opam.exec opam switch ~args:(bin :: args) }
+    | _ ->
+      Promise.return
+        { opam_package; cmd = Opam.exec opam switch ~args:(bin :: args) }
+
+  let lsp ?args toolchain =
+    make ~opam_package:"ocaml-lsp-server" toolchain "ocamllsp" ?args
+
+  let dune ?args toolchain = make ~opam_package:"dune" toolchain "dune" ?args
+
+  let ocamlformat ?args toolchain =
+    make ~opam_package:"ocamlformat" toolchain "ocamlformat" ?args
+
+  let all toolchain =
+    Promise.all_list
+      [ lsp toolchain ~args:[ "--version" ]
+      ; ocamlformat toolchain ~args:[ "--version" ]
+      ]
+
+  let detect_missing toolchain =
+    let open Promise.Syntax in
+    let missing_command_opt t =
+      let cmd = t.cmd in
+      let+ output =
+        let open Promise.Result.Syntax in
+        let* command = Cmd.check cmd in
+        Cmd.output command
+      in
+      match output with
+      | Ok _ -> None
+      | Error _ -> Some t
+    in
+    let* all = all toolchain in
+    Promise.List.filter_map missing_command_opt all
+end
+
+let compiler_of_version v =
+  match (Platform.t, Platform.arch) with
+  | Win32, X32 -> "ocaml-variants." ^ v ^ "+mingw32c"
+  | Win32, _ -> "ocaml-variants." ^ v ^ "+mingw64c"
+  | _ -> "ocaml-base-compiler." ^ v
+
+let switch_of_version v = "vscode-ocaml-toolchain." ^ v
+
+let compiler_version project_sandbox =
+  let default = "4.11.1" in
+  match project_sandbox with
+  | None -> Promise.return default
+  | Some sandbox -> (
+    let open Promise.Syntax in
+    let+ result = Sandbox.ocaml_version sandbox in
+    match result with
+    | Error _ -> default
+    | Ok version -> version )
+
+let is_sandbox_installed opam switch =
+  let open Promise.Syntax in
+  let cmd = Opam.exec opam switch ~args:[ "true" ] in
+  let+ result = Cmd.output cmd in
+  match result with
+  | Ok _ -> true
+  | Error _ -> false
+
+let setup_toolchain_sandbox project_sandbox opamroot =
+  let open Promise.Syntax in
+  let* opam_opt = Opam.make ~root:opamroot () in
+  match opam_opt with
+  | None -> Promise.return (Error "Opam is not available")
+  | Some opam ->
+    let* compiler_version = compiler_version project_sandbox in
+    let compiler_name = compiler_of_version compiler_version in
+    let sandbox_name = switch_of_version compiler_version in
+    let open Promise.Result.Syntax in
+    let+ _ =
+      Opam.switch_create opam ~name:sandbox_name ~args:[ compiler_name ]
+      |> Cmd.output
+    in
+    ()
+
+let get_install_command { opam = opam, switch; _ } tools =
+  let opam_packages = List.map tools ~f:Tool.opam_package in
+  Opam.install opam switch ~packages:opam_packages
+
+let with_progress ~msg f =
+  let open Promise.Syntax in
+  let progress_options =
+    ProgressOptions.create ~location:(`ProgressLocation Window) ~title:msg
+      ~cancellable:false ()
+  in
+  let task ~progress:_ ~token:_ =
+    let+ result = f () in
+    match result with
+    | Ok _ -> Ok ()
+    | Error err ->
+      show_message `Error "%s" err;
+      Error err
+  in
+  Window.withProgress
+    Interop.Js.((module Result (Unit) (String)))
+    ~options:progress_options ~task
+
+let opam_init opam opamroot =
+  let open Promise.Syntax in
+  let* exists = Fs.exists (opamroot |> Path.to_string) in
+  if exists then
+    Promise.Result.return ()
+  else
+    with_progress
+      ~msg:"Initializing Platform Tools (this might take a few minutes)"
+      (fun () ->
+        let+ result = Opam.init opam |> Cmd.output in
+        Result.map_error result ~f:(fun err ->
+            Printf.sprintf
+              "The creation of the toolchain Opam switch failed: %s" err))
+
+let install_missing_tools t tools opamroot =
+  let open Promise.Syntax in
+  match tools with
+  | [] -> Promise.Result.return ()
+  | tools ->
+    with_progress
+      ~msg:"Installing Platform Tools (this might take a few minutes)"
+      (fun () ->
+        let opam, switch = t.opam in
+        let* is_installed = is_sandbox_installed opam switch in
+        let* _ =
+          match is_installed with
+          | false -> setup_toolchain_sandbox t.project_sandbox opamroot
+          | true -> Promise.Result.return ()
+        in
+        let cmd = get_install_command t tools in
+        let+ result = Cmd.output cmd in
+        Result.map_error result ~f:(fun err ->
+            Printf.sprintf "The installation of the Platform tools failed: %s"
+              err))
+
+let upgrade t =
+  if t.is_new then (
+    log "Is new, won't upgrade";
+    Promise.Result.return ()
+  ) else
+    let opam, switch = t.opam in
+    with_progress ~msg:"Checking for Platform Tools updates" (fun () ->
+        let open Promise.Syntax in
+        let* _ = Opam.update opam switch |> Cmd.output in
+        let+ result = Opam.upgrade opam switch |> Cmd.output in
+        Result.map_error result ~f:(fun err ->
+            Printf.sprintf "The upgrade of the Platform tools failed: %s" err))
+
+let setup ?project_sandbox () =
+  let open Promise.Syntax in
+  let* compiler_version = compiler_version project_sandbox in
+  let sandbox_name = switch_of_version compiler_version in
+  let* opamroot = get_opamroot () in
+  let* opam_opt =
+    let open Promise.Option.Syntax in
+    let* opam = Opam.make ~root:opamroot () in
+    let+ switch = Opam.Switch.of_string sandbox_name |> Promise.return in
+    (opam, switch)
+  in
+  match opam_opt with
+  | None ->
+    let msg =
+      "Opam is required to install the toolchain. Install it manually to use \
+       the extension."
+    in
+    show_message `Error "%s" msg;
+    Promise.return (Error msg)
+  | Some ((opam, switch) as opam_switch) ->
+    let open Promise.Result.Syntax in
+    let* () = opam_init opam opamroot in
+    let open Promise.Syntax in
+    let* is_installed = is_sandbox_installed opam switch in
+    let t =
+      { opam = opam_switch
+      ; project_sandbox
+      ; version = compiler_version
+      ; is_new = not is_installed
+      }
+    in
+    let* missing_tools =
+      if is_installed then
+        Tool.detect_missing t
+      else
+        Tool.all t
+    in
+    let open Promise.Result.Syntax in
+    let+ () = install_missing_tools t missing_tools opamroot in
+    t
+
+let to_pretty_string { version; _ } = "OCaml " ^ version
+
+let get_lsp_command ?args t = Tool.lsp ?args t |> Promise.map Tool.cmd
+
+let get_dune_command ?args t = Tool.dune ?args t |> Promise.map Tool.cmd
+
+let get_ocamlformat_command ?args t =
+  Tool.ocamlformat ?args t |> Promise.map Tool.cmd
--- a/src/toolchain.mli
+++ b/src/toolchain.mli
@@ -1,0 +1,26 @@
+(** Manages OCaml Platform toolchain
+
+    The toolchain is installed in a private sandbox, managed by VSCode only,
+    where the tools we depend on are installed automatically.
+
+    The sandbox uses Opam, but it does not conflict with users who use Esy as
+    their package managers, since the tools are separated from the project. *)
+
+type t
+
+val setup : ?project_sandbox:Sandbox.t -> unit -> (t, string) result Promise.t
+
+val upgrade : t -> (unit, string) result Promise.t
+
+(* Helper utils *)
+
+(** Extract lsp command and arguments *)
+val get_lsp_command : ?args:string list -> t -> Cmd.t Promise.t
+
+(** Extract a dune command *)
+val get_dune_command : ?args:string list -> t -> Cmd.t Promise.t
+
+(** Extract a ocamlformat command and arguments *)
+val get_ocamlformat_command : ?args:string list -> t -> Cmd.t Promise.t
+
+val to_pretty_string : t -> string
--- a/vscode/interop/interop.ml
+++ b/vscode/interop/interop.ml
@@ -101,6 +101,14 @@ module Js = struct
     type t = Ojs.t [@@js]
   end
 
+  module Unit = struct
+    type t = unit
+
+    let t_of_js _ = ()
+
+    let t_to_js _ = undefined
+  end
+
   module Bool = struct
     type t = bool [@@js]
   end
@@ -115,6 +123,31 @@ module Js = struct
 
   module Option (T : T) = struct
     type t = T.t option [@@js]
+  end
+
+  module Result (Ok : T) (Error : T) = struct
+    type t = (Ok.t, Error.t) result
+
+    type js_result =
+      { case : string
+      ; ok : Ojs.t
+      ; error : Ojs.t
+      }
+    [@@js]
+
+    let t_of_js js_val =
+      match js_result_of_js js_val with
+      | { case = "ok"; ok; _ } -> Ok ([%js.to: Ok.t] ok)
+      | { case = "error"; error; _ } -> Error ([%js.to: Error.t] error)
+      | _ -> assert false
+
+    let t_to_js = function
+      | Ok ok ->
+        let ok = [%js.of: Ok.t] ok in
+        js_result_to_js { case = "ok"; ok; error = undefined }
+      | Error error ->
+        let error = [%js.of: Error.t] error in
+        js_result_to_js { case = "error"; error; ok = undefined }
   end
 
   module Or_undefined (T : T) = struct
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,11 @@
   is located in the same folder as the open file or creates the dune file in a
   draft mode (not yet saved on the disk). (#499)
 
+- Install OCaml Platform tools automatically (#463)
+
+  The platform tools (a.k.a toolchain) are installed in a private opam switch
+  that should only be used by `vscode-ocaml-platform`.
+
 ## 1.5.1
 
 - Improve highlighting of type parameters and `module type of` (#461)
--- a/README.md
+++ b/README.md
@@ -8,15 +8,13 @@ _Please report any bugs you encounter._
 
 ## Quick start
 
-1. Install this extension from
+1. [Install Opam](https://opam.ocaml.org/doc/Install.html)
+2. Install this extension from
    [the VSCode Marketplace](https://marketplace.visualstudio.com/items?itemName=ocamllabs.ocaml-platform)
    (or by entering `ext install ocamllabs.ocaml-platform` at the command palette
    <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>
    (<kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> on MacOS)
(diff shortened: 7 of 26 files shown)


src/toolchain.ml
```diff
@@ -416,23 +418,120 @@ let get_command toolchain bin args : Cmd.t =
...
-       consider checking and suggesting installation for other tools: formatter, etc. *)
-    Error (Printf.sprintf "Toolchain initialisation failed: %s" msg)
+  | Ok _ -> None
+  | Error _ -> Some binary
+
+let run_setup toolchain =
+  let open Promise.Syntax in
+  let* missing_tools =
+    [ ("ocaml-lsp-server", get_lsp_command toolchain ~args:[ "--version" ])
+    ; ("ocamlformat", get_ocamlformat_command toolchain ~args:[ "--version" ])
```
ulugbekna, line 517: Just a heads up regarding formatters and formatting:

You may have noticed that formatting is done by ocaml-lsp, not the extension. Moreover, `ocamllsp` invokes either `ocamlformat` or `refmt` based on the file that needs to be formatted. So it may make sense to handle logic for detecting whether a formatter is installed on the ocaml-lsp side. The lang server could send a notification if it can't find the required tool.

If you choose to do it on the extension side, keep in mind that a user may only need refmt installed or ocamlformat or both, so it's not straightforward.

ulugbekna: Sorry my TODO comment was misleading

tmattio: Makes sense - do you know if there's a specific notification for that purpose? I've tried to see if LSP had requests/notifications for this kind of things, but did not find anything.

tmattio:
So, there's a problem when trying to install dependencies in a switch that uses `ocaml-system`, as opposed to `ocaml-base-compiler`. That's because the commands won't inherit the global environment and thus `ocamlc` won't be in the path.

One solution is to install `ocaml-base-compiler` on behalf of the user when this happens, but then the tools installation takes **a lot** of time.

Another solution would be to use the global env when running opam commands, but I don't know if that's possible, and if it is, I'm not sure this is a good idea.

Any insights on this?

ulugbekna:
There isn't a specific notification for that purpose afaik, but you can
define custom notifications. Currently, custom notifications in ocaml-lsp
are called `Unknown_notification`s, iirc, and are `assert false`d instead
of being handled. I would recommend adding support for those, then define a
custom notif for missing tools.

The above taken into account, I would concentrate this PR on installing
missing ocaml-lsp and create another PR for other tools.
----

Sidenote: I haven't looked into the whole code and don't know what the
UX/UI look like, but another heads up: we need to handle installation in
both opam and esy.

It may make sense to start shaping up a common interface for `Opam` and
`Esy` modules with a fn `install`, for example.

tmattio:
> but you can define custom notifications

@rgrinberg I know you were wary of adding custom requests, how do you feel about this? 

> I would concentrate this PR on installing missing ocaml-lsp and create another PR for other tools

Good point. I'll try to handle it in this PR and if it ends up being cumbersome I'll split it into another one.

> It may make sense to start shaping up a common interface for `Opam` and `Esy` modules with a fn `install`, for example.

That's what this PR implements 🙂

tmattio:
@rgrinberg As discussed, I change the way we install the tools to install them in a private switch. It's working but needs a bit more work, in particular, to install in a private switch with the same OCaml version the users are using in their sandbox.

There are now too modules:

- `Sandbox`, for the users' project sandbox
- `Toolchain`, for the private switch containing OCaml Platform tools

The installation is **really** slow the first time, as expected, so we'll want to optimize that with pre-built binaries/`opam-tools` when possible.

rgrinberg:
> So, there's a problem when trying to install dependencies in a switch that uses ocaml-system, as opposed to ocaml-base-compiler. That's because the commands won't inherit the global environment and thus ocamlc won't be in the path.
Using the system compiler is out of question.

Using the system compiler is out of question anyway.

> I know you were wary of adding custom requests, how do you feel about this?

Very negatively. 

Would it be possible to move unrelated commits to separate PR's? It's quite hard to review the work otherwise.

> The installation is really slow the first time, as expected, so we'll want to optimize that with pre-built binaries/opam-tools when possible.

The other reason why I prefer to use opam-tools is that I'd like to add some second class support for non vscode users. It won't be done automatically, but at least they will need to run a binary that works around all the limitations.

mnxn:
> There are now two modules:
> 
>  - `Sandbox`, for the users' project sandbox
> 
> - `Toolchain`, for the private switch containing OCaml Platform tools

This terminology should be switched. "Sandbox" implies isolation of some kind, and `Global`/`Custom` are not isolated at all; they just contain a set of tools, a toolchain.

A "private switch containing OCaml Platform tools" fits the "Sandbox" label better since it's a toolchain isolated from the user's project.

> - What to do if the user does not have Opam? Can we install it for them (🙈)?
> - What to do if the user does not have Opam but has Esy? Should we create a private sandbox with it?

A user that has esy installed should not have to install opam to install the tools. This would be especially annoying on Windows, where the setup process for opam-repository-mingw is not very simple or fast.

I would expect that the extension would try to install ocamllsp in my current toolchain, using `opam install` or `esy add`.

src/toolchain.mli
```diff
@@ -1,85 +1,45 @@
...
-  | Opam of Opam.t * Opam.Switch.t
-  | Esy of Esy.t * Path.t
-  | Global
-  | Custom of string
+val get_installed_sandbox : unit -> Sandbox.t option Promise.t
 
-val equal : t -> t -> bool
+val setup_toolchain_sandbox :
+     unit
+  -> (unit, [> `Opam_not_available | `Switch_create_failed of string ]) result
```
rgrinberg, line 17: Are we doing something specific for these errors? If not, then there's no need to introduce specific error types.

The old codebase suffered from this a problem: a massive amount of error types that all eventually ended up flushed down as strings. Let's not reintroduce it.

.ocamlformat
```diff
@@ -3,6 +3,8 @@ break-separators=before
 dock-collection-brackets=false
 break-sequences=true
 doc-comments=before
+wrap-comments = true
+parse-docstrings = true
```
rgrinberg, line 7: Should come in a separate PR.

src/bindings/node/node.ml
```diff
@@ -86,9 +86,15 @@ module ChildProcess = struct
...
+  val killed : t -> bool [@@js.get "killed"]
+
   val on : t -> string -> (int -> unit) -> unit [@@js.call]
 
-  let on_close t close = on t "close" close
+  let on_close t f = on t "close" f
+
+  let on_error t f = on t "error" f
+
+  val kill : t -> string -> bool [@@js.call]
```
rgrinberg, line 97: Large binding changes should also be separate.

rgrinberg:
> This terminology should be switched. "Sandbox" implies isolation of some kind, and Global/Custom are not isolated at all; they just contain a set of tools, a toolchain.

Global/Custom may indeed be isolated, but we just don't know their structure. For example they could come from a nix environment. I agree that this is confusing though.

> A user that has esy installed should not have to install opam to install the tools. This would be especially annoying on Windows, where the setup process for opam-repository-mingw is not very simple or fast.

Yes, that makes sense. Relying on opam is a probably a band aid until we get binary installs. However, I do not think that we should complicate the process by adding a separate pipeline for esy. The goal of this feature is to be transparent and reliable. We aren't going to get there if we need to deal with an additional package manager.

.ocamlformat
rgrinberg: Actually, we should just copy the config in the dune repo both here and in lsp. It will give us a consistent style.

tmattio:
> I would expect that the extension would try to install ocamllsp in my current toolchain, using opam install or esy add.

That's how I implemented this first, but we reviewed the design with @rgrinberg and think it makes more sense to isolate the tools from the user's current sandbox.

One reason is to optimize the installation: if we install the tools in the sandbox every time, we waste the users' time for every new project they open

Another reason is that the tools have dependencies. These dependencies may conflict with the users' project, we shouldn't ask users to change their project dependencies because of the tooling.

Lastly, it'll be much simpler to manage and update the platform tools if we control the environment.

tmattio:
> This would be especially annoying on Windows, where the setup process for opam-repository-mingw is not very simple or fast.

I agree, it'd prefer not to have system dependencies as well and installing opam on Windows seems... We don't want to go there before 2.2...

At the same time I agree with @rgrinberg that this is a temporary solution until we get binaries, at which point it'll become a fallback solution. For Esy users who don't have opam, I think we can try to see if the tools are in the user sandbox if we cannot use the private toolchain (or the other way around).

tmattio:
> Would it be possible to move unrelated commits to separate PR's? It's quite hard to review the work otherwise.

I extracted the renaming of the `Toolchain` module in #466, but I rebased on top of it, since it depends on this change. Let's merge this PR first maybe to clean up the diff?

src/bindings/node/node.ml
tmattio: I reverted this, it's not needed anymore

tmattio:
@rgrinberg @mnxn It works well for me, but as it is a critical part of the extension (if this does not work, the extension is unusable), would you mind testing on your side to make sure you're happy with everything?

mnxn:
Broken on windows:

image

<details>
<summary>Output channel log</summary>

```json
external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "switch",
    "list",
    "-s"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "4.11.1+mingw64c\n",
    "stderr": ""
  }
}

1 switches
external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "ocamlc",
    "--version"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "4.11.1\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=vscode-ocaml-toolchain.4.11.1",
    "--",
    "true"
  ],
  "result": {
    "exitCode": 2,
    "stdout": "",
    "stderr": "[ERROR] The selected switch vscode-ocaml-toolchain.4.11.1 is not installed.\n"
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=vscode-ocaml-toolchain.4.11.1",
    "--",
    "true"
  ],
  "result": {
    "exitCode": 2,
    "stdout": "",
    "stderr": "[ERROR] The selected switch vscode-ocaml-toolchain.4.11.1 is not installed.\n"
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "which",
    "ocamlc"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "/home/Max/.opam/4.11.1+mingw64c/bin/ocamlc\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "which",
    "ocamlformat"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "/home/Max/.opam/4.11.1+mingw64c/bin/ocamlformat\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "which",
    "ocamllsp"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "/home/Max/.opam/4.11.1+mingw64c/bin/ocamllsp\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "ocamlc",
    "--version"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "4.11.1\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "switch",
    "create",
    "vscode-ocaml-toolchain.4.11.1",
    "ocaml-base-compiler.4.11.1"
  ],
  "result": {
    "exitCode": 5,
    "stdout": "",
    "stderr": "[ERROR] No compiler matching 'ocaml-base-compiler.4.11.1' found, use 'opam switch list-available' to see what is available, or use '--packages' to select packages explicitly.\n"
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\opam.exe",
  "args": [
    "install",
    "--switch=vscode-ocaml-toolchain.4.11.1",
    "-y",
    "ocaml-base-compiler",
    "ocaml-lsp-server",
    "ocamlformat"
  ],
  "result": {
    "exitCode": 2,
    "stdout": "",
    "stderr": "[ERROR] The selected switch vscode-ocaml-toolchain.4.11.1 is not installed.\n"
  }
}


```

</details>

The configured opam switch that I'm using does have all of the tools available. A few things I noticed:

- `opam exec --switch=vscode-ocaml-toolchain.4.11.1 -- true` is being run twice

- `which` is returning unix paths for Cygwin, which won't work because VSCode is running from normal Windows.

tmattio:
Thanks a lot for testing @mnxn! I'll fix these and will make sure it runs properly on Windows 🙂

rgrinberg:
Will try to get around to this this week. Might be delayed until next week though.

tmattio:
@mnxn 

> which is returning unix paths for Cygwin, which won't work because VSCode is running from normal Windows.

We're only considering the exit code of `which`, so this should still be ok, no?

src/toolchain.ml
```diff
@@ -0,0 +1,161 @@
...
+           let* cmd = cmd in
+           missing_command_opt binary cmd)
+    |> Promise.all_list
+    |> Promise.map List.filter_opt
+end
+
+let compiler_of_version v =
+  match (Platform.t, Platform.arch) with
+  | Win32, X32 -> "ocaml-base-compiler." ^ v ^ "+mingw32c"
+  | Win32, _ -> "ocaml-base-compiler." ^ v ^ "+mingw64c"
```
tmattio, line 59: @mnxn Do you know if there's going to be any issue using a 64 bits version of the compiler in a 32 bits Cygwin?

mnxn:
> We're only considering the exit code of which, so this should still be ok, no?

Oh, if we're just using the exit codes that should be fine.

src/toolchain.ml
rgrinberg: Before we worry about this, let's make sure there are even 32 bit users out there. I highly doubt there are any left.

mnxn: Based on https://fdopen.github.io/opam-repository-mingw/installation/, it says "you can use [either archive] with either the 32-bit or 64-bit version of cygwin". The ocaml compiler isn't linked against cygwin1.dll either, so the architecture of cygwin shouldn't matter.

What does matter is the architecture of the system, which you handle how I expect here.

The only thing that might need changing is the prefix. On the same [installation page](https://fdopen.github.io/opam-repository-mingw/installation/), the documentation refers to `ocaml-variants` instead of `ocaml-base-compiler`.

tmattio:
@mnxn This should work properly on Windows now, feel free to try again 🙂

mnxn:
Installation of the tools seems to work fine, but it installs them even if they're already available in my current switch.

<details>
<summary>Output channel log</summary>

```json
external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "switch",
    "list",
    "-s"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "4.11.1+mingw64c\n",
    "stderr": ""
  }
}

1 switches
external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "ocamlc",
    "--version"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "4.11.1\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=vscode-ocaml-toolchain.4.11.1",
    "--",
    "true"
  ],
  "result": {
    "exitCode": 2,
    "stdout": "",
    "stderr": "[ERROR] The selected switch vscode-ocaml-toolchain.4.11.1 is not installed.\n"
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=vscode-ocaml-toolchain.4.11.1",
    "--",
    "true"
  ],
  "result": {
    "exitCode": 2,
    "stdout": "",
    "stderr": "[ERROR] The selected switch vscode-ocaml-toolchain.4.11.1 is not installed.\n"
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "which",
    "ocamlformat"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "/home/Max/.opam/4.11.1+mingw64c/bin/ocamlformat\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "which",
    "ocamllsp"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "/home/Max/.opam/4.11.1+mingw64c/bin/ocamllsp\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "ocamlc",
    "--version"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "4.11.1\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "switch",
    "create",
    "vscode-ocaml-toolchain.4.11.1",
    "ocaml-variants.4.11.1+mingw64c"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "\n<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>\n[ocaml-variants.4.11.1+mingw64c] found in cache\n\n<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>\n-> installed base-bigarray.base\n-> installed base-threads.base\n-> installed base-unix.base\n-> installed ocaml-variants.4.11.1+mingw64c\n-> installed ocaml-config.1\n-> installed ocaml.4.11.1\nDone.\n# Run eval $(opam env) to update the current shell environment\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "install",
    "--switch=vscode-ocaml-toolchain.4.11.1",
    "-y",
    "ocaml-lsp-server",
    "ocamlformat"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "The following actions will be performed:\n  - install seq                     base     [required by tyxml]\n  - install conf-m4                 1        [required by ocamlfind]\n  - install ocamlbuild              0.14.0   [required by fpath, uuseg]\n  - install cmdliner                1.0.4    [required by ocamlformat]\n  - install dune                    2.7.1    [required by ocaml-lsp-server, ocamlformat]\n  - install ocamlfind               1.8.1    [required by ocaml-lsp-server]\n  - install uchar                   0.0.2    [required by uutf]\n  - install stdlib-shims            0.1.0    [required by ocaml-lsp-server]\n  - install sexplib0                v0.14.0  [required by base, ppxlib]\n  - install result                  1.5      [required by ocaml-lsp-server]\n  - install re                      1.9.0    [required by ocamlformat]\n  - install ppx_derivers            1.2.1    [required by ppxlib]\n  - install ocaml-version           3.0.0    [required by ocamlformat]\n  - install ocaml-migrate-parsetree 2.1.0    [required by ppxlib]\n  - install ocaml-compiler-libs     v0.12.3  [required by ppxlib]\n  - install menhirSdk               20201122 [required by menhir]\n  - install menhirLib               20201122 [required by menhir]\n  - install fix                     20201120 [required by ocamlformat]\n  - install easy-format             1.3.2    [required by yojson]\n  - install dune-build-info         2.7.1    [required by ocaml-lsp-server, ocamlformat]\n  - install cppo                    1.6.6    [required by odoc]\n  - install topkg                   1.0.3    [required by fpath, uuseg]\n  - install csexp                   1.3.2    [required by ocaml-lsp-server]\n  - install ppxlib                  0.20.0   [required by ocamlformat]\n  - install menhir                  20201122 [required by ocamlformat]\n  - install biniou                  1.2.1    [required by yojson]\n  - install uutf                    1.0.2    [required by ocamlformat]\n  - install astring                 0.8.5    [required by fpath, odoc]\n  - install dune-configurator       2.7.1    [required by base]\n  - install yojson                  1.7.0    [required by ocaml-lsp-server]\n  - install uucp                    13.0.0   [required by uuseg]\n  - install tyxml                   4.4.0    [required by odoc]\n  - install fpath                   0.7.3    [required by ocamlformat]\n  - install base                    v0.14.0  [required by ocamlformat]\n  - install ppx_yojson_conv_lib     v0.14.0  [required by ocaml-lsp-server]\n  - install dot-merlin-reader       3.4.2    [required by ocaml-lsp-server]\n  - install uuseg                   13.0.0   [required by ocamlformat]\n  - install odoc                    1.5.2    [required by ocamlformat]\n  - install stdio                   v0.14.0  [required by ocamlformat]\n  - install ocaml-lsp-server        1.3.0\n  - install ocamlformat             0.16.0\n===== 41 to install =====\n\n<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>\n[astring.0.8.5] found in cache\n[base.v0.14.0] found in cache\n[biniou.1.2.1] found in cache\n[cmdliner.1.0.4] found in cache\n[cppo.1.6.6] found in cache\n[csexp.1.3.2] found in cache\n[dot-merlin-reader.3.4.2] found in cache\n[dune.2.7.1] found in cache\n[dune-build-info.2.7.1] found in cache\n[dune-configurator.2.7.1] found in cache\n[easy-format.1.3.2] found in cache\n[fix.20201120] found in cache\n[fpath.0.7.3] found in cache\n[menhir.20201122] found in cache\n[menhirLib.20201122] found in cache\n[menhirSdk.20201122] found in cache\n[ocaml-compiler-libs.v0.12.3] found in cache\n[ocaml-lsp-server.1.3.0] found in cache\n[ocaml-migrate-parsetree.2.1.0] downloaded from https://github.com/ocaml-ppx/ocaml-migrate-parsetree/releases/download/v2.1.0/ocaml-migrate-parsetree-v2.1.0.tbz\n[ocaml-version.3.0.0] downloaded from https://github.com/ocurrent/ocaml-version/releases/download/v3.0.0/ocaml-version-v3.0.0.tbz\n[ocamlbuild.0.14.0] found in cache\n[ocamlfind.1.8.1] found in cache\n[odoc.1.5.2] found in cache\n[ocamlformat.0.16.0] downloaded from https://github.com/ocaml-ppx/ocamlformat/releases/download/0.16.0/ocamlformat-0.16.0.tbz\n[ppx_derivers.1.2.1] found in cache\n[ppx_yojson_conv_lib.v0.14.0] found in cache\n[re.1.9.0] found in cache\n[result.1.5] found in cache\n[sexplib0.v0.14.0] found in cache\n[stdio.v0.14.0] found in cache\n[stdlib-shims.0.1.0] found in cache\n[topkg.1.0.3] found in cache\n[tyxml.4.4.0] found in cache\n[ppxlib.0.20.0] downloaded from https://github.com/ocaml-ppx/ppxlib/releases/download/0.20.0/ppxlib-0.20.0.tbz\n[uchar.0.0.2] found in cache\n[uucp.13.0.0] found in cache\n[uuseg.13.0.0] found in cache\n[uutf.1.0.2] found in cache\n[yojson.1.7.0] found in cache\n\n<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>\n-> installed seq.base\n-> installed conf-m4.1\n-> installed cmdliner.1.0.4\n-> installed ocamlbuild.0.14.0\n-> installed ocamlfind.1.8.1\n-> installed uchar.0.0.2\n-> installed topkg.1.0.3\n-> installed uutf.1.0.2\n-> installed astring.0.8.5\n-> installed fpath.0.7.3\n-> installed dune.2.7.1\n-> installed cppo.1.6.6\n-> installed dune-build-info.2.7.1\n-> installed easy-format.1.3.2\n-> installed fix.20201120\n-> installed menhirLib.20201122\n-> installed menhirSdk.20201122\n-> installed ocaml-version.3.0.0\n-> installed ppx_derivers.1.2.1\n-> installed result.1.5\n-> installed sexplib0.v0.14.0\n-> installed stdlib-shims.0.1.0\n-> installed uucp.13.0.0\n-> installed re.1.9.0\n-> installed ocaml-migrate-parsetree.2.1.0\n-> installed ocaml-compiler-libs.v0.12.3\n-> installed csexp.1.3.2\n-> installed tyxml.4.4.0\n-> installed biniou.1.2.1\n-> installed uuseg.13.0.0\n-> installed dune-configurator.2.7.1\n-> installed yojson.1.7.0\n-> installed menhir.20201122\n-> installed odoc.1.5.2\n-> installed ppx_yojson_conv_lib.v0.14.0\n-> installed dot-merlin-reader.3.4.2\n-> installed ppxlib.0.20.0\n-> installed base.v0.14.0\n-> installed stdio.v0.14.0\n-> installed ocaml-lsp-server.1.3.0\n-> installed ocamlformat.0.16.0\nDone.\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "which",
    "ocamllsp"
  ],
  "result": {
    "exitCode": 0,
    "stdout": "/home/Max/.opam/4.11.1+mingw64c/bin/ocamllsp\n",
    "stderr": ""
  }
}

external command {
  "bin": "C:\\dev\\ocaml\\usr\\local\\bin\\ocaml-env.exe",
  "args": [
    "exec",
    "--",
    "opam",
    "exec",
    "--switch=4.11.1+mingw64c",
    "--",
    "ocamllsp"
  ]
}
````

</details>

In these logs, `which` is shown to return 0, but it still creates the new switch.

**Comment by tmattio** [comment]
> Installation of the tools seems to work fine, but it installs them even if they're already available in my current switch.

That's the expected behavior, or at least the one I intended. The binaries used should still be the ones found in your switch, but the toolchain is set up regardless of the state of your sandbox. (since the toolchain is global to every project, and not specific to your current project).

I'm not sure if everyone agrees with the behavior though?

**Comment by mnxn** [comment]
It's a bit weird user experience, since it takes 5 minutes to compile all the tools just to use what I already had available. 
I guess it's fine since it's a one-time setup. (We should make it possible to update the tools in this toolchain in a later PR)


A few more things I noticed:

- `vscode-ocaml-toolchain` doesn't show up in the toolchain selection:

| Before | After |
| - | - |
| before | after |

Is this intentional?


- If I try to switch to a global sandbox (that doesn't have ocamlc avilable), the sandbox doesn't change and I get a `ERR spawn ocamlc ENOENT` in the developer console. I don't think the extension is checking whether `ocamlc` exists before doing `ocamlc --version`.

- `which` is not available on Windows unless the user has the cygwin executable on their PATH. We should use command prompt's `where` instead, which should work everywhere on windows (`cmd /c where`). This has the benefit of returning Windows paths in the output channel logs.

File: src/sandbox.ml [hunk 6]
```diff
@@ -413,24 +414,29 @@ let get_command sandbox bin args : Cmd.t =
...
+  match sandbox with
+  | Opam (opam, switch) -> Some (Opam.exec opam ~switch ~args:tools)
+  | Esy (esy, manifest) -> Some (Esy.exec esy ~manifest ~args:tools)
+  | _ -> None
+
+let ocaml_version sandbox =
+  let cmd = get_command sandbox "ocamlc" [ "--version" ] in
+  let open Promise.Result.Syntax in
+  let+ output = Cmd.output cmd in
+  String.strip output
```
mnxn, line 442: Need a `Cmd.check` here.

src/sandbox.ml
```diff
@@ -413,24 +414,29 @@ let get_command sandbox bin args : Cmd.t =
     in
     Shell command
 
-let get_lsp_command ?(args = []) sandbox : Cmd.t =
-  get_command sandbox "ocamllsp" args
+let has_command sandbox bin =
+  let open Promise.Syntax in
+  let cmd = get_command sandbox "which" [ bin ] in
```
mnxn, line 419: On Windows, this should be `get_command sandbox "cmd" [ "/c"; "where"; bin ]`.

mnxn (changes requested):

tmattio:
> vscode-ocaml-toolchain doesn't show up in the toolchain selection: Is this intentional?

Yes, the toolchain switches should not be tampered with by the user, so not displaying it in the sandbox candidates mitigates the risk.

mnxn (approved):

mnxn:
Looks good to me. Haven't encountered any other problems on Windows so far.

The README should probably be updated to describe the new workflow.

rgrinberg:
Hold off on merging for now. I'd like to test this out myself as well.

rgrinberg:
Any idea why CI failed on windows?

rgrinberg:
Just tested this out. Seems to be working smoothly. However, I have some concerns:

* These internal switches should not be visible to the user. Source based installation using opam is just an implementation detail.  Can we try using `OPAMROOT` to isolate our internal switches?

* We run the solver as part of the installation. That's quite fragile by itself and is incompatible with our plans for binary installation. We should have a fixed build plan for `os * ocaml` using lock files. And binary installation should use binaries that correspond to these build plans.

* We aren't handling upgrades. We should allow the user the option to upgrade ocamllsp when new versions are released.

tmattio:
> We should have a fixed build plan for os * ocaml using lock files.

I'm not sure how you would do that. AFAIK, opam can only install dependencies from lock files if they are in the filesystem, are you suggesting we store the lock file somewhere and execute the opam command where we've stored them?

rgrinberg:
> are you suggesting we store the lock file somewhere and execute the opam command where we've stored them?

Shipping such lock files with the plugin is certainly one option. It's not the only option, but it seems reasonable to me.

tmattio:
I have added support for upgrading the platform tools, but I now realize it is incompatible with providing a fixed build plan for the vscode switches (because upgrading will update to the latest dependencies, which we don't control).

A workaround would be to trigger the upgrade when the build plan for the switch has changed (this happens when we release a new version of the extension with an updated build plan). In which case we import the switch once again.

Alternatively, I'm in favor of not providing any build plan at all, even for the initial installation. IMO using Opam solver is solid enough considering that source installation will be the fallback and will be replaced with an installation of pre-built binaries as you mentioned.

rgrinberg:
I've thought over this and I'm still not sure what's the best course of action. On one hand, I hate having the solver decide the user's fat, on the other hand, I'm not 100% confident with making this work with lockfiles.

> A workaround would be to trigger the upgrade when the build plan for the switch has changed (this happens when we release a new version of the extension with an updated build plan). In which case we import the switch once again.

That sounds like the reasonable thing to do, and I don't see upgrades as the problem. The problem is that I imagine the whole process to generate lockfiles for all these combinations will be too unwieldy. The way I see it, we have 3 options:

1) Accept the PR as is.
2) Use lockfiles
3) Wait until binary installation.

@mnxn do you have a preference for any of these options?

tmattio:
I thought more about it too. There are two problems with the current approach:

- The platform tools may have dependency conflicts
- The switch creation takes way too long

I explored two alternatives:

- Reducing the switch creation time by caching the compiler (the same way Esy does). However, the first switch creation does not benefit from it, and that's the most important for us.
- Building the platform-tools with ocaml-ci, and downloading the binaries. However, ocaml-ci is not quite ready yet and we will have to provide a solution from source as a fallback anyway.

I discussed with @NathanReb and we thought one possible solution could be to make all of the platform-tools mono repositories (using `opam-monoreop`). In this scenario: 

- We wouldn't have to manage a separate opam switch, as there wouldn't be any dependency conflict.
- We could re-use the user's switch and install the tools in it if they are not already installed.
- We wouldn't need lock files, as the only dependency would be dune.

It seems to address all of the limitations of the other approaches while keeping the solution simple.

What do you think?

cc @emillon - Are we at a stage where we can release monorepos on opam?

rgrinberg:
I think using monorepo is fine, but we must be able to install it in the background ourselves (preferably without opam). If the user has to install this new tool manually, it defeats the purpose of automating the installation of lsp.

tmattio:
We discussed this PR with @avsm and think it can be merged as-is.

We agreed that there are some shortcomings in the OCaml Platform workflow that we need to address, but they will take time to put in place, and this PR still brings a lot of value for newcomers in the meantime.

@rgrinberg @mnxn I rebased and did some QA, but if you have some cycles to test this one last time, it would be great as it's a pretty invasive change. Also, let me know if you have any other concerns about the implementation.

src/toolchain.ml
```diff
@@ -0,0 +1,260 @@
+open Import
+
+let opamroot =
+  let home = Node.Os.homedir () in
```
mnxn, line 4: I changed this to use Node's `os.homedir` because Windows doesn't use the `$HOME` environmental variables.

Before this change, the extension failed to activate on Windows.

tmattio: I was sure Windows used `HOME`! Thanks @mnxn

src/toolchain.ml
```diff
@@ -0,0 +1,260 @@
...
+          | true ->
+            let* () = upgrade t in
+            Tool.detect_missing t
+        in
+        let+ () = install_missing_tools t missing_tools in
+        ()
+    in
+    Promise.return (Ok t)
+
+let to_pretty_string { version; _ } = version
```
mnxn, line 253: I think it would be better if the pretty string had "OCaml" at the beginning so it's more apparent what extension the status bar item is coming from.
```suggestion
let to_pretty_string { version; _ } = "OCaml " ^ version
```

mnxn:
@tmattio:

I pushed a commit that uses the opam `--root` flag instead of the OPAMROOT environmental variable.

OPAMROOT didn't seem to work with Windows and `ocaml-env`, I always got errors like
```
[ERROR] Opam has not been initialised, please run `opam init'
```

I also changed the toolchain setup to wait for opam to finish installing the tools. Otherwise, the language server wouldn't start. Upgrading toolchain packages is still done asynchronously.

src/toolchain.ml
```diff
@@ -0,0 +1,240 @@
...
+end
+
+let opam_init opam opamroot =
+  let open Promise.Syntax in
+  let* exists = Fs.exists (opamroot |> Path.to_string) in
+  if exists then
+    Promise.Result.return ()
+  else
+    let+ result = Opam.init opam |> Cmd.output in
+    Result.map ~f:(fun _ -> ()) result
```
mnxn, line 89: `opam init` takes a bit of time on the computers I tried it on. Maybe it should display progress too?

mnxn:
Just noticing this now, but I think the treeview buttons for removing an entire switch should require a confirmation of some kind. Same for removing a package.

Maybe in another PR?

tmattio:
Thanks @mnxn! I improved the workflow a bit to display progress for init and make sure that we start the server before we try to upgrade the toolchain (otherwise, opam exec seems to wait for the update to finish).

src/toolchain.ml
```diff
@@ -135,59 +127,71 @@ let get_install_command { opam = opam, switch; _ } tools =
...
+    ~options:progress_options ~task
+
+let opam_init opam opamroot =
+  let open Promise.Syntax in
+  let* exists = Fs.exists (opamroot |> Path.to_string) in
+  if exists then
+    Promise.Result.return ()
+  else
+    with_progress
+      ~msg:"Installing Platform Tools (this might take a few minutes)"
```
mnxn, line 155: I think this should have a different message from `install_missing_tools` so the user has a little more information.

Maybe `Initializing Platform Toolchain (this might take a few minutes)`?

vscode/interop/interop.ml
```diff
@@ -117,6 +125,31 @@ module Js = struct
...
+      ; ok : Ojs.t
+      ; error : Ojs.t
+      }
+    [@@js]
+
+    let t_of_js js_val =
+      match js_result_of_js js_val with
+      | { case = "ok"; ok; _ } -> Ok ([%js.to: Ok.t] ok)
+      | { case = "error"; error; _ } -> Error ([%js.to: Error.t] error)
+      | _ -> assert false
```
mnxn, line 142: I changed this implementation because the previous version would reach the `assert false` when the `Ok` type is `unit`:
```
rejected promise not handled within 1 second: 0,248,Assert_failure,-10,0,vscode/interop/interop.ml,142,13
```

Because `unit` was turned into `undefined`, the function couldn't determine the difference between a unit value or the absence of one.

Checking a tag like I do here should avoid the problem.

mnxn (approved):
Works well for me. Good job!

rgrinberg (approved):

PR #464, opened 2020-12-01
Fix opam switch detection

tmattio:
The output of the `opam switch show` command contains whitespace, so we need to strip the output.


rgrinberg (approved):

mnxn (approved):

rgrinberg:
Btw, this should technically be in CHANGES.

PR #466, opened 2020-12-02
Rename toolchain to sandbox

tmattio:
Extracted from #463


.ocamlformat
```diff
@@ -3,6 +3,8 @@ break-separators=before
 dock-collection-brackets=false
 break-sequences=true
 doc-comments=before
+wrap-comments = true
+parse-docstrings = true
```
rgrinberg, line 7: As mentioned. Let's just copy it from dune, but use the latest 0.15.x

rgrinberg (approved):
Looks good to me. To clarify the terminology:

* sandbox - a unit of "isolation" that allows us to resolve binaries and libraries 

* toolchain - a specific sandbox we use for tools.

Currently we conflate the two by installing tools into the user’s build sandbox. In the future, we'll have a separate sandbox for the toolchain. 

Do these terms make sense for everyone?

PR #471, opened 2020-12-06
Initialize extension correctly

rgrinberg:
We should initialize the extension and register all commands even if the
language server fails to load.

### DIFF
--- a/src/vscode_ocaml_platform.ml
+++ b/src/vscode_ocaml_platform.ml
@@ -19,29 +19,41 @@ let activate (extension : ExtensionContext.t) =
      vscode [output] pane for logs *)
   Process.Env.set "OCAML_LSP_SERVER_LOG" "-";
   let open Promise.Syntax in
-  let* sandbox = Sandbox.of_settings_or_detect () in
-  let is_fallback = Option.is_empty sandbox in
-  let sandbox = Option.value sandbox ~default:Sandbox.Global in
-  let+ instance = Extension_instance.make sandbox in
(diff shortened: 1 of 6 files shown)


mnxn (approved):
Looks good to me.
Thanks for the quick fix!

rgrinberg:
@mnxn I pushed another commit to make things more robust. We should register the extension as soon as possible without waiting for any promises. Lastly, I've also simplified a bunch of redundant api's.

src/vscode_ocaml_platform.ml
```diff
@@ -19,29 +19,42 @@ let activate (extension : ExtensionContext.t) =
...
-       sandbox: they may ignore the pop-up leaving the extension hanging, so we
-       use fallback; w/ a proper detection mechanism, we would redo work in rare
-       cases *)
-  then
-    let (_ : unit Promise.t) = suggest_to_pick_sandbox instance in
+  let sandbox = Sandbox.of_settings_or_detect () in
+  let (_ : unit Promise.t) =
+    let* sandbox = sandbox in
+    let is_fallback = Option.is_empty sandbox in
+    (* register things with vscode, making sure to register their disposables *)
```
mnxn, line 32: I think this comment should be moved a couple lines up to where the registration functions are being called.

mnxn (approved):

src/vscode_ocaml_platform.ml
rgrinberg: I removed this comment entirely. It only explains the obvious.

PR #477, opened 2020-12-13
Change the label name used in the changelog action

smorimoto:


smorimoto:
This just removes the delimiters and unifies other label styles.

tmattio (approved):

PR #478, opened 2020-12-15
Update promise_jsoo and jsonoo dependencies

mnxn:
Updates to the latest version of `promise_jsoo` and `jsonoo`, which stop unnecessarily linking to the `js_of_ocaml-ppx` library. 

The size of the generated JS file decreases to about 9 MB and the VSIX file decreases to about 2 MB.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -35,9 +35,9 @@
   (core_kernel
    (>= v0.14.0))
   (promise_jsoo
-   (>= 0.3.0))
+   (>= 0.3.1))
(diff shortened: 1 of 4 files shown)


tmattio:
I'm surprised that removing an unused library makes the compiled JS code smaller. I thought Jsoo was doing dead code elimination?

rgrinberg (approved):
There are probably module side effects somewhere in the ppx stack.

PR #481, opened 2020-12-17
Use ocaml-env for opam commands on windows

mnxn:
This PR makes the extension execute `ocaml-env exec -- opam` for any opam commands on Windows. Previously, VSCode would have to be opened from cygwin to have a proper opam environment. Now, VSCode can be opened normally and the extension handles the environment.

I made the `ocaml.useOcamlEnv` setting default to true because this should be the smoothest way to use the extension on Windows now. Having the option lets users continue to use the old workflow by accessing `opam` directly.

@tmattio: I think this should (partially) help implement #463 on Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Highlight token aliases in Menhir associativity declarations (#473)
 - Activate the extension when workspace contains OCaml, Reason sources or
   project marker files. (#482)
+- Add `ocaml.useOcamlEnv` setting to determine whether to use `ocaml-env` for
+  opam commands from OCaml for Windows (#481)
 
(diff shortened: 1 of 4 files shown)


tmattio (approved):
Awesome, thanks for doing this @mnxn!

rgrinberg (approved):

rgrinberg:
I forgot about the error case - what happens when ocaml-env is absent?

mnxn:
> I forgot about the error case - what happens when ocaml-env is absent?

Due to the use of `Cmd.check_spawn`, the behavior is the same as if `opam` is absent: opam won't show up in the available sandboxes. It won't interfere if the user is using esy or a global sandbox.

PR #482, opened 2020-12-17
Add rules for project detection

tmattio:
Currently, the extension is activated only when the user opens a supported file. This PR adds rules so that the extension is activated when opening an OCaml project.


package.json
```diff
@@ -31,7 +31,24 @@
     "onLanguage:dune-workspace",
     "onCommand:ocaml.select-sandbox",
     "onCommand:ocaml.open-terminal",
-    "onCommand:ocaml.open-terminal-select"
+    "onCommand:ocaml.open-terminal-select",
+    "onCommand:ocaml.open-terminal-select",
```
smorimoto, line 35: This seems to be an unintentional duplication.

tmattio: Indeed, thanks!

package.json
```diff
@@ -31,7 +31,23 @@
...
     "onCommand:ocaml.select-sandbox",
     "onCommand:ocaml.open-terminal",
-    "onCommand:ocaml.open-terminal-select"
+    "onCommand:ocaml.open-terminal-select",
+    "workspaceContains:dune-project",
+    "workspaceContains:**/dune-project",
+    "workspaceContains:*.opam",
+    "workspaceContains:**/*.opam",
+    "workspaceContains:*.ml",
+    "workspaceContains:**/*.ml",
```
rgrinberg, line 40: Are you sure we need both `*.foo` and `**/*.foo`?

From the docs:

> ** to match any number of path segments, including none

So I think `**/.foo` should be enough.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 # Unreleased
 
 - Highlight token aliases in Menhir associativity declarations (#473)
+- The extension is not activated when an OCaml project is detected (#482)
```
rgrinberg, line 6: If we're describing a bug fix, we should allude to that with something like

> Fix activation of extension. The extension was not activated when an OCaml project was detected.

Or we can simply describe the new behavior:

> Activate the extension when workspace contains OCam, Reason sources or project marker files.

tmattio: Meant to write "now":
"The extension is now activated when an OCaml project is detected"

tmattio: But your version is more precise, will replace it :)

package.json
tmattio: Seems to work as well.

rgrinberg (approved):

PR #483, opened 2020-12-17
Add bindings for tree view APIs

tmattio:
This PR adds bindings to the TreeView API.

I copied the documentation from the VSCode API for the functions of the modules that I added, but if it's not a habit we'd like to take when adding new bindings, I can remove it.


rgrinberg:
> I copied the documentation from the VSCode API for the functions of the modules that I added, but if it's not a habit we'd like to take when adding new bindings, I can remove it.

Whatever makes it easier for the development of this plugin.

I skimmed through the bindings and it looks good. Why do we need to copy the proposed api?

tmattio:
> Whatever makes it easier for the development of this plugin.

It doesn't change anything for end-users. it's purely if we, as developers of the extension, prefer to have the documentation of the bindings, or if we prefer to refer to the VSCode API documentation instead.

> Why do we need to copy the proposed api?

I'm not sure.. From what I can see, the bindings here should not need the proposed API, but when I'm using them and run VSCode, I have a popup error saying proposed APIs are not available during development.

rgrinberg:
> I'm not sure.. From what I can see, the bindings here should not need the proposed API, but when I'm using them and run VSCode, I have a popup error saying proposed APIs are not available during development.

That's a bit concerning. I don't mind using a proposed API, but we should identify what's the proposed API that we are using.

tmattio:
I tried again and don't have the error anymore, that was probably a binding I added and removed afterward. I removed the proposed API

mnxn:
> I copied the documentation from the VSCode API for the functions of the modules that I added, but if it's not a habit we'd like to take when adding new bindings, I can remove it.

I think we should avoid copying VSCode documentation unless we can automatically generate it. The API documentation changes quite often between releases so it's likely that the copied documentation will out-of-date at some point. The documentation on the VSCode website will always be up-to-date and it is more complete since our bindings don't cover the whole API.

vscode/vscode/vscode.ml
```diff
@@ -1810,39 +1812,254 @@ module Workspace = struct
...
+
+  let iconPath_of_js js_val =
+    if Ojs.has_property js_val "path" then
+      `Uri ([%js.to: Uri.t] js_val)
+    else if Ojs.has_property js_val "id" then
+      `ThemeIcon ([%js.to: ThemeIcon.t] js_val)
+    else if Ojs.type_of js_val = "string" then
+      `String ([%js.to: string] js_val)
+    else
+      assert false
```
mnxn, line 1875: For `iconPath_of_js`, what if js_val is a lightDarkIcon? That would assert false right now.

vscode/vscode/vscode.ml
```diff
@@ -1810,39 +1812,254 @@ module Workspace = struct
...
     [@@js.get "vscode.window.activeTextEditor"]
 
-  val showQuickPick :
-       choices:QuickPickItem.t list
-    -> ?options:QuickPickOptions.t
-    -> ?token:CancellationToken.t
-    -> unit
-    -> QuickPickItem.t or_undefined Promise.t
-    [@@js.global "vscode.window.showQuickPick"]
+  val visibleTextEditors : unit -> TextEditor.t array
```
mnxn, line 2033: We should prefer to use lists for the bindings since they're easier to manipulate in the extension code.

vscode/vscode/vscode.ml
```diff
@@ -1810,39 +1812,254 @@ module Workspace = struct
...
+    [@@js.get "vscode.window.onDidChangeActiveTextEditor"]
 
-  val showQuickPick :
-       items:string list
-    -> ?options:QuickPickOptions.t
-    -> ?token:CancellationToken.t
-    -> unit
-    -> string or_undefined Promise.t
-    [@@js.global "vscode.window.showQuickPick"]
+  val onDidChangeVisibleTextEditors : unit -> TextEditor.t array Event.t
```
mnxn, line 2039: Use a list here too.

vscode/vscode/vscode.ml
tmattio: Done, thanks!

vscode/vscode/vscode.ml
tmattio: Done

vscode/vscode/vscode.ml
```diff
@@ -1810,39 +1812,275 @@ module Workspace = struct
...
+      let dark_js = Ojs.get js_val "dark" in
+      let light =
+        if Ojs.has_property light_js "parse" then
+          `Uri (Uri.t_of_js light_js)
+        else
+          `String ([%js.to: string] light_js)
+      in
+      let dark =
+        if Ojs.has_property dark_js "parse" then
+          `Uri (Uri.t_of_js dark_js)
```
mnxn, line 1870: `Uri.t_of_js` can be `[%js.to: Uri.t]` here.

tmattio: Are there any advantages of using the PPX, or is it just a convention?

mnxn: I think the PPX just calls the function, but using it here keeps it consistent with the rest of the bindings and makes it clear which type the JS value is being converted to.

mnxn (approved):

PR #486, opened 2020-12-18
Add an OCaml activity tab

tmattio:
**Warning for interested users**: this is an experimental feature. We will iterate on it heavily until we consider it stable, until then, you should expect that features in the activity tab will be added and removed without notice. Your feedback is welcome, so don't hesitate to open an issue to report bugs or suggest ideas.

---

This PR introduces a new activity tab for OCaml with the following sections:

- **PACKAGES** - List of Opam switches and packages in each of them
- **COMMANDS** - Commands provided by the VSCode extension, such as sandbox selection and lsp server refresh.
- **HELP AND FEEDBACK** - Help and feedback section with links to community communication channels

image

## Questions

- What should we do with the dependencies? Displaying a list does not seem very useful. Should we allow users to update/delete them in this PR, or do we want to merge as-is and introduce new functionalities in other PRs?
- Which commands would you like to have? Would a build/clean command be helpful?
- Do you see a category that should be included in the first version?


rgrinberg:
Good work. Some initial feedback:

* Let's remove the tutorials link from feedback for now. This thing is far too out of date to be useful.
* The name "Packages" over the switches section should be renamed to "Switches"
* Can we highlight the currently selected switch in that view?
* Does it make sense to perhaps include a view for the currently selected sandbox? I'm assuming that this will be more useful than seeing all switches most of the time.
* Is there a way to visually distinguish package names from versions inside the package listing?
* Can we show the version of OCaml used for switches? This is quite useful for local switches for example.
* Do we really need the restart language server command to be listed? I know the scala plugin does it, but I just don't see this command as a useful part of a workflow (unless our extension is very broken).

src/treeview_packages.ml
```diff
@@ -0,0 +1,151 @@
...
+    | Switch (Named name) ->
+      List [ Atom "switch"; List [ Atom "named"; Atom name ] ]
+    | Switch (Local path) ->
+      let name = Path.to_string path in
+      List [ Atom "switch"; List [ Atom "local"; Atom name ] ]
+    | Dependency (dep, None) -> List [ Atom "dependency"; Atom dep ]
+    | Dependency (dep, Some t) ->
+      List [ Atom "dependency"; Atom dep; List [ Atom "parent"; sexp_of_t' t ] ]
+
+  let sexp_of_t = sexp_of_t'
```
rgrinberg, line 20: Why do we need this `sexp_of_t'` alias?

src/treeview_help.ml
```diff
@@ -0,0 +1,118 @@
...
+
+let register extension =
+  let extension_path = Vscode.ExtensionContext.extensionPath extension in
+  let treeDataProvider =
+    Vscode.TreeDataProvider.create
+      ~getTreeItem:(getTreeItem ~extension_path)
+      ~getChildren:(getChildren ~extension_path)
+      ()
+  in
+  let _ =
```
rgrinberg, line 114: Is it correct to drop the disposable here?

src/treeview_packages.ml
```diff
@@ -0,0 +1,151 @@
...
+             "An error occured while reading the switch dependencies: %s" err)
+        ()
+    in
+    Promise.return None
+
+let get_dependencies ~extension_path element =
+  let open Promise.Syntax in
+  let dependency =
+    match Vscode.TreeItem.id element with
+    | None -> assert false
```
rgrinberg, line 112: If there's an `assert false`, we should leave a comment that explains why.

src/treeview_packages.ml
```diff
@@ -0,0 +1,151 @@
...
+    | Some dep -> Dependency.of_string dep
+  in
+  match dependency with
+  | Dependency.Switch switch ->
+    let+ deps = get_switch_dependencies switch in
+    deps |> Option.map ~f:(List.map ~f:(make_dependency_item ~extension_path))
```
rgrinberg, line 118: Can we refactor the code so that this common bit:

```
deps |> Option.map ~f:(List.map ~f:(make_dependency_item ~extension_path))
```

only appears once.

src/treeview_packages.ml
```diff
@@ -0,0 +1,151 @@
...
+      match element with
+      | None -> `Promise (Promise.return (Some items))
+      | Some element -> `Promise (get_dependencies ~extension_path element)
+    in
+    let getTreeItem ~element = Promise.return element in
+    let treeDataProvider =
+      Vscode.TreeDataProvider.create ~getTreeItem ~getChildren ()
+    in
+    let _ =
+      Vscode.Window.registerTreeDataProvider ~viewId:"ocaml-packages"
```
rgrinberg, line 148: Same as before: is it ok to drop disposables like this?

src/treeview_help.ml
tmattio: We have no use for it, I don't think this has any impact.

src/treeview_help.ml
```diff
@@ -0,0 +1,118 @@
...
+  ; discuss_item ~extension_path
+  ; github_item ~extension_path
+  ]
+
+let getTreeItem ~extension_path:_ ~element = Promise.return element
+
+let getChildren ~extension_path ~element =
+  match element with
+  | None -> `Promise (Promise.return (Some (items ~extension_path)))
+  | Some _ -> `Promise (Promise.return (Some []))
```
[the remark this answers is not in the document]
tmattio: What's `Value` here?

tmattio:
- [ ] Let's remove the tutorials link from feedback for now. This thing is far too out of date to be useful.
- [ ] The name "Packages" over the switches section should be renamed to "Switches"
- [ ] Can we highlight the currently selected switch in that view?
- [ ] Does it make sense to perhaps include a view for the currently selected sandbox? I'm assuming that this will be more useful than seeing all switches most of the time.
- [ ] Is there a way to visually distinguish package names from versions inside the package listing?
- [ ] Can we show the version of OCaml used for switches? This is quite useful for local switches for example.
- [ ] Do we really need the restart language server command to be listed? I know the scala plugin does it, but I just don't see this command as a useful part of a workflow (unless our extension is very broken).

src/treeview_help.ml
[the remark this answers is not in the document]
rgrinberg: It's the constructor for this variant:

```
  type 'a t =
    [ `Value of 'a option
    | `Promise of 'a option Promise.t
    ]
```

src/treeview_help.ml
rgrinberg: To restart the extension, we need to dispose of it and all the resources it acquired. So I think we should also dispose the tree acquired here as well. See `Extension_instance.disposable`.

rgrinberg: In previous versions of the plugin, we'd drop these disposables and it would make restarting the plugin after an upgrade wonky.

tmattio: Done, thanks :)

src/treeview_help.ml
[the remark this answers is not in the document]
tmattio: Done!

tmattio:
Thanks for the feedback Rudi!

I addressed most of the items you suggested, and also added the following:

- Open documentation of a package
- Remove a switch
- Refresh switches

There are two points I didn't address:

> Can we highlight the currently selected switch in that view?

The only highlighting possible in the tree view seems to be the same as when you search a string, as shown here:



That's probably not what we want and I don't think there is another way to highlight a label.

If we have a view for the current sandbox, that's not going to be critical anyways IMO.

> Does it make sense to perhaps include a view for the currently selected sandbox?

Yes! What would see in this view? Would it just be a list of packages in the current switch?

I agree this is something very useful, but maybe we want to bring it in another PR, this one being quite large already.

FYI, here's what the UI looks like now:

image

src/treeview_switches.ml
```diff
@@ -0,0 +1,207 @@
...
+      assert false
+    | Some dep -> Dependency.of_string dep
+  in
+  let dependencies_opt_to_tree_items deps_opt =
+    let* deps_opt = deps_opt in
+    match deps_opt with
+    | None -> Promise.return None
+    | Some deps ->
+      let* items =
+        List.fold_left ~init:(Promise.return [])
```
rgrinberg, line 148: I'd add `Promise.List.all` to `promise_jsoo`. It's ok to just leave a TODO for now.

src/treeview_switches.ml
```diff
@@ -0,0 +1,207 @@
...
+                  let+ deps =
+                    make_item ~extension_path (Dependency.Switch switch)
+                  in
+                  Some deps)
+              switches
+          in
+          Some items
+      in
+      `Promise items
+    | Some element -> `Promise (get_dependencies ~extension_path element)
```
rgrinberg, line 193: This clause should come first because it's shorter.

src/treeview_switches.ml
```diff
@@ -0,0 +1,207 @@
...
+      let items =
+        match opam with
+        | None -> Promise.return None
+        | Some opam ->
+          let* switches = Opam.switch_list opam in
+          let+ items =
+            Promise.List.filter_map
+              (function
+                | Opam.Switch.Named name
+                  when String.is_prefix name ~prefix:"vscode-ocaml-toolchain" ->
```
rgrinberg, line 181: I'd remove this special prefix detection for now, because I'd prefer if we handled private switches differently.

src/opam.ml
```diff
@@ -23,6 +23,119 @@ module Switch = struct
     | Local x, Local y -> Path.equal x y
     | Named x, Named y -> String.equal x y
     | _, _ -> false
+
+  let path =
+    let home = Node.Process.Env.get "HOME" |> Stdlib.Option.get in
+    function
+    | Named s -> Path.(of_string home / ".opam" / s)
```
rgrinberg, line 30: This should be `$ opam var root` to respect `OPAMROOT`.

rgrinberg (approved):
Ready to merge after the last few minor remarks are addressed.

rgrinberg:
> That's probably not what we want and I don't think there is another way to highlight a label.

Alright, let's drop this for now.

> Yes! What would see in this view? Would it just be a list of packages in the current switch?

Yes, that's what I had in mind. It can come later.

src/treeview_switches.ml
tmattio: Ah yes, I did the private switch modification in the other PR also. Will remove

src/treeview_switches.ml
tmattio: Pushed here: https://github.com/mnxn/promise_jsoo/pull/5

rgrinberg: So this can just use `Promise.all_list` for now.

src/treeview_switches.ml
```diff
@@ -0,0 +1,198 @@
...
+    dependencies_opt_to_tree_items deps
+
+let register extension =
+  let open Promise.Syntax in
+  let extension_path = Vscode.ExtensionContext.extensionPath extension in
+  let+ opam = Opam.make () in
+  let getChildren ~element :
+      [ `Promise of TreeItem.t list option Promise.t
+      | `Value of TreeItem.t list option
+      ] =
```
mnxn, line 163: This type annotation can be `TreeItem.t list Vscode.ProviderResult.t`

tmattio: Thanks @mnxn, just added it to debug something, I'll remove it instead

src/extension_commands.ml
```diff
@@ -84,10 +86,78 @@ let switch_impl_intf =
...
+             "Cannot delete a switch from a package dependency."
+      | Switch switch -> (
+        let open Promise.Syntax in
+        let* opam_opt = Opam.make () in
+        match opam_opt with
+        | None ->
+          Promise.return
+          @@ show_message `Warn "Opam could not be found on your system."
+        | Some opam -> (
+          let* result = Opam.remove_switch opam switch |> Cmd.output in
```
mnxn, line 111: The following `Promise.return`s can be removed if this is a map (`let+`).

src/extension_commands.ml
```diff
@@ -84,10 +86,78 @@ let switch_impl_intf =
...
+          | Ok _ ->
+            let (_ : Ojs.t option Promise.t) =
+              Vscode.Commands.executeCommand
+                ~command:Extension_consts.Commands.refresh_switches ~args:[]
+            in
+            Promise.return
+            @@ show_message `Info "The switch has been removed successfully." )
+        )
+    in
+    let (_ : unit Promise.t) = try_removing () in
```
rgrinberg, line 123: Why do we need this `try_removing` wrapper? Also, why not wait for the promise to finish? (by returning it)

tmattio: > Why do we need this try_removing wrapper

We don't :)

> Also, why not wait for the promise to finish

The handler must have type `Extension_instance.t -> args:Ojs.t list -> unit`,  so I don't think we can wait for the promise.

src/treeview_help.ml
```diff
@@ -0,0 +1,93 @@
...
+    Vscode.TreeDataProvider.create
+      ~getTreeItem:(getTreeItem ~extension_path)
+      ~getChildren:(getChildren ~extension_path)
+      ()
+  in
+  let disposable =
+    Vscode.Window.registerTreeDataProvider ~viewId:"ocaml-help"
+      ~treeDataProvider
+  in
+  Promise.return disposable
```
ulugbekna, line 93: IIUC, even though this function returns a promise, it runs synchronously unless you use `Promise.make` not `Promise.return`, which just wraps the value in a promise and nothing more. Was this the intention?

ulugbekna: Same for `register` in other treeview_... files.

src/extension_commands.mli
```diff
@@ -1,8 +1,11 @@
...
-  ; handler : Extension_instance.t -> unit -> unit
+  ; handler : Extension_instance.t -> args:Ojs.t list -> unit
   }
 
+val make_command :
+  string -> (Extension_instance.t -> args:Ojs.t list -> unit) -> command
+
 (** Module to manage commands[1] across the extension.
 
     Module does not have public API for command creation on purpose. One should
```
ulugbekna, line 11: The comment says specifically that there is no `make_command` in the module API, so that the module handles command disposable subscription across the whole extension. 

It may be worth to either respect the module's goal of handling all commands (e.g., by exposing current unexposed `command` and use it along with `register_all_commands`) or remove that comment.

src/vscode_ocaml_platform.ml
```diff
@@ -26,6 +27,25 @@ let activate (extension : ExtensionContext.t) =
...
+      handler
+    |> Extension_commands.register extension instance
+  in
+  let (_ : unit Promise.t) =
+    let+ disposable = Treeview_commands.register extension in
+    ExtensionContext.subscribe extension ~disposable
+  in
+  let (_ : unit Promise.t) =
+    let+ disposable = Treeview_help.register extension in
+    ExtensionContext.subscribe extension ~disposable
```
ulugbekna, line 47: More of a question: are we sure `ExtensionContext.subscribe extension ~disposable` is thread-safe? From this code, the subscriptions happens in a "background" thread, iiuc.

ulugbekna (commented):
Sorry for a late-to-the-party comments.

src/treeview_help.ml
tmattio: `Promise.make` has type `(resolve:('a -> unit) -> reject:('e -> unit) -> unit) -> 'a Promise.t`, so it wouldn't work in this case (can't return the disposable), unless I'm missing something?

src/extension_commands.mli
tmattio: This one was a bit awkward, I wanted to respect the design decision for this module, but the `refresh-switch` command needs access to the event emitter created in the treeview. Passing it in the `register_all_commands` did not make sense IMO.

That's a problem I figure will happen often, especially now that we have a UI that will need the creation of more commands. With this in mind, I think it might make sense to revisit the design of the extension commands to have a single pattern (either we create all of the commands in `vscode_ocaml_platform.ml`, because they have dependencies that should be available here, or we keep them in this module, but it will not be properly isolated).

src/treeview_help.ml
tmattio: I guess we could pass a callback to subscribe instead. I'll try that.

tmattio: Seems I misunderstood how to use `Promise.make`, my bad 🙂

ulugbekna: If we're sure things we're doing are thread-safe, then I think you could do :

```ocaml
let register extension = 
  Promise.make @@ fun ~resolve ~reject -> 
  let extension_path = Vscode.ExtensionContext.extensionPath extension in
  let treeDataProvider =
    Vscode.TreeDataProvider.create
      ~getTreeItem:(getTreeItem ~extension_path)
      ~getChildren:(getChildren ~extension_path)
      ()
  in
  Vscode.Window.registerTreeDataProvider ~viewId:"ocaml-help" ~treeDataProvider
  |> resolve
```

src/extension_commands.mli
ulugbekna: You decide how to go forward, but if you want to keep the current design: 

Expose `command : name -> handler -> ()` from `extension_commands.ml` that creates a command and adds it to the module-local mutable list `commands`. Call `register_all_commands` in `vscode_ocaml_platform.ml`, which uses `commands` to register all commands with the ext context.

src/treeview_help.ml
ulugbekna: > Seems I misunderstood how to use Promise.make, my bad 🙂
Not at all! Using promises API can be tricky at the beginning. 
I found [this](https://www.npmjs.com/package/bs-promise) to be helpful to see how to create promises.

src/vscode_ocaml_platform.ml
tmattio: I'm not sure how to test this, but:

- I did not notice any unexpected behavior that would be explained by this.
- Looking at VSCode extension codes online, subscribing in an async function is not something uncommon.

ulugbekna:
btw, you can as well pass the ext context to register function and register
the disposable there as in other `register` fns such as
`Dune_formatter.register` (though they are synchronous).

ulugbekna:
A couple of bugs I experienced after the PR (can create a separate issue, if necessary): 

- [ ] 1. Some commands such as "restart language server" disappeared (at least for me) (when I remove "command palette" spec in `package.json`, the commands are back. I'd assume "when" clauses for the commands may be somewhat off)
- [ ] 2. New commands "remove", "refresh" should specify "category" in their declaration in `package.json`; otherwise, we have a command palette entry simply as "Refresh" with no "OCaml:" prefix or indication that it's for removing a switch
- [ ] 3. `remove` command results in failure if triggered from command palette; I assume it wasn't assumed to be visible from there based on the `package.json` additions. I would make them visible. 
- [ ] 4. Switches in the activity tab show not the compiler version but "base-bigarray.base" (at least for me). I already notified Thibaut about this

The UI is experimental, so those should be no problem except for the first one :-)

tmattio:
Thanks for testing this thoroughly @ulugbekna 🙂 

1 - Where did they disappear from? The Restart Language Server" particularly shouldn't appear in the command tree view, I removed it.
2 - Agree
3 - I'd prefer to make it invisible from the command palette instead. Do you know if that's possible?
4 - Will look into this one, could you provide me with the `switch-state`? (e.g. in `~/.opam/4.11.1/.opam-switch/switch-state`)

ulugbekna:
1. From the command palette. Before I patched the "when" clauses in "command palette" in `package.json`, all commands in "command palette" section of `package.json` weren't displayed in the command palette as in the screenshot below

image

3. Unfortunately, idk. 
4. Here you go (`~/.opam/default/.opam-switch/switch-state`): 

```
opam-version: "2.0"
compiler: [
  "base-bigarray.base"
  "base-threads.base"
  "base-unix.base"
  "ocaml.4.11.1"
  "ocaml-base-compiler.4.11.1"
  "ocaml-config.1"
]
roots: [
  "angstrom.0.15.0"
  "async.v0.14.0"
  "async_kernel.v0.14.0"
 ... (omitted for brevity)
```

I think the problem is that `ocaml.4.11.1` doesn't always come first in the `compiler` section.

tmattio:
I clearly misunderstood what the `compiler` section was, the name is a bit confusing IMO.

It's a bit odd to see both `ocaml.4.11.1` and `ocaml-base-compiler.4.11.1` here, aren't they both compilers? How can they co-exist?

tmattio:
>  Before I patched the "when" clauses in "command palette" in package.json

What was the patch? 🙂

ulugbekna:
I replaced the current "when" with `"when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"`.

I don't like this solution (otherwise, I'd make a PR) because the commands aren't shown if the opened file is not ocaml-related, which is bad; one should be able to restart the lang server or change the sandbox regardless what file they have open. I couldn't find a "when" clause akin to "extensionStatus == activated", however.

ulugbekna:
> I clearly misunderstood what the compiler section was, the name is a bit confusing IMO.

yeah

> It's a bit odd to see both ocaml.4.11.1 and ocaml-base-compiler.4.11.1 here, aren't they both compilers? How can they co-exist?

I have those two in both `default` and my local switch for `ocaml-git`. I guess that duplication doesn't mean there are two compilers. I'm not sure.

tmattio:
>  I couldn't find a "when" clause akin to "extensionStatus == activated", however.

I don't think the commands are contributed if the extension is not activated, so it should be safe to simply remove the when clause

ulugbekna:
sounds good; we can try

tmattio:
I pushed a fix for the first three in https://github.com/ocamllabs/vscode-ocaml-platform/pull/488/commits/06d1667fd6183adb8cb8b1a64726a14970be7d65. The commands are displayed properly, they have a category and the commands that are specific to the tree view are not accessible from the command palette

PR #487, opened 2020-12-23
Support Eliom file extensions

tmattio:


mnxn (approved):
Needs a change log entry. 

Should probably be mentioned in the README's features section.

PR #488, opened 2020-12-23
Register provider and commands in treeview modules

tmattio:
This PR improves the registration of tree view providers:

- The registrations are now async and don't depend on each other
- The tree view modules register their own command, so the `Extension_commands` can stay isolated.

cc @ulugbekna

### DIFF
--- a/src/treeview_commands.ml
+++ b/src/treeview_commands.ml
@@ -46,7 +46,7 @@ let getChildren ~extension_path ~element =
   | None -> `Value (Some (items ~extension_path))
   | Some _ -> `Value (Some [])
 
-let register extension (_ : Extension_instance.t) =
+let register extension =
   let extension_path = Vscode.ExtensionContext.extensionPath extension in
   let treeDataProvider =
     Vscode.TreeDataProvider.create
--- a/src/treeview_commands.ml
+++ b/src/treeview_commands.ml
@@ -47,20 +47,15 @@ let getChildren ~extension_path ~element =
   | Some _ -> `Value (Some [])
 
 let register extension (_ : Extension_instance.t) =
-  let (_ : unit Promise.t) =
-    Promise.make @@ fun ~resolve ~reject:_ ->
-    let extension_path = Vscode.ExtensionContext.extensionPath extension in
-    let treeDataProvider =
-      Vscode.TreeDataProvider.create
-        ~getTreeItem:(getTreeItem ~extension_path)
-        ~getChildren:(getChildren ~extension_path)
-        ()
-    in
-    let disposable =
-      Vscode.Window.registerTreeDataProvider ~viewId:"ocaml-commands"
-        ~treeDataProvider
-    in
-    Vscode.ExtensionContext.subscribe extension ~disposable;
-    resolve ()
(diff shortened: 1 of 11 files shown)


src/treeview_commands.ml
```diff
@@ -46,16 +46,21 @@ let getChildren ~extension_path ~element =
...
-let register extension =
-  let extension_path = Vscode.ExtensionContext.extensionPath extension in
-  let treeDataProvider =
-    Vscode.TreeDataProvider.create
-      ~getTreeItem:(getTreeItem ~extension_path)
-      ~getChildren:(getChildren ~extension_path)
-      ()
+let register extension (_ : Extension_instance.t) =
+  let (_ : unit Promise.t) =
+    Promise.make @@ fun ~resolve ~reject:_ ->
```
rgrinberg, line 51: Unless my understanding of JS is woefully out of date, this is no different than using `Promise.return`. No additionally concurrency is gained from writing the code this way.

ulugbekna:
My comment regarding use of `Promise.return` vs `Promise.make` was just to warn that using `Promise.return` doesn't make anything more asynchronous. If a function contains an IO operation that can be done asynchronously, one should use `Promise.make` to achieve that. 

Since subscription operations don't return a promise, there is no use in wrapping them in a promise either by `make` or `return`. Sorry for misleading.


A quick dirty experiment that seems to show the difference between `return` and `make`: 

```ocaml
let print_async_0 n =
  if n = 2 then read_line () |> ignore;
  print_endline @@ string_of_int n;
  Promise.return ()

let print_async_1 n =
  Promise.make (fun ~resolve ~reject:_ ->
    if n = 2 then read_line () |> ignore;
    string_of_int n |> print_endline |> resolve)

(* with print_async_1: prints 1, 3, and fails
    with print_async_0: prints 1 and fails *)
let () =
  let print_async = print_async_0 in
  let _ = print_async 1 in
  let _ = print_async 2 in
  let _ = print_async 3 in
  ()
```

tmattio:
I removed the promises when they were not useful, but I believe the rest of the changes are still relevant: it improves the setup of treeview and handling of commands.

src/treeview_commands.mli
```diff
@@ -1,3 +1,2 @@
-(** Register the ocaml-commands tree view and returns the provider as a
-    disposable. *)
-val register : Vscode.ExtensionContext.t -> Vscode.Disposable.t Promise.t
+(** Register the ocaml-commands tree view. *)
+val register : Vscode.ExtensionContext.t -> Extension_instance.t -> unit
```
rgrinberg, line 2: Why did you add the `Extension_instance.t` argument? I don't see it being used.

tmattio: Consistency of the API. All of the other `register` functions in the code take a pair of extension context and extension instance. We don't use it though, so I can remove it.

rgrinberg: It's better to get rid of it unless it adds some serious convenience. Much easier to reason about functions when we know what parameters they need.

tmattio: Done

rgrinberg (approved):

PR #495, opened 2021-01-06
Add a sandbox tree view

tmattio:
This PR adds a sandbox view to the OCaml activity tab.

It builds on top of https://github.com/ocamllabs/vscode-ocaml-platform/pull/497 to provide a cleaner implementation of the treeview using the functorized VSCode bindings.

Where the Opam Switch view displays a list of switches on the user system and the opam packages in each of them, the Sandbox view displays the list of packages installed in the currently selected sandbox.

image

## Switch view

In addition to the Sandbox view, the following changes have been implemented on the Opam Switch view:

- The direct descendants of the switch are now the "root" packages: the packages that have been installed by the user manually.
- The packages now list their own dependencies

## Sandbox view

The sandbox view displays the package the same way as the Opam Switch one, that is, the top-level items are the root packages of the sandbox, and each package lists its own dependencies.

The view also provides the following features:

- Uninstalling a package
- Opening documentation for a package
- Upgrading the packages of the sandbox
- Installing packages (the command will open an input box to prompt the user the list of packages to install)
- Refreshing the view

## Esy support

The Sandbox view only supports Opam sandboxes for now. It'd be nice to support Esy sandboxes as well, so I implemented the boilerplate and let TODOs (`TODO: Implement Esy sandbox inspection`) in the code to identify the pieces we need to implement to support Esy.

Here are the things that need to be implemented:

- Read the list of `root` packages (the ones installed by the user directly)
- Read the list of all installed packages
- Command to upgrade package
- Command to install packages
- Command to remove packages
- For each package, read:
  - The name
  - The version
  - The documentation
  - The synopsis
  - The dependencies

cc @andreypopp @ManasJayanth - Not sure if that's something you'd be willing to contribute? If not, some pointers on how you think the above should be implemented would be greatly appreciated (especially the package discovery part, the format of the lock files and where the packages are stored is a bit confusing to me).

## Next

With this PR in, we have almost all the features we talked about regarding the UI and the interaction with Opam.

If we're happy with it, we could release it and let users provide feedback on what they would like to see.

If we're not quite satisfied with it yet, it'd be nice to identify the blockers. I have the following list of remaining features, but they don't seem critical:

- Display packages that can be upgraded
- Pin/unpin packages
- New project if empty

@rgrinberg what do you think, any blockers for a release?

cc @avsm since you were enthusiastic about having a UI for Opam in VSCode - is there anything you would have liked that is not in there?

### DIFF
--- a/src/bindings/node/node.ml
+++ b/src/bindings/node/node.ml
@@ -1,5 +1,11 @@
 open Interop
 
+let __filename () =
+  Js_of_ocaml.Js.Unsafe.eval_string "__filename" |> Js_of_ocaml.Js.to_string
+
+let __dirname () =
+  Js_of_ocaml.Js.Unsafe.eval_string "__dirname" |> Js_of_ocaml.Js.to_string
+
 module Process = struct
   val cwd : unit -> string [@@js.global "process.cwd"]
 
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -91,3 +91,5 @@ let register extension instance { id; handler } =
 
 let register_all_commands extension instance =
   List.iter ~f:(register extension instance) !commands
+
+let register id handler = command id handler |> ignore
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -92,4 +92,6 @@ let register extension instance { id; handler } =
 let register_all_commands extension instance =
   List.iter ~f:(register extension instance) !commands
 
-let register id handler = command id handler |> ignore
+let register ~id handler =
+  let (_ : command) = command id handler in
+  ()
--- a/src/extension_commands.mli
+++ b/src/extension_commands.mli
@@ -31,4 +31,4 @@ val register_all_commands :
   Vscode.ExtensionContext.t -> Extension_instance.t -> unit
 
 val register :
-  Vscode.ExtensionContext.t -> Extension_instance.t -> command -> unit
+  string -> (Extension_instance.t -> args:Ojs.t list -> unit) -> unit
--- a/src/extension_commands.mli
+++ b/src/extension_commands.mli
@@ -31,4 +31,4 @@ val register_all_commands :
   Vscode.ExtensionContext.t -> Extension_instance.t -> unit
 
 val register :
-  string -> (Extension_instance.t -> args:Ojs.t list -> unit) -> unit
+  id:string -> (Extension_instance.t -> args:Ojs.t list -> unit) -> unit
--- a/src/opam.ml
+++ b/src/opam.ml
@@ -1,115 +1,5 @@
 open Import
 
-module Switch = struct
-  type t =
-    | Local of Path.t  (** if switch name is directory name where it's stored *)
-    | Named of string  (** if switch is stored in ~/.opam *)
-
-  let of_string = function
-    | "" -> None
-    | switch_name ->
-      let switch_name = String.strip switch_name in
-      if Char.equal switch_name.[0] '/' then
-        Some (Local (Path.of_string switch_name))
-      else
-        Some (Named switch_name)
-
-  let name = function
-    | Named s -> s
-    | Local p -> Path.to_string p
-
-  let equal x y =
-    match (x, y) with
-    | Local x, Local y -> Path.equal x y
-    | Named x, Named y -> String.equal x y
-    | _, _ -> false
-end
-
-module Package = struct
-  type t =
-    { name : string
-    ; version : string
-    ; documentation : string option
-    ; synopsis : string option
-    ; path : Path.t
-    }
-
-  let name t = t.name
-
-  let version t = t.version
-
-  let documentation t = t.documentation
-
-  let synopsis t = t.synopsis
-
-  let rec documentation_of_fields = function
-    | [] -> None
-    | OpamParserTypes.Variable (_, "doc", String (_, v)) :: _ -> Some v
-    | _ :: rest -> documentation_of_fields rest
-
-  let rec synopsis_of_fields = function
-    | [] -> None
-    | OpamParserTypes.Variable (_, "synopsis", String (_, v)) :: _ -> Some v
-    | _ :: rest -> synopsis_of_fields rest
-
-  let of_path path =
-    match String.split (Path.basename path) ~on:'.' with
-    | name :: version_parts ->
-      let open Promise.Syntax in
-      let opam_filepath = Path.(path / "opam") |> Path.to_string in
-      let+ file_content = Fs.readFile opam_filepath in
-      let OpamParserTypes.{ file_contents; _ } =
-        OpamParser.FullPos.string file_content opam_filepath
-        |> OpamParser.FullPos.to_opamfile
-      in
-      let version = String.concat version_parts ~sep:"." in
-      let documentation = documentation_of_fields file_contents in
-      let synopsis = synopsis_of_fields file_contents in
-      Some { name; version; path; documentation; synopsis }
-    | _ -> Promise.return None
-
-  let sexp_of_t t =
-    Sexp.List
-      [ Sexp.Atom "name"
-      ; Sexp.Atom t.name
-      ; Sexp.Atom "version"
-      ; Sexp.Atom t.version
-      ; Sexp.Atom "documentation"
-      ; Sexp.Atom (Option.value t.documentation ~default:"")
-      ; Sexp.Atom "synopsys"
-      ; Sexp.Atom (Option.value t.synopsis ~default:"")
-      ; Sexp.Atom "path"
-      ; Sexp.Atom (t.path |> Path.to_string)
-      ]
-
-  let t_of_sexp = function
-    | Sexp.List
-        [ Sexp.Atom "name"
-        ; Sexp.Atom name
-        ; Sexp.Atom "version"
-        ; Sexp.Atom version
-        ; Sexp.Atom "documentation"
-        ; Sexp.Atom documentation
-        ; Sexp.Atom "synopsys"
-        ; Sexp.Atom synopsis
-        ; Sexp.Atom "path"
-        ; Sexp.Atom path
-        ] ->
-      { name
-      ; version
-      ; path = Path.of_string path
-      ; documentation =
-          ( match documentation with
-          | "" -> None
-          | _ -> Some documentation )
-      ; synopsis =
-          ( match synopsis with
-          | "" -> None
-          | _ -> Some synopsis )
-      }
-    | _ -> assert false
-end
-
 type t = Cmd.spawn
 
 let opam_binary = Path.of_string "opam"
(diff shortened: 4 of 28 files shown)


src/opam.mli
```diff
@@ -21,32 +44,61 @@ module Package : sig
...
+val package_dependencies :
+  Package.t -> (Package.t list, string) result Promise.t
+
+(** Uninstall a package from a switch by calling [opam uninstall]. *)
+val package_uninstall : t -> Switch.t -> Package.t -> Cmd.t
+
+(** {4 General utilities} *)
+
+(** Execute an opam sub-command with in the given switch. *)
+val exec : t -> Switch.t -> string list -> Cmd.t
```
rgrinberg, line 101: We need an `args` label on naked `string list` parameters like this one.

src/treeitem_dependency.ml
```diff
@@ -0,0 +1,180 @@
...
+    let* packages = Sandbox.root_packages sandbox in
+    match packages with
+    | Ok packages ->
+      let names =
+        packages |> List.map ~f:(fun n -> Package (n, Some (Sandbox sandbox)))
+      in
+      Promise.return (Some names)
+    | Error err ->
+      let* _ =
+        Vscode.Window.showInformationMessage
```
rgrinberg, line 147: We have `show_message` to simplify code like this.

src/treeitem_dependency.mli
```diff
@@ -0,0 +1,26 @@
+type t =
+  | Package : Sandbox.Package.t * t option -> t
+  | Sandbox : Sandbox.t -> t
```
rgrinberg, line 3: It's weird that this type is reused between the sandbox and switches treeview. Is it correct that that in the sandbox view, only `Package` can appear? If that's the case, we should really be dealing with two different types for the different views.

rgrinberg (changes requested):

src/sandbox.mli
```diff
@@ -16,6 +14,20 @@
     performance, direct user to install missing tools etc). Having a single
     [Sandbox.make()], for instance, would not make it this flexible. *)
 
+module Package : sig
+  type t = Opam of Opam.Package.t
```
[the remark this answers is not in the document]
tmattio: Nope, I made the type abstract

src/opam.ml
```diff
@@ -18,119 +62,119 @@ module Switch = struct
...
-    | [] -> None
-    | OpamParserTypes.Variable (_, "synopsis", String (_, v)) :: _ -> Some v
-    | _ :: rest -> synopsis_of_fields rest
+  type t = OpamParserTypes.opamfile_item list
+
+  let of_switch opam switch =
+    let open Promise.Result.Syntax in
+    let* path = Switch.path opam switch in
+    let switch_state_filepath =
+      Path.(path / ".opam-switch" / "switch-state") |> Path.to_string
```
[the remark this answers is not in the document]
tmattio: Done

src/opam.ml
```diff
@@ -18,119 +62,119 @@ module Switch = struct
...
+    in
+    let+ file_content =
+      Fs.readFile switch_state_filepath |> Promise.map Result.return
+    in
+    let OpamParserTypes.{ file_contents; _ } =
```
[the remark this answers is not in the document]
tmattio: Done

src/opam.ml
```diff
@@ -158,69 +211,81 @@ let switch_show ?cwd t =
...
-    let+ output = Cmd.output cmd in
-    match output with
-    | Ok s ->
-      let root = String.strip s |> Path.of_string in
-      Ok Path.(root / n)
-    | Error err -> Error err )
-  | Local p -> Promise.return (Ok Path.(p / "_opam"))
-
-let get_switch_packages t switch =
+let root_packages t switch =
```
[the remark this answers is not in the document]
tmattio: Done

src/treeitem_dependency.mli
```diff
@@ -0,0 +1,26 @@
...
+
+val label : t -> string
+
+val description : t -> string option Promise.t
+
+val tooltip : t -> string option
+
+val context_value : t -> string
+
+val icon : extension_path:string -> t -> Vscode.TreeItem.LightDarkIcon.t
```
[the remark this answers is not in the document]
tmattio: I removed the need for `extension_path` entirely by adding bindings to `__filename` in the Node bindings

src/treeitem_dependency.ml
tmattio: Done

src/opam.mli
tmattio: Done

src/treeitem_dependency.mli
tmattio: I split the implementations (actually re-implemented it with the new VSCode API), so the types are not shared anymore.

tmattio:
I rebased on top of https://github.com/ocamllabs/vscode-ocaml-platform/pull/497, the only relevant commit for review is https://github.com/ocamllabs/vscode-ocaml-platform/pull/495/commits/60c5c6d0c33964778fe136d2775e8973007ff18c.

mnxn:
@tmattio: there were a few more changes in #497 before it was merged, so I think you'll have to rebase again.

src/extension_commands.ml
```diff
@@ -91,3 +91,5 @@ let register extension instance { id; handler } =
 
 let register_all_commands extension instance =
   List.iter ~f:(register extension instance) !commands
+
+let register id handler = command id handler |> ignore
```
rgrinberg, line 95: We should specify the type of what we're ignoring.

src/extension_commands.mli
```diff
@@ -31,4 +31,4 @@ val register_all_commands :
   Vscode.ExtensionContext.t -> Extension_instance.t -> unit
 
 val register :
-  Vscode.ExtensionContext.t -> Extension_instance.t -> command -> unit
+  string -> (Extension_instance.t -> args:Ojs.t list -> unit) -> unit
```
rgrinberg, line 34: Could you add a label to the naked `string` argument?

src/opam.ml
```diff
@@ -18,119 +68,152 @@ module Switch = struct
...
+    | None ->
+      Promise.return
+        (Error "Could not get the root packaged from the switch state")
+    | Some l ->
+      Promise.List.filter_map
+        (fun pkg ->
+          let package_path =
+            (* The package path is never the root, so it's safe to use
+               [value_exn] *)
+            Path.parent (path package) |> fun x -> Option.value_exn x
```
rgrinberg, line 206: `fun x ->` is not needed here.

src/opam.ml
```diff
@@ -158,69 +250,51 @@ let switch_show ?cwd t =
...
+  match callback switch_state with
+  | Some l ->
+    let* path = Switch.path t switch in
+    Promise.List.filter_map
+      (fun name ->
+        let package_path = Opam_path.package_dir path name in
+        Package.of_path package_path)
+      l
+    |> Promise.map Result.return
+  | None ->
```
rgrinberg, line 289: `None ->` should come first.

src/opam.ml
```diff
@@ -158,69 +250,51 @@ let switch_show ?cwd t =
...
-      Path.(path / ".opam-switch" / "switch-state") |> Path.to_string
-    in
-    let+ file_content = Fs.readFile switch_state_filepath in
-    let OpamParserTypes.{ file_contents; _ } =
-      OpamParser.string file_content switch_state_filepath
-    in
-    compiler_of_fields file_contents
+  let+ switch_state = Switch_state.of_switch t switch in
+  match switch_state with
+  | Error _ -> None
```
rgrinberg, line 272: Should we log the error perhaps?

src/treeview_sandbox.ml
```diff
@@ -0,0 +1,206 @@
...
+  let tooltip t = Sandbox.Package.synopsis t
+
+  let context_value t =
+    match Sandbox.Package.documentation t with
+    | Some _ -> "package-with-doc"
+    | None -> "package"
+
+  let icon _ =
+    TreeItem.LightDarkIcon.
+      { light = `String (Node.__filename () ^ "/../../assets/number-light.svg")
```
rgrinberg, line 23: Why don't we use `extension_path`?

src/sandbox.ml
```diff
@@ -1,31 +1,68 @@
...
-   fallback - project_root is different from Package_manager root (Eg. Opam
-   (Path.of_string "/foo/bar")). Project root is the directory where manifest
-   file (opam/esy.json/package.json) was found. Package_manager root is the
-   directory that contains the manifest file responsible for setting up the
-   sandbox - the two are same for Esy and Opam project but different for
-   bucklescript. Bucklescript projects have this manifest file abstracted away
-   from the user (at least at the moment) - Manifest: abstracts functions
-   handling manifest files of the supported package managers *)
+module Package = struct
+  type t =
```
rgrinberg, line 4: This type feels wrong to me for similar reasons as some other types that were removed in this PR. The reason why it feels wrong is that for either esy or opam, we always know there's kind of package we're dealing with in the tree view. It is never the case the case we're actually dealing with a collection of mixed packages. That suggests that we should use   a form of polymorphism in the sandbox treeview that reflects this. Either functors or type indexed values seem to fit the task much better.

src/opam.ml
tmattio: Somehow, `Path.parent (path package) |> Option.value_exn` fails to compile with:

```
This expression has type
  ?here:Base__Source_code_position0.t ->
  ?error:Base.Error.t -> ?message:string -> 'a option -> 'a
but an expression was expected of type Path.t option -> 'b
```

src/treeview_sandbox.ml
tmattio: Propagating `extension_path` all the way to this function was very inconvenient, and looking at how other extensions are implemented, it seems to be the norm to use a relative path for the assets.

src/opam.ml
rgrinberg: Ah yes, I frogot that we use base. I'd suggest you write it as `let x = Option.value_exn (Path.parent (path package))`. `let` bindings are made for this kind of thing.

src/treeview_sandbox.ml
rgrinberg: Sure, we can use this approach instead. But then we should use it consistently everywhere. I'd also suggest to create a helper function `val asset : fname:string -> string` to encapsulate this detail.

src/sandbox.ml
tmattio: I agree the type does not modelize reality perfectly, however, I don't think it's a huge problem: the variants are not exposed in the interface, and the only way to get a list of packages is through `Sandbox.packages` and `Sandbox.root_packages`. 

FWIW, I tried removing the `Package` module and use functors in `Treeview_sandbox`. The problem is that we need the sandbox to instantiate the functor, and we have to get the sandbox in `getItems`, otherwise refreshing the treeview will not work as the function will use an old sandbox. Instantiating the functor in `getItems` is not possible, however, as the type of the functor instance would leak from the scope of the function (it won't compile).

In any case, having a `Package` module in `Sandbox` was intentional (a sandbox package that abstracts the package manager seemed like a good design), so I tried to use type-indexed value, as you suggested, with the following representation:

```ocaml
module Impl = struct
  type 'a t =
    { name : 'a -> string
    ; version : 'a -> string
    ; documentation : 'a -> string option
    ; synopsis : 'a -> string option
    ; has_dependencies : 'a -> bool
    ; dependencies : 'a -> ('a list, string) result Promise.t
    ; remove : string list -> (unit, string) result Promise.t
    ; install : string list -> (unit, string) result Promise.t
    ; upgrade : unit -> (unit, string) result Promise.t
    ; root_packages : unit -> ('a list, string) result Promise.t
    }
end

type 'a t =
  { pkg : 'a
  ; impl : 'a Impl.t
  }
```

But then functions such as `val packages : t -> (Package.t list, string) result Promise.t` can't be implemented.

Considering that the possibility of having a mixed list of packages is avoided by not exposing the type, I am not sure exploring other potential solutions is really worth it. Happy to try other solutions you might think of if you disagree though 🙂

rgrinberg: > But then functions such as val packages : t -> (Package.t list, string) result Promise.t can't be implemented.

The signature would need to be `val packages : 'a t -> ('a list, string) result Promise.t`. We'd have to push the type indexed value further.

Alright, then let's leave it as is for now. Just add a comment in the type describing the current caveats. We'll try to fix it another time.

src/treeview_sandbox.ml
tmattio: Done, I added an `asset` function in `Path`

src/sandbox.ml
tmattio: Done, I also pushed my explorations here:

- Functorized tree view: https://github.com/tmattio/vscode-ocaml-platform/commit/8011335d3c79b7a267122e7d62fe0dd13ae5bb67
- type-indexed values for sandbox packages: https://github.com/tmattio/vscode-ocaml-platform/commit/11c1dcdb36080427cb9b81c618ddc35e759298c3

rgrinberg (approved):

src/bindings/node/node.ml
```diff
@@ -1,5 +1,11 @@
 open Interop
 
+let __filename () =
+  Js_of_ocaml.Js.Unsafe.eval_string "__filename" |> Js_of_ocaml.Js.to_string
+
+let __dirname () =
+  Js_of_ocaml.Js.Unsafe.eval_string "__dirname" |> Js_of_ocaml.Js.to_string
+
```
mnxn, line 8: Can these be written with gen_js_api?
```ocaml
val __filename : unit -> string [@@js.global "__filename"]
```

Or do the `__` variables have special behavior that requires an eval?

tmattio: This fails with `__filename is not a function`. I guess `val __filename : string [@@js.global "__filename"]` would work, but this would make the variable static. Jsoo compiles to a single JS file, so not a problem in practice, but still, does not seem like a good idea.

mnxn: My bad, I meant `[@@js.get "__filename"]`

"Global getter" from https://github.com/LexiFi/gen_js_api/blob/master/VALUES.md

tmattio: This one fails with `cannot read property length of undefined` 😕

mnxn: Strange. I guess it's some weird node.js magic then.

Leaving the eval is fine in that case.

mnxn (approved):

tmattio:
@rgrinberg Now that this is merged, do you see any blocker for a release?

rgrinberg:
I don't see any blockers. Seems like a good time to make a release.

PR #497, opened 2021-01-13
Refactor vscode bindings with functors

mnxn:
Overview:

- Use module functors to model generic vscode types.
Previously generic types would be specialized to a specific type or implemented in a hacky way (like the event type).

- Use first class modules for generic functions.
Generic functions need a way to access the `t_of_js`/`t_to_js` functions to properly convert to/from JS types. This interacts well with the functorized types..
- Use generative functors for individual interface/class types.
Reduces  a bit of boilerplate and makes whether a type is an interface or class part of the code instead of a comment.

- Introduce `Js.T` module type.
Useful for functor argument type and reduces repetition of `type t`/`val t_of_js`/`val t_to_js` in the mli files.

- Fix various bindings.
The lack of proper generic types (until now) prevented certain functions from being implemented correctly or at all.


src/treeview_switches.ml
```diff
@@ -240,17 +244,21 @@ let register extension =
...
     in
-    let getTreeItem ~element = Promise.return element in
-    let event_emitter = Vscode.EventEmitter.make () in
-    let event = Vscode.EventEmitter.event event_emitter in
+    let getTreeItem ~element = `Value element in
+    let module EventEmitter =
+      Vscode.EventEmitter.Make (Interop.Js.Or_undefined (TreeItem)) in
+    let event_emitter = EventEmitter.make () in
+    let event = EventEmitter.event event_emitter in
+    let module TreeDataProvider = Vscode.TreeDataProvider.Make (Vscode.TreeItem) in
```
tmattio, line 252: This is great, once this PR is in, I'll refactor the `Treeview_switches` module to use a custom module instead of `TreeItem`. At the moment, we serialize the `Dependency.t` as a string in the ID of the `TreeItem` and get it back in `getChildren`, this is a really unwelcomed hack, so can't wait to get rid of it 🙌

tmattio (approved):
LGTM. I'm a bit wary of using so many first-class modules, it makes the API less intuitive IMO, but it seems to be solving a ton of limitations, especially for the Tree View part of the API, so I won't complain 🙂

tmattio:
I have rebased my work on the UI to use this PR and it's indeed much cleaner.

@rgrinberg Do you have any concerns about this PR? I'll wait for it to be merged to submit the work I've been doing on the UI

vscode/vscode/vscode.mli
```diff
@@ -674,78 +582,73 @@ module ConfigurationTarget : sig
...
+    ; workspaceFolderValue : 'a option
+    ; defaultLanguageValue : 'a option
+    ; globalLanguageValue : 'a option
+    ; workspaceLanguageValue : 'a option
+    ; workspaceFolderLanguageValue : 'a option
+    ; languageIds : string list option
     }
 
-  val get : t -> section:string -> unit -> Jsonoo.t option
+  val get : (module Js.T with type t = 'a) -> t -> section:string -> 'a option
```
rgrinberg, line 604: Is this ever called with anything other than `module Jsonoo`?

vscode/vscode/vscode.ml
```diff
@@ -2185,16 +2370,16 @@ module Window = struct
...
-    -> task:
-         (progress:Progress.t -> token:CancellationToken.t -> Ojs.t Promise.t)
-    -> Ojs.t Promise.t
+  val withProgress : options:ProgressOptions.t -> task:Ojs.t -> Ojs.t
     [@@js.global "vscode.window.withProgress"]
 
-  let withProgress ~options ~task =
-    let task ~progress ~token = Obj.magic (task ~progress ~token) in
-    Obj.magic (withProgress ~options ~task)
+  let withProgress (type a) (module T : Js.T with type t = a) ~options ~task :
```
rgrinberg, line 2376: I might be not understanding the point of this, but why do we need to pass this `T` module`? we aren't doing anything with its fields from what I can see.

mnxn: It's used in a gen_js_api extension for converting `T.t Promise.t` types in the function body.

vscode/vscode/vscode.mli
mnxn: I don't think so.  I'd like to leave this function as-is to be faithful to the VSCode API. 

I just added `get_json` and `inspect_json` functions for easier access to the JSON output.

rgrinberg: It's OK, the specializtion to json aren't really necessary. I was just thinking if `Ojs.t` should have been used instead. I suppose `get_default` becomes easier to use if we pass the conversion functions.

vscode/vscode/vscode.mli
```diff
@@ -2118,26 +1913,36 @@ module Window : sig
...
        command:string
     -> callback:
          (   textEditor:TextEditor.t
           -> edit:TextEditorEdit.t
-          -> args:Ojs.t list
+          -> args:Js.Any.t list
           -> unit)
     -> Disposable.t
 
   val executeCommand :
```
rgrinberg, line 1941: I don't think you should use pass a module in situations where it doesn't add any type safety. In this situation, returning `Ojs.t` seems to be just as good, and simpler to boot.

vscode/vscode/vscode.mli
```diff
@@ -674,78 +582,77 @@ module ConfigurationTarget : sig
...
+    ; languageIds : string list option
     }
 
-  val get : t -> section:string -> unit -> Jsonoo.t option
+  val get : (module Js.T with type t = 'a) -> t -> section:string -> 'a option
+
+  val get_json : t -> section:string -> Jsonoo.t option
+
+  val get_default :
+       (module Js.T with type t = 'a)
```
rgrinberg, line 609: It would be nice to introduce an alias like this:

```
  type 'a js = (module T with type t = 'a)
```

To avoid having to type the module all the time.

By the way, did you consider just using a record instead:
```ocaml
type 'a js =
  { from : Ojs.t -> 'a
  ; to_ : 'a -> Ojs.t
  }
```

I don't think it's all that important, but records are generally much more convenient to work with.

vscode/vscode/vscode.mli
```diff
@@ -792,82 +699,71 @@ module StatusBarItem : sig
...
+    -> unit
+    -> Disposable.t
 
-  val t_to_js : ('a -> Ojs.t) -> 'a t -> Ojs.t
+  module Make (T : Js.T) : Js.T with type t = T.t t
 end
 
 module EventEmitter : sig
-  type t
+  type 'a t
```
rgrinberg, line 737: It's a bit of a shame that we're losing polymorphism here. Now we can't have a single `EventEmitter.fire` that will work for all event emitters. We could mend by just putting the dictionary inside the event emitter.

```ocaml
type 'a t
val make : (module Js.T with type t = 'a)-> 'a t
```

It's a little more code in the binding, but at least people don't need to instantiate modules to use the API. This is how various data structures in base work. What do you think?

vscode/vscode/vscode.mli
mnxn: I added the alias.

I considered doing it with records, but I think it's easier with first class modules because all of the modules in `vscode.ml` already conform to `Js.T` module type. It's as simple as just packing the module into a value. Making it a record requires extra code without much benefit in my opinion.

Using first class modules also adds some symmetry between generic types (applying a module to a functor) and generic functions (applying a first class module to a function).

rgrinberg: Yeah, that sounds good. First class modules become more of a pain when you need lightweight combinators like `val iso : 'a js -> ('a -> 'b) -> ('b -> 'a) -> 'b js.t`, but we have no use for them yet.

vscode/vscode/vscode.mli
mnxn: I feel that passing around the `to_js`/`of_js` values inside of `'a t` would add significant overhead, both to the runtime and code necessary to manipulate such values.

The benefit of functors is that once they're instantiated, they work like bindings to normal types. Inside the functor body, bindings can be written with gen_js_api that automatically use `T.t_to_js`/`T.t_of_js`.

Using a custom representation that bundles those functions would require each method of the generic type to do the conversions, which would get verbose very quickly.

rgrinberg: It's true that it's a little slower, but I doubt that it will make much difference for us. I've used this method gainfully in native ocaml, and I'm pretty certain js optimizes such code even better.

The point about boilerplate is more important. It's true that it adds more unpacking, but this overhead will become meaningless compared to the overhead of multiple modules for every type.

For now we only have a few callers. Let's leave things as is

rgrinberg (approved):

PR #498, opened 2021-01-13
Fix switch impl intf twice

ulugbekna:
* Fixes https://github.com/ocaml/ocaml-lsp/issues/358 

The PR provides a _quick_ fix. If the new mli file isn't empty, we don't insert inferred interface. 
Not sure whether it is better to replace the contents with the (potentially new) interface or maybe show a diff window with new interface. 

* Adds more structured support for URI schemes
* Improves 'Switch_impl_intf.show_file' by refactoring, renaming, and removing redundant value returned 

---

Reviewing commit by commit should be easiest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,15 @@
 # Change Log
 
-# Unreleased
+## Unreleased
 
 - Highlight token aliases in Menhir associativity declarations (#473)
+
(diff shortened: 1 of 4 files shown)


tmattio (approved):

rgrinberg:
Looks ok to me. I didn't go over the CI failure, is that expected?

PR #499, opened 2021-01-13
New command: open "current" 'dune' file 

ulugbekna:
The PR adds a command to open the dune file located in the same folder the currently open file. The command is available in the command palette iff an ocaml-related (ocaml, reason, etc.) file is open. 

If the dune file is missing, a new dune file is opened in a draft (non-saved) mode. 

The PR is based on #498 

The dune rpc is coming and can replace the implementation, but the code innocuously small, so it shouldn't be a big deal. 

Without this command, I do `cmd+p`, type 'dune' and have to remember which library I'm in to disambiguate among many other dune files 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -183,7 +183,7 @@
       {
         "command": "dune.current-dune-file",
         "category": "Dune",
-        "title": "Open current dune file"
+        "title": "Open Dune File (located in the same folder)"
       }
     ],
     "keybindings": [
@@ -213,7 +213,7 @@
       "commandPalette": [
         {
           "command": "dune.current-dune-file",
-          "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
+          "when": "editorIsOpen"
         },
         {
           "command": "ocaml.refresh-switches",
--- a/package.json
+++ b/package.json
@@ -207,21 +207,6 @@
     "menus": {
       "commandPalette": [
         {
-          "command": "ocaml.select-sandbox"
-        },
-        {
-          "command": "ocaml.server.restart"
-        },
-        {
-          "command": "ocaml.open-terminal"
-        },
-        {
-          "command": "ocaml.open-terminal-select"
-        },
-        {
-          "command": "ocaml.switch-impl-intf"
-        },
-        {
           "command": "ocaml.refresh-switches",
           "when": "false"
         },
--- a/package.json
+++ b/package.json
@@ -179,6 +179,11 @@
           "light": "assets/document-search-light.svg",
           "dark": "assets/document-search-dark.svg"
         }
+      },
+      {
+        "command": "dune.current-dune-file",
+        "category": "Dune",
+        "title": "Open current dune file"
       }
     ],
     "keybindings": [
@@ -206,6 +211,10 @@
     ],
     "menus": {
       "commandPalette": [
+        {
+          "command": "dune.current-dune-file",
+          "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
+        },
         {
           "command": "ocaml.refresh-switches",
           "when": "false"
--- a/package.json
+++ b/package.json
@@ -181,8 +181,8 @@
         }
       },
       {
-        "command": "dune.current-dune-file",
-        "category": "Dune",
+        "command": "ocaml.current-dune-file",
+        "category": "OCaml",
         "title": "Open Dune File (located in the same folder)"
       }
     ],
@@ -212,7 +212,7 @@
     "menus": {
       "commandPalette": [
         {
-          "command": "dune.current-dune-file",
+          "command": "ocaml.current-dune-file",
           "when": "editorIsOpen"
         },
         {
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -90,8 +90,8 @@ module Dune_commands = struct
       match Vscode.Window.activeTextEditor () with
       | None ->
         assert false
-        (* this command is available in the command palette only when a
-           ocaml/reason/mehnir/ocamllex files are open *)
+        (* this command is available (in the command) palette only when a file
+           is open *)
       | Some text_editor ->
         let doc = TextEditor.document text_editor in
         let uri = TextDocument.uri doc in
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -96,11 +96,14 @@ module Dune_commands = struct
         let doc = TextEditor.document text_editor in
         let uri = TextDocument.uri doc in
         let dune_file_uri =
-          Uri.fsPath uri |> Path.of_string |> fun path ->
-          Path.relative path "../dune" |> Path.to_string |> Uri.file
-          |> fun uri -> Uri.toString uri ()
+          let path = Uri.fsPath uri |> Path.of_string in
+          let uri =
+            Path.relative path "../dune" |> Path.to_string |> Uri.file
+          in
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -152,19 +157,8 @@
...
-        {
-          "command": "ocaml.open-terminal"
-        },
-        {
-          "command": "ocaml.open-terminal-select"
-        },
-        {
-          "command": "ocaml.switch-impl-intf"
+          "command": "dune.current-dune-file",
+          "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
```
rgrinberg, line 161: Why is this command disabled for non source files? It should work in any context.

src/extension_commands.ml
```diff
@@ -84,6 +84,27 @@ let switch_impl_intf =
...
+      match Vscode.Window.activeTextEditor () with
+      | None ->
+        assert false
+        (* this command is available in the command palette only when a
+           ocaml/reason/mehnir/ocamllex files are open *)
+      | Some text_editor ->
+        let doc = TextEditor.document text_editor in
+        let uri = TextDocument.uri doc in
+        let dune_file_uri =
+          Uri.fsPath uri |> Path.of_string |> fun path ->
```
rgrinberg, line 99: Could you use let bindings here? It's weird to see `let x = foo` written as `foo |> fun x ->`.

src/extension_commands.ml
```diff
@@ -84,6 +84,27 @@ let switch_impl_intf =
...
+          Path.relative path "../dune" |> Path.to_string |> Uri.file
+          |> fun uri -> Uri.toString uri ()
```
rgrinberg, line 101: and here

src/extension_commands.ml
```diff
@@ -84,6 +84,27 @@ let switch_impl_intf =
...
+        in
+        open_file_in_text_editor dune_file_uri |> ignore
```
rgrinberg, line 103: Whenever we ignore something, we should specify its type.

src/extension_commands.ml
```diff
@@ -84,6 +84,27 @@ let switch_impl_intf =
...
   command Extension_consts.Commands.switch_impl_intf handler
 
+module Dune_commands = struct
+  let open_current_dune_file =
+    let handler (_instance : Extension_instance.t) ~args:_ =
+      match Vscode.Window.activeTextEditor () with
+      | None ->
+        assert false
+        (* this command is available in the command palette only when a
+           ocaml/reason/mehnir/ocamllex files are open *)
```
rgrinberg, line 94: It should really be available when any source file is open.

package.json
ulugbekna: The command says 'open "current" dune file', where "current" is understood as the dune component (lib/exec) that the source file is a part of. Q: What would you expect "current dune file" be for, say, a `package.json` in the project root? I was trying to be strict here to not confuse users.
---
All in all, I've been using this command for some time and find it super useful! It could, however, be improved by jumping directly to the dune stanza/component, which the source is a component of.

rgrinberg: >  Q: What would you expect "current dune file" be for, say, a package.json in the project root? I was trying to be strict here to not confuse users.

If there's a dune file in the same dir as `package.json`, I would expect it to jump there. Note that one could have a dune file for C, C++, Coq, cram, and all kinds of other source files. So if anything, sticking to a subset is equally confusing.

> It could, however, be improved by jumping directly to the dune stanza/component, which the source is a component of.

Yes, that would be nicer. Especially given that the source file might not belong to any dune file and the command would still jump to a dune file.

ulugbekna: Will update to account for any open file then.

ulugbekna:
This is ready btw. Would be nice to merge before there are more merge conflicts.

tmattio (approved):

package.json
```diff
@@ -179,6 +179,11 @@
           "light": "assets/document-search-light.svg",
           "dark": "assets/document-search-dark.svg"
         }
+      },
+      {
+        "command": "dune.current-dune-file",
+        "category": "Dune",
+        "title": "Open Dune File (located in the same folder)"
```
mnxn, line 186: I think we should keep the command IDs under a consistent organization.

This should be `ocaml.current-dune-file` under the OCaml category so it's as discoverable as the other commands. The "Dune" keyword is already present in the command title, so the other parts are redundant.

tmattio: Oops, oversight on my part, I agree with @mnxn

src/extension_commands.ml
```diff
@@ -84,6 +84,30 @@ let switch_impl_intf =
...
+        let uri = TextDocument.uri doc in
+        let dune_file_uri =
+          let path = Uri.fsPath uri |> Path.of_string in
+          let uri =
+            Path.relative path "../dune" |> Path.to_string |> Uri.file
+          in
+          Uri.toString uri ()
+        in
+        (open_file_in_text_editor dune_file_uri : TextEditor.t Promise.t)
+        |> ignore
```
mnxn, line 106: I'd prefer to use an explicit let binding here:
```ocaml
let (_ : TextEditor.t Promise.t) = ... in ()
```

rgrinberg (approved):

src/extension_consts.ml
```diff
@@ -26,6 +28,10 @@ module Commands = struct
   let open_switches_documentation = ocaml_prefixed "open-switches-documentation"
 
   let open_sandbox_documentation = ocaml_prefixed "open-sandbox-documentation"
+
+  module Dune = struct
+    let open_current_dune_file = dune_prefixed "current-dune-file"
+  end
```
mnxn, line 34: This code still has to be updated to accommodate the change in command ID.

ulugbekna: Fixed, thanks!

src/extension_commands.mli
```diff
@@ -24,6 +24,10 @@ val open_terminal : command
 
 val switch_impl_intf : command
 
+module Dune_commands : sig
+  val open_current_dune_file : command
+end
+
```
mnxn, line 30: Having a separate module is unnecessary now. This can be flattened so `open_current_dune_file` is on the same level as the other commands.

src/extension_consts.ml
```diff
@@ -26,6 +26,10 @@ module Commands = struct
   let open_switches_documentation = ocaml_prefixed "open-switches-documentation"
 
   let open_sandbox_documentation = ocaml_prefixed "open-sandbox-documentation"
+
+  module Dune = struct
+    let open_current_dune_file = ocaml_prefixed "current-dune-file"
+  end
```
mnxn, line 32: Wrapping this in a module is unnecessary too.

mnxn (changes requested):
The command should be documented in the README for users.

src/extension_commands.ml
ulugbekna: The current code seems terser (2 vs 4 lines) and slightly easier to comprehend (we don't do anything with the returned promise, so we ignore it explicitly) IMHO. Feel free to change it if you feel strongly about it.

src/extension_commands.mli
ulugbekna: I want the `Dune` and `Dune_commands` modules to separate somehow dune-related constants and commands.

src/extension_commands.ml
mnxn: We use let bindings to ignore values in the rest of the codebase, so I like to stay consistent. I pushed the change.

src/extension_commands.mli
mnxn: I see your point, but can't we introduce separation later once it's necessary? A module with only one value is just adding noise at this point.

mnxn (approved):

ulugbekna:
Thanks, Max!

PR #500, opened 2021-01-18
Remove BuckleScript support instructions

tmattio:
ReScript now has its own extension, so we can safely say we don't support it and redirect to the new extension instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,40 @@ the `usr/local/bin` folder relative to the installation directory).
 
 ### ReScript / BuckleScript
 
-ReScript (previously BuckleScript) is not supported, use [rescript-vscode](https://github.com/rescript-lang/rescript-vscode) instead.
+The new ReScript syntax (`res` and `resi` files) is not supported, you should
+use [rescript-vscode](https://github.com/rescript-lang/rescript-vscode) instead.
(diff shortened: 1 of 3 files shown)


mnxn:
rescript-vscode doesn't support ocaml or reason syntax so I think we should leave these instruction up for now. At least until ocaml-lsp drops support for old ocaml versions and bucklescript.

mnxn (approved):

tmattio:
> rescript-vscode doesn't support ocaml or reason

Wasn't aware of that! I updated the instructions to recommend the ReScript extension if people want to use the new syntax, hopefully, it'll make it clearer when to chose which extension.

README.md
```diff
@@ -41,10 +47,10 @@ npm install esy --global
 ```json
 {
   "dependencies": {
-    "@opam/ocaml-lsp-server": "1.1.0",
+    "@opam/ocaml-lsp-server": "*",
     "@opam/ocamlfind-secondary": "*",
     "@opam/reason": "*",
-    "ocaml": "4.6.x"
+    "ocaml": "4.11.x"
```
Khady, line 53: for bucklescript you probably want to keep 4.6?

PR #501, opened 2021-01-19
Use opam lock in CI cache

mnxn:
Similar to https://github.com/ocaml/ocaml-lsp/pull/318, this PR uses `opam lock` to create lockfiles which are used for restoring the opam cache. This should significantly speed up the CI.

Also updates and fixes typos in opam author list.


smorimoto:
On Windows, it's meaningless and should be disabled. As another option, I think we can try actions-ml on this project.
https://github.com/actions-ml
I'm not very positive about dependency caching, and just using the compiler cache can reduce setup time from 8 minutes to 1 minute.

mnxn:
> On Windows, it's meaningless and should be disabled.

Why is it meaningless on Windows?

smorimoto:
Oops, I missed `steps.opam-root-dir-path.outputs.dir`!

mnxn:
It seems the Windows build fails at uploading the opam cache because it's using cygwin's tar, which expect cygwin paths:

```
Post job cleanup.
c:\cygwin\bin\tar.exe --posix --use-compress-program "zstd -T0" -cf cache.tzst -P -C D:/a/vscode-ocaml-platform/vscode-ocaml-platform --files-from manifest.txt --force-local
/usr/bin/tar: C\:\\cygwin\\home\runneradmin\\.opam: Cannot stat: No such file or directory
/usr/bin/tar: Exiting with failure status due to previous errors
Warning: Tar failed with error: The process 'c:\cygwin\bin\tar.exe' failed with exit code 2
```

Same problem with the esy cache that I didn't notice until now:

```
Post job cleanup.
c:\cygwin\bin\tar.exe --posix --use-compress-program "zstd -T0" -cf cache.tzst -P -C D:/a/vscode-ocaml-platform/vscode-ocaml-platform --files-from manifest.txt --force-local
/usr/bin/tar: C\:\\Users\runneradmin\\.esy\003___________________________________________________________\\i: Cannot stat: No such file or directory
/usr/bin/tar: Exiting with failure status due to previous errors
Warning: Tar failed with error: The process 'c:\cygwin\bin\tar.exe' failed with exit code 2
```

smorimoto:
You should restore the cache before you run setup-ocaml. Also if it hits the cache, you should run `opam upgrade --fixup` to check the consistency of installed opam packages.

mnxn:
I don't know if adding MSYS2 to PATH fixed the tar issue, but I'm getting a different error now:

```
Unable to reserve cache with key esy-windows-latest-e240c2aa53c9875242687e3f17b46872e850b0a18ad1aac5cdf0415476445acd, another job may be creating this cache.
```

```
Unable to reserve cache with key opam-windows-latest-4.11.1-9865fab9e2d488ce7031cc5e3fc1dcacd393f563bd5eda7e6e426c55d9096146, another job may be creating this cache.
```

I'm not sure what to make of this.

smorimoto:
In general, managing the build cache is often complex, and this is a common way to avoid this problem:
https://circleci.com/docs/2.0/caching/#managing-caches

mnxn:
Windows has been giving me too much trouble. The GNU tar from MSYS2 had the same error as the tar from Cygwin.
I don't want to spend any more time messing with the PATH so I've just disabled the opam cache on Windows.

The caching is still useful, even if it only for macOS and ubuntu.

tmattio (approved):

smorimoto:
Given that the version of the opam package changes gradually, using opam lock to cache the entire switch does not seem to be a good idea. actions-ml is trying to speed it up with a different approach and I think it's worth trying.

mnxn:
actions-ml does look interesting. Would you like to open a different PR?

smorimoto:
Yes, I will. The built-in dune cache support is really interesting. In addition, it provides a cache of variants such as flambda.

mnxn:
Closed in favor of https://github.com/ocamllabs/vscode-ocaml-platform/pull/510

Issue #490, opened 2020-12-28
Running the toplevel phrase surrounding the text cursor (à la tuareg C-c C-e)


ysalmon:
I am just discovering VSCode/VSCodium with the OCaml extension. It has many nice features compared to the rather austere Emacs+Tuareg I am used to.

I miss one thing however : the ability to run the toplevel phrase surrounding the text cursor with utop (or the standard OCaml REPL).
This would be especially useful in my context, which is teaching introductory OCaml to students with extremely scarce lab sessions : we do not write complete programs but rather a small bunch of functions to implement (part of) an algorithm.

Currently, I can use the Run selected text from the Terminal menu, after having run utop. But this is rather crude as it does not add the ``;;`` that utop toplevel requires at the end. Also, one has to manually select the whole phrase.

Is it possible to use the OCaml extension to do so ?
I think the required procedures already exist :

- in a .ml file without any ``;;``, I get correct type annotations for each function and
- the breadcrumbs display the name of the function or type definition in which the text cursor currently is, so it has a way to associate a cursor position to the surrounding phrase.

tmattio:
Hey @ysalmon,

Thanks for the issue. Sending top-level expressions to Utop is in our list and the feature is described here https://github.com/ocamllabs/vscode-ocaml-platform/issues/451.

We're currently working on the UI and Dune integration, but interactivity will probably come soon after these.

FayardProf:
Hi @tmattio,

I think that what would be useful to us is the "Terminal oriented" version of #451.

Thanks for your work and bonnes fêtes de fin d'année,
François

ulugbekna:
#452 considers integration with `utop` - a feature, which I also miss dearly.

tmattio:
Closing in favor of #452, the PR linked here should be merged soon anyways.

Issue #452, opened 2020-11-25
Integration with utop


ulugbekna:
We could have a command that suggests (maybe using QuickPick menu) to run `dune utop` with one of the available libraries in the currently selected sandbox/toolchain.

It would be nice to have the quickpick items sorted by the most recently used. 

Instead of "open terminal" > run "dune utop <lib name>", user would call menu with libs available and pick one. 

Integration with dune may be necessary, so a part of this feature could be useful for running dune tasks.

PR #504, opened 2021-01-20
Add Utop integration

tmattio:
This PR adds a code evaluation feature that sends selected expressions to the REPL when pressing `Shift+Enter` (or selecting "Evaluate Expression" in the editor menu).

The REPL is provided with Utop and is instantiated using `dune utop`, so this feature depends on both dune and utop to be available in the user's sandbox.

Closes https://github.com/ocamllabs/vscode-ocaml-platform/issues/452
Closes https://github.com/ocamllabs/vscode-ocaml-platform/issues/490

**TODO**

- [X] Fix code not being evaluated the first time when utop takes too long to be initialized
- [ ] Add popup to install utop if not available in the sandbox (most likely depends on #495 which adds Opam package installation capabilities)
- [x] Document the commands in the readme

### DIFF
--- a/package.json
+++ b/package.json
@@ -419,6 +419,7 @@
           "items": "string",
           "default": null
         }
+      }
     },
     "configurationDefaults": {
       "[ocaml]": {
--- a/package.json
+++ b/package.json
@@ -184,13 +184,23 @@
         "command": "ocaml.current-dune-file",
         "category": "OCaml",
         "title": "Open Dune File (located in the same folder)"
+      },
+      {
+        "command": "ocaml.open-repl",
+        "category": "OCaml",
+        "title": "Open REPL"
+      },
+      {
+        "command": "ocaml.evaluate-selection",
+        "category": "OCaml",
+        "title": "Evaluate Selection"
       }
     ],
     "keybindings": [
       {
         "command": "ocaml.switch-impl-intf",
         "key": "Alt+O",
-        "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
+        "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
       },
       {
         "command": "editor.action.codeAction",
@@ -198,7 +208,7 @@
         "args": {
           "kind": "destruct"
         },
-        "when": "editorLangId  == ocaml || editorLangId == reason"
+        "when": "editorLangId == ocaml || editorLangId == reason"
       },
       {
         "command": "editor.action.codeAction",
@@ -206,10 +216,22 @@
         "args": {
           "kind": "inferred_intf"
         },
-        "when": "editorLangId  == ocaml.interface || editorLangId == reason"
+        "when": "editorLangId == ocaml.interface || editorLangId == reason"
+      },
+      {
+        "command": "ocaml.evaluate-selection",
+        "key": "Shift+Enter",
+        "when": "editorTextFocus && editorLangId == ocaml || editorTextFocus && editorLangId == ocaml.interface || editorTextFocus && editorLangId == reason || editorTextFocus && editorLangId == ocaml.ocamllex || editorTextFocus && editorLangId == ocaml.menhir"
       }
     ],
     "menus": {
+      "editor/context": [
+        {
+          "command": "ocaml.evaluate-selection",
+          "group": "OCaml",
+          "when": "editorTextFocus && editorLangId == ocaml || editorTextFocus && editorLangId == ocaml.interface || editorTextFocus && editorLangId == reason || editorTextFocus && editorLangId == ocaml.ocamllex || editorTextFocus && editorLangId == ocaml.menhir"
+        }
+      ],
       "commandPalette": [
         {
           "command": "ocaml.current-dune-file",
@@ -252,7 +274,7 @@
         {
           "command": "ocaml.switch-impl-intf",
           "key": "Alt+O",
-          "when": "editorLangId  == ocaml || editorLangId  == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir",
+          "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir",
           "group": "navigation"
         }
       ],
@@ -379,8 +401,24 @@
           ],
           "items": "string",
           "default": null
+        },
+        "ocaml.repl.path": {
+          "description": "The path of the REPL that the extension uses",
+          "type": [
+            "string",
+            "null"
+          ],
+          "default": null
+        },
+        "ocaml.repl.args": {
+          "description": "The REPL arguments that the extension uses",
+          "type": [
+            "array",
+            "null"
+          ],
+          "items": "string",
+          "default": null
         }
-      }
     },
     "configurationDefaults": {
       "[ocaml]": {
--- a/src/bindings/node/node.ml
+++ b/src/bindings/node/node.ml
@@ -7,7 +7,7 @@ let __dirname () =
   Js_of_ocaml.Js.Unsafe.eval_string "__dirname" |> Js_of_ocaml.Js.to_string
 
 module Timeout = struct
-  type t = private Ojs.t [@@js]
+  include Class.Make ()
 
   val hasRef : t -> bool [@@js.get]
 
@@ -27,12 +27,16 @@ module Process = struct
 
   val platform : string [@@js.global "process.platform"]
 
+  val arch : string [@@js.global "process.arch"]
+
   module Env = struct
     val env : Ojs.t [@@js.global "process.env"]
 
     let get k = [%js.to: string or_undefined] (Ojs.get env k)
 
     let set k v = Ojs.set env k ([%js.of: string] v)
+
+    val env : string Interop.Dict.t [@@js.global "process.env"]
   end
 end
 
--- a/src/bindings/node/node.ml
+++ b/src/bindings/node/node.ml
@@ -6,6 +6,22 @@ let __filename () =
 let __dirname () =
   Js_of_ocaml.Js.Unsafe.eval_string "__dirname" |> Js_of_ocaml.Js.to_string
 
+module Timeout = struct
+  type t = private Ojs.t [@@js]
+
+  val hasRef : t -> bool [@@js.get]
+
+  val ref : t -> t [@@js.get]
+
+  val refresh : t -> t [@@js.get]
+
+  val unref : t -> t [@@js.get]
+end
+
+val setInterval : (unit -> unit) -> int -> Timeout.t [@@js.global]
+
+val setTimeout : (unit -> unit) -> int -> Timeout.t [@@js.global]
+
 module Process = struct
   val cwd : unit -> string [@@js.global "process.cwd"]
 
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -1,6 +1,6 @@
 open Import
 
-module ReplPath = struct
+module Repl_path = struct
   type t = string option
 
   let of_json json =
@@ -16,7 +16,7 @@ module ReplPath = struct
   let t = Settings.create ~scope:Global ~key ~of_json ~to_json
 end
 
-module ReplArgs = struct
+module Repl_args = struct
   type t = string list option
 
   let of_json json =
@@ -33,10 +33,10 @@ module ReplArgs = struct
 end
 
 let get_repl_path () =
-  Option.join (Settings.get ~section:"ocaml.repl" ReplPath.t)
+  Option.join (Settings.get ~section:"ocaml.repl" Repl_path.t)
 
 let get_repl_args () =
-  Option.join (Settings.get ~section:"ocaml.repl" ReplArgs.t)
+  Option.join (Settings.get ~section:"ocaml.repl" Repl_args.t)
 
 let name = "REPL"
 
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -1,0 +1,191 @@
+open Import
+
+module ReplPath = struct
+  type t = string option
+
+  let of_json json =
+    let open Jsonoo.Decode in
+    nullable string json
+
+  let to_json (t : t) =
+    let open Jsonoo.Encode in
+    nullable string t
+
+  let key = "path"
+
+  let t = Settings.create ~scope:Global ~key ~of_json ~to_json
+end
+
+module ReplArgs = struct
+  type t = string list option
+
+  let of_json json =
+    let open Jsonoo.Decode in
+    nullable (list string) json
+
+  let to_json (t : t) =
+    let open Jsonoo.Encode in
+    nullable (list string) t
+
+  let key = "args"
+
+  let t = Settings.create ~scope:Global ~key ~of_json ~to_json
+end
+
+let get_repl_path () =
+  Option.join (Settings.get ~section:"ocaml.repl" ReplPath.t)
+
+let get_repl_args () =
+  Option.join (Settings.get ~section:"ocaml.repl" ReplArgs.t)
+
+let name = "REPL"
+
+let has_utop sandbox =
+  let open Promise.Syntax in
+  let cmd = Sandbox.get_command sandbox "utop" [ "-version" ] in
+  let+ result = Cmd.output cmd in
+  match result with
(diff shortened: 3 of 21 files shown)


src/repl.ml
```diff
@@ -0,0 +1,96 @@
+open Import
+
+let name = "REPL"
+
+let t = ref None
```
rgrinberg, line 5: Is there a way to make this part of the instance state?

src/repl.ml
```diff
@@ -0,0 +1,96 @@
...
+  else
+    code ^ ";;"
+
+module Command = struct
+  let _open_repl =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      let (_ : unit Promise.t) =
+        let open Promise.Syntax in
+        let sandbox = Extension_instance.sandbox instance in
+        let+ _ = create_terminal sandbox in
```
rgrinberg, line 57: Should be `let+ (_ : ..)` with a proper type annotation.

src/repl.ml
```diff
@@ -0,0 +1,96 @@
...
+  if start_line = end_line && start_char = end_char then
+    let document = Vscode.TextEditor.document text_editor in
+    let line = Vscode.TextDocument.lineAt document ~line:start_line in
+    Vscode.TextLine.text line
+  else
+    let document = Vscode.TextEditor.document text_editor in
+    Vscode.TextDocument.getText document ~range:(selection :> Vscode.Range.t) ()
+
+let prepare_code code =
+  let end_chars = String.sub code ~pos:(String.length code - 2) ~len:2 in
```
rgrinberg, line 45: Base should have a function for suffix checking. Something like `String.is_suffix` I think.

src/repl.ml
```diff
@@ -0,0 +1,96 @@
...
+        let sandbox = Extension_instance.sandbox instance in
+        let+ term = create_terminal sandbox in
+        match term with
+        | Error err -> show_message `Error "Could not start the REPL: %s" err
+        | Ok term ->
+          let code = get_code textEditor in
+          if String.length code > 0 then
+            let code = prepare_code code in
+            Terminal_sandbox.send term code
+          else
```
rgrinberg, line 77: No need for the trailing `else`.

src/terminal_sandbox.ml
```diff
@@ -69,23 +69,30 @@ let get_shell_args () =
...
   let shellArgs = `Strings args in
   Window.createTerminal ~name ~shellPath ~shellArgs ()
 
 let dispose = Terminal.dispose
 
 let show t = Terminal.show t ()
+
+let send t text =
+  let last_char = text.[String.length text - 1] in
+  let addNewLine = not (Char.equal last_char '\n') in
```
rgrinberg, line 97: Should use `String.is_suffix` as well.

src/terminal_sandbox.ml
```diff
@@ -69,23 +69,30 @@ let get_shell_args () =
...
   let shell_path = get_shell_path () in
   let shell_args = get_shell_args () in
   let ({ Cmd.bin; args } as command) =
-    match Sandbox.get_command sandbox shell_path shell_args with
-    | Spawn spawn -> spawn
-    | Shell command_line -> (
-      match Platform.shell with
-      | Sh bin -> { bin; args = [ "-c"; command_line ] }
-      | PowerShell bin -> { bin; args = [ "-c"; "& " ^ command_line ] } )
+    Option.value command
```
rgrinberg, line 76: Use a pattern match here instead of `Option.value`. It's not obvious that whatever is generating `~default` won't fail, so let's just be safe and only call it when needed.

src/vscode_ocaml_platform.ml
```diff
@@ -13,7 +13,17 @@ let suggest_to_pick_sandbox instance =
       ()
   in
   Option.iter selection ~f:(fun () ->
-      Extension_commands.select_sandbox.handler instance ~args:[])
+      let open Promise.Syntax in
+      let (_ : unit Promise.t) =
```
tmattio, line 17: This should probably use executeCommand instead

package.json
```diff
@@ -124,32 +124,54 @@
...
+        "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex || editorLangId == ocaml.menhir"
       },
       {
         "command": "editor.action.codeAction",
         "key": "Alt+D",
         "args": {
           "kind": "destruct"
         },
-        "when": "editorLangId  == ocaml || editorLangId == reason"
+        "when": "editorLangId == ocaml || editorLangId == reason"
```
rgrinberg, line 151: I think this toplevel only works for OCaml unfortunately.

vscode/vscode/vscode.mli
```diff
@@ -304,6 +304,25 @@ module Selection : sig
 
   val t_to_js : t -> Ojs.t
 
+  val start : t -> Position.t
```
tmattio, line 307: Should be possible to use `include module type of ... with type t :=`

rgrinberg:
Looks good. I don't find `$ dune utop` too useful at its current state, but its nice to have this for when we do improve `$ dune utop`.

tmattio:
>  I don't find $ dune utop too useful at its current state, but its nice to have this for when we do improve $ dune utop.

If users want to evaluate code that depends on the current library modules, that wouldn't work without `dune utop`, right? Although I probably need to add `--implicit-bindings` for it to work.

rgrinberg:
> If users want to evaluate code that depends on the current library modules, that wouldn't work without dune utop, right? Although I probably need to add --implicit-bindings for it to work.

Yeah, we need to use dune somehow. The issue for me is that it's only possible to access the public api's via `$ dune utop`. It would be much better to have something like `$ dune utop bar/foo.ml` that will allow us to peek into the internals of `Foo`.

package.json
tmattio: That's for the destruct command, I just removed a space

src/repl.ml
tmattio: Good idea, done

tmattio:
This is ready I think. I'm not too happy with the number of workarounds in there, so it would be nice to revisit the implementation once we have the toolchain ready (#463) and Dune RPC. In the meantime, the current workflow is the following:

- If `dune build` passes and the current sandbox has `utop` installed, the REPL will be `dune utop`
- If `dune build` fails and the current sandbox has `utop` installed, the REPL will be `utop` 
- Else, the REPL will be `ocaml`

src/terminal_sandbox.ml
```diff
@@ -69,23 +69,35 @@ let get_shell_args () =
...
   let shellArgs = `Strings args in
   Window.createTerminal ~name ~shellPath ~shellArgs ()
 
 let dispose = Terminal.dispose
 
 let show t = Terminal.show t ()
+
+let send t text =
+  let addNewLine =
+    if String.is_suffix text ~suffix:"\n" then
```
rgrinberg, line 98: `not (String.is_suffix text ~suffix:"\n")`

src/repl.ml
```diff
@@ -0,0 +1,152 @@
...
+      ~id:Extension_consts.Commands.evaluate_selection handler
+end
+
+let register extension instance =
+  let disposable =
+    Vscode.Window.onDidCloseTerminal ()
+      ~listener:(fun terminal ->
+        if String.equal (Vscode.Terminal.name terminal) name then
+          Extension_instance.close_repl instance
+        else
```
rgrinberg, line 148: `else` not needed.

src/repl.ml
```diff
@@ -0,0 +1,152 @@
...
+let has_utop sandbox =
+  let open Promise.Syntax in
+  let cmd =
+    let bin =
+      match Platform.t with
+      | Win32 -> "where"
+      | Darwin
+      | Linux
+      | Other ->
+        "which"
```
rgrinberg, line 14: `Cmd.which` already exists

src/repl.ml
```diff
@@ -0,0 +1,152 @@
...
+  | Some term ->
+    Terminal_sandbox.show term;
+    Promise.return (Ok term)
+  | None -> (
+    let open Promise.Syntax in
+    let* has_utop = has_utop sandbox in
+    let* can_build = can_build sandbox in
+    let cmd =
+      match (has_utop, can_build) with
+      | true, true -> Sandbox.get_command sandbox "dune" [ "utop" ]
```
rgrinberg, line 52: What would be simpler than all of this is to just ask the user what command to use to launch the repl.

src/repl.ml
```diff
@@ -0,0 +1,152 @@
...
+  else
+    code ^ ";;"
+
+module Command = struct
+  let _open_repl =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      let (_ : unit Promise.t) =
+        let open Promise.Syntax in
+        let sandbox = Extension_instance.sandbox instance in
+        let+ (_ : (Terminal.t, string) result) =
```
rgrinberg, line 113: Do we want to at least log the error?

src/repl.ml
tmattio: We would still have to detect if utop is available and handle the potential errors. I'm not sure asking the user would simplify the code that much

rgrinberg: Sorry, I meant simplify from the point of the user. Currently, this functionality doesn't really work if the user wants to use `down`, `rtop`, or just pass some flags to `dune utop`. Also, it requires remembering a few implementation details about the project that aren't too intuitive.  If we let the user pick how they launch the toplevel we can sidestep all of this.

tmattio: I see, maybe we could have a setting for the command used to run the repl? This way users would have a good default, and they could change their repl if needed.

I think most users would be happy with utop, as it's the recommended repl for OCaml, so I wouldn't want to interrupt the workflow with a popup for them.

tmattio: I added settings to set the REPL path and arguments. Here are the new settings:

| Name                               | Description                                                                                             | Default |
| ---------------------------------- | ------------------------------------------------------------------------------------------------------- | ------- |
| `ocaml.repl.path.linux`            | The path of the REPL that the extension uses on Linux                                                   | `null`  |
| `ocaml.repl.path.osx`              | The path of the REPL that the extension uses on macOS                                                   | `null`  |
| `ocaml.repl.path.windows`          | The path of the REPL that the extension uses on Windows                                                 | `null`  |
| `ocaml.repl.args.linux`            | The REPL arguments that the extension uses on Linux                                                     | `null`  |
| `ocaml.repl.args.osx`              | The REPL arguments that the extension uses on macOS                                                     | `null`  |
| `ocaml.repl.args.windows`          | The REPL arguments that the extension uses on Window                                                    | `null`  |

rgrinberg: Why is it per platform? I get that it might be useful in some cases, but most users only use a single platform.

tmattio: Agree it's not very useful, I removed them in favor of `ocaml.repl.path` and `ocaml.repl.args`

tmattio:
@rgrinberg This is ready to merge if the addition of the settings looks good to you.

src/repl.ml
```diff
@@ -0,0 +1,191 @@
+open Import
+
+module ReplPath = struct
```
rgrinberg, line 3: `Repl_path`.

src/repl.ml
```diff
@@ -0,0 +1,191 @@
...
+  let to_json (t : t) =
+    let open Jsonoo.Encode in
+    nullable string t
+
+  let key = "path"
+
+  let t = Settings.create ~scope:Global ~key ~of_json ~to_json
+end
+
+module ReplArgs = struct
```
rgrinberg, line 19: `Repl_args`.

src/repl.ml
```diff
@@ -0,0 +1,191 @@
...
+  Option.join (Settings.get ~section:"ocaml.repl" ReplPath.t)
+
+let get_repl_args () =
+  Option.join (Settings.get ~section:"ocaml.repl" ReplArgs.t)
+
+let name = "REPL"
+
+let has_utop sandbox =
+  let open Promise.Syntax in
+  let cmd = Sandbox.get_command sandbox "utop" [ "-version" ] in
```
rgrinberg, line 45: This check isn't quite right for `$ dune utop` since we don't use the binary. In any case, I suggest that we use `Bin.which "utop"` to make a check like this (as we do elsewhere).

src/repl.ml
```diff
@@ -0,0 +1,191 @@
...
+  let start_char =
+    Vscode.Selection.start selection |> Vscode.Position.character
+  in
+  let end_char = Vscode.Selection.end_ selection |> Vscode.Position.character in
+  if start_line = end_line && start_char = end_char then
+    let document = Vscode.TextEditor.document text_editor in
+    let line = Vscode.TextDocument.lineAt document ~line:start_line in
+    Vscode.TextLine.text line
+  else
+    let document = Vscode.TextEditor.document text_editor in
```
rgrinberg, line 136: This binding occurs in both `then` and `else` clauses. Hence it can exist outside the `if` altogether.

vscode/vscode/vscode.mli
```diff
@@ -262,6 +262,8 @@ end
 module Selection : sig
   include Js.T with type t = private Range.t
 
+  include module type of Range with type t := t
+
```
mnxn, line 266: These lines can be combined as 

```ocaml
include module type of Range with type t = private Range.t
```

src/bindings/node/node.ml
```diff
@@ -6,6 +6,22 @@ let __filename () =
 let __dirname () =
   Js_of_ocaml.Js.Unsafe.eval_string "__dirname" |> Js_of_ocaml.Js.to_string
 
+module Timeout = struct
+  type t = private Ojs.t [@@js]
```
mnxn, line 10: Use `include Class.Make ()` instead.

mnxn (changes requested):

src/repl.ml
tmattio: `Cmd.check` is global so doesn't do exactly what we want here. Let's say the user has utop in its path (globally installed, or in the global switch), `Cmd.check` will return `Ok`, but `dune utop` and `opam exec -- utop` would fail.

rgrinberg: Yes, but you can fix this by passing the PATH of the sandbox to which.

tmattio: True, although we'd need to get the env differently for Esy, and we'd need to run a command (i.e. `opam env` or the esy equivalent) and parse them to access the env of the sandbox. The benefit of `Cmd.which` is that we can save a command execution, which won't be the case in this specific case.

rgrinberg: > although we'd need to get the env differently for Esy, and we'd need to run a command (i.e. opam env or the esy equivalent) and parse them to access the env of the sandbox

We had code that did this before. I wonder where it went 🤔 In any case it would be useful to have it.

> The benefit of Cmd.which is that we can save a command execution, which won't be the case in this specific case.

My goal isn't to shave off a command invocation, but to stop relying on the command lines of these tools and treat them all consistently. However, I just checked and we already rely on `--version` for lsp, so I suppose we can just leave this be and get rid of this stuff another time.

tmattio: > We had code that did this before. I wonder where it went

I did something to use a custom `env` for `Cmd.check` and `Cmd.output` in #463, and actually ported these bits in this PR, but I don't recall seeing code that calls `opam env`.

tmattio:
@rgrinberg @mnxn let me know if there's anything else you'd like me to change

mnxn (approved):

rgrinberg (approved):

d4hines:
I would _love_ to try this one out! Any chance we can get a release (or a beta/insiders release?). Thanks!

ulugbekna:
@d4hines in case you don't want to wait for the release, you can try the new extension by cloning the repo, installing the dependencies using opam, and running `make install`.

smorimoto:
Alternatively, you can select the latest master build here to download the artifact and install it. This is probably the easiest way.
https://github.com/ocamllabs/vscode-ocaml-platform/actions/workflows/main.yml?query=branch%3Amaster

```
wget <URL>
unzip <PATH>
code --install-extension <FILE>
```

d4hines:
Thanks for the advice @ulugbekna and @smorimoto ; however, I'm having issues: any command I try to run in gives me an error like:
```
Command 'OCaml: Select a Sandbox for this Workspace' resulted in an error (command 'ocaml.select-sandbox' not found)
```

To clarify, I'm on Windows10, running the Insider's build of VS Code, using the Remote WSL extension to develop on Arch Linux WSL2, so it might be a perfect storm 😅

d4hines:
I have no idea why, but the issue was that whenever I installed the VSIX manually or with `make install`, many node_modules ended up missing. Running `yarn` inside the `~/.vscode-server/extensions/ocaml-platform-blah-blah` folder fixed the issue.

PR #506, opened 2021-01-25
Add comment when inferring interfaces

tmattio:
With this PR, the inferred interfaces are prepended with a comment when switching to a non-existing file.

image

This addresses user feedback that indicated that switching to the interface of a module was sometimes confusing as users didn't know if the interface had been inferred, or if it already existed.

@rgrinberg At this point, I'm also convinced that this feature has passed the threshold of hacks and complexity. This is my last attempt, but I'll just remove the whole thing if you prefer.

### DIFF
--- a/src/switch_impl_intf.ml
+++ b/src/switch_impl_intf.ml
@@ -21,6 +21,10 @@ let insert_inferred_intf ~source_uri client text_editor =
   if String.is_suffix source_uri ~suffix:".ml" then
     (* If the source file was a .ml, infer the interface *)
     let+ inferred_intf = send_infer_intf_request client source_uri in
+    let inferred_intf =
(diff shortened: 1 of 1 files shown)


mnxn:
I'm not sure I like the idea of inserting such a comment. 
If a user is using the interface inference as part of their programming workflow, they would have to delete the comment every time.

Wouldn't a notification be more intuitive?

rgrinberg:
Even more intuitive would be a code lens entry saying its generated. But that's too much hassle. We can keep this feature around for now, but let's not waste time improving it. There's more important things to improve and I wouldn't be surprised if all of this ends up being obsolete once the protocol improves.

tmattio:
Alright, closing for now then

PR #510, opened 2021-01-27
Use actions-ml instead for CI

smorimoto:
It deploys the pre-built compiler cache for Actions runners in Ubuntu and macOS. It also tries to speed up builds as much as possible by handling the dune and opam caches internally, and it finds the latest OCaml patch version, so maintenance costs are really kept to a minimum. In addition, the log output has been greatly improved.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
           - ubuntu-latest
           - windows-latest
         ocaml-version:
-          - 4.11.1
+          - 4.11.x
         node-version:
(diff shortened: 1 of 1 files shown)


smorimoto:
There is still a problem in actions/cache doesn't work with Cygwin as expected. I already have a spec to fix it in actions/cache, so I will fix it soon.

mnxn (approved):

smorimoto:
The failure of the test on Windows has nothing related to this PR itself so I will merge this.

PR #518, opened 2021-02-02
Fix Windows Opam local switch detection

tmattio:


src/opam.ml
```diff
@@ -9,7 +9,7 @@ module Switch = struct
     | "" -> None
     | switch_name ->
       let switch_name = String.strip switch_name in
-      if Char.equal switch_name.[0] '/' then
+      if Node.Path.isAbsolute switch_name then
```
rgrinberg, line 12: You should use `Path.is_absolute (Path.of_string ..))` so that we stick to one path module

rgrinberg (approved):

PR #527, opened 2021-02-10
Add lint jobs to GitHub Actions workflow

smorimoto:
This PR adds 2 lint jobs. (There is currently no format checking for OCaml code.)


smorimoto:
Hmm, I'm seeing a weird error on Windows that's not related to this change. By the way, there doesn't seem to be a reason to run CI on Windows or macOS, but is there any reason to do that?

PR #539, opened 2021-03-04
Update VSCode bindings for 1.54 

mnxn:
Adds changes from the latest VSCode API to bindings.


mnxn:
This should be merged after dependabot updates the VSCode dependency to 1.54.

smorimoto (approved):

PR #541, opened 2021-03-05
Display shell command output 

mnxn:
Displays the output of shell commands in a (more) user-friendly way:

image

The output channel is updated as the process writes to stdout, unlike the JSON logging which has to wait for the process to complete. 

The package management commands in the activity tab (add/remove package or switch) will automatically focus on the output channel.

The effect of this feature is that users will be able to see the progress of installing/modifying their sandbox, whereas previously there would only be an infinitely-loading notification. This is especially useful when adding packages with a lot of dependencies that take a long time to install.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Highlight OCaml source code inside Cinaps comments (#547)
 
+- Show the output of shell commands in an output channel. The output channel is
+  automatically focused when running a package management command from the OCaml
+  activity tab (#541)
+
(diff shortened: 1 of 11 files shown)


rgrinberg:
It seems like a strange thing to have an option for. If the progress bar isn't satisfying, we should just focus on the command output and be done with it.

mnxn:
> If the progress bar isn't satisfying, we should just focus on the command output

You're right. I removed the option.

rgrinberg (approved):

PR #551, opened 2021-03-07
Ask for confirmation when removing opam packages or switches

mnxn:
Removing a switch or package is currently done without a confirmation box. Users could accidentally uninstall a package or switch and have to wait a long time to reinstall everything they need.

A simple confirmation dialog (using `showInformationMessage`) should be sufficient to avoid accidental removals.

### DIFF
--- a/src/import.ml
+++ b/src/import.ml
@@ -69,11 +69,13 @@ let open_file_in_text_editor target_uri =
   let+ text_editor = Window.showTextDocument ~document:(`TextDocument doc) () in
   text_editor
 
-let ask_confirmation message =
+let with_confirmation message ~yes ?(no = "Cancel") f =
   let open Promise.Syntax in
-  let+ choice =
+  let* choice =
     Vscode.Window.showInformationMessage ~message
-      ~choices:[ ("Yes", true); ("No", false) ]
+      ~choices:[ (yes, true); (no, false) ]
       ()
   in
-  Option.value choice ~default:false
+  match choice with
+  | Some true -> f ()
+  | _ -> Promise.return ()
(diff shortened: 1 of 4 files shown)


src/import.ml
```diff
@@ -68,3 +68,12 @@ let open_file_in_text_editor target_uri =
   in
   let+ text_editor = Window.showTextDocument ~document:(`TextDocument doc) () in
   text_editor
+
+let ask_confirmation message =
+  let open Promise.Syntax in
+  let+ choice =
+    Vscode.Window.showInformationMessage ~message
+      ~choices:[ ("Yes", true); ("No", false) ]
```
tmattio, line 76: I would make the actions more explicit:  e.g. "Remove switch" "Cancel"

src/treeview_switches.ml
```diff
@@ -101,22 +101,31 @@ module Command = struct
...
-              Vscode.Commands.executeCommand
-                ~command:Extension_consts.Commands.refresh_switches ~args:[]
-            in
-            let (_ : Ojs.t option Promise.t) =
-              Vscode.Commands.executeCommand
-                ~command:Extension_consts.Commands.refresh_sandbox ~args:[]
-            in
-            show_message `Info "The switch has been removed successfully." )
+          let* confirmed =
+            ask_confirmation
```
tmattio, line 107: What do you think of a function `with_confirmation` that takes a `unit -> unit Promise.t` and executes it only if the user confirmed it?

tmattio (approved):
LGTM, apart from my comment on having more explicit action buttons

src/treeview_switches.ml
mnxn: Sounds good.

PR #568, opened 2021-03-15
Refine quick start guide

smorimoto:


mnxn:
I think it would be better to keep it all in the README so it's all visible in the VSCode extension details:

image

With this change, users would have to click the link and view the instructions in their browser.

smorimoto:
Oh, indeed. But I just wanted to make a clear distinction because the current one looks a little messy. Well, it's not a strong opinion.

PR #570, opened 2021-03-23
Update to OCaml 4.12

mnxn:
CI was failing because the master branch of ocaml-lsp requires 4.12.

Also pins gen_js_api to 1.06 because 1.0.7 introduces some breaking changes.

PR #583, opened 2021-04-14
Update ocamlformat to 0.18.0

smorimoto:


mnxn:
When I last tried to upgrade gen_js_api to 1.0.7, I got errors with our existing bindings: https://github.com/LexiFi/gen_js_api/issues/148

I'll wait on the CI, but it's possible 1.0.8 has the same problems.

smorimoto:
Oh, indeed. It seems we need to fix it first.

mnxn:
@smorimoto: I updated gen_js_api to 1.0.8, can you rebase this PR on top of master please?

smorimoto:
@mnxn Sure!

.ocamlformat
```diff
@@ -1,4 +1,5 @@
-version=0.15.0
+version=0.18.0
+profile=conventional
```
smorimoto, line 2: See https://github.com/ocaml/dune/pull/4351

mnxn (approved):
Looks good.

I think we can re-enable ocamlformat in CI now. It was disabled in https://github.com/ocamllabs/vscode-ocaml-platform/pull/570. Should that be done in this PR or a new one?

smorimoto:
> I think we can re-enable ocamlformat in CI now. It was disabled in #570. Should that be done in this PR or a new one?

I will do that in a new one.

PR #588, opened 2021-04-21
implement custom request ocamllsp/dumpTrace

ulugbekna:
Related to https://github.com/ocaml/ocaml-lsp/pull/424


tmattio:
@ulugbekna Do we still want to work on this?

PR #589, opened 2021-04-23
Update gen_js_api to 1.0.8

mnxn:
Changes include:

- Adding `include [%js: ... ]` for `val` bindings (needed for ppxlib, see https://github.com/LexiFi/gen_js_api/issues/148).
- Using `Ojs.get_prop_ascii`/`Ojs.set_prop_ascii` instead of the deprecated `Ojs.get`/`Ojs.set` functions.
- Using `[@@js.get]` instead of heuristic `[@@js]` annotations.


tmattio:
That's great thanks @mnxn!

I have been working on the bindings for some time as well (lately here: https://github.com/tmattio/js-bindings/) and am planning on landing a PR to update the VSCode and Node bindings in a few days. In addition to upgrading to the latest version of 1.0.8, the will cover the entire VSCode and Node API so as to unblock us on several projects (Notebooks integration, AST explorer, welcome page, tools installation, etc.) as well as finally migrating the tests to OCaml.

I haven't really communicated about this work, so I apologize for causing some overlaps with the work you've been doing with this PR.

The PR looks good to me as-is though, so I'm in favor of merging it and we can discuss the bindings change independently of it.

rgrinberg (approved):

smorimoto (approved):

smorimoto:
@tmattio The idea is especially great for my team, but note that the bindings generated by ts2ocaml are not intended to be packaged like that.

smorimoto:
With that in mind, I think we still need to postpone using them a little. The reason we haven't released the tool to the public yet is that we were concerned that the use of it in production would significantly increase our maintenance burden. In other words, at this stage, it would be best to use it only to satisfy your curiosity.

tmattio:
Thanks for the additional context @smorimoto. The last thing I want is to increase maintenance time on your side 🙂

To be a bit more specific, the work I'm doing on increasing the coverage of our bindings started before you invited me to the repo. I was happy enough to use `ts2ocaml` to speed up this work (and hopefully provided useful feedback), but the bindings are something we need for numerous projects and the need for it is not the result of having access to the tool.

I want to state that I'm not expecting the release of the bindings linked above to add any additional work for you. As you can see in the bindings, I've done a lot of work on the interfaces after they have been generated, so the intended workflow was never to re-generate them (it's been two weeks I'm working on these after their bootstrapping, I'm not planning on doing that every time there is an update of `ts2ocaml` 😉).

I had verified with @cannorin on Slack that releasing packages containing generated code was ok at this stage (also considering that the actual code is now closer to the manual bindings I was writing before the generation than they are to the bootstrapped ones), and it seemed to be ok. 

I hope this clears up the concerns you're having, but please feel free to reach out on Slack if you'd like to discuss this further and define a plan you and your team are comfortable with 🙂

smorimoto:
Oh, I should have written the comments a little more carefully! 😨

First, our goal is to generate cleaner code without the manual work you do now. We expect that the bindings will be included in this repository as they are generated. That is, we don't expect a lot of bindings to be published to the opam repository. We have seen a number of unmaintained bindings in the various language communities, and we really want to avoid them. But as cannorin said, there is no reason to stop you, and that's completely fine. I just wanted to make clear that the project's goals and our rough plans are going in different directions.

And the other thing is that we are now students, working on it in a very limited time. We wanted to reduce the possibility of the tool becoming a blocking factor in this repository, for example.

PR #591, opened 2021-04-24
Reduce package size

smorimoto:
This greatly reduces package size and improves performance by bundling dependencies into a single file.

Before: `ocaml-platform.vsix (293 files, 2.34MB)`
After: `ocaml-platform.vsix (59 files, 347.19KB)`


smorimoto:
The only concern is that the build output used for the packaged and actual local debugging is strictly different. But there is no serious problem.

Makefile
```diff
@@ -1,9 +1,10 @@
 build:
-	dune build @vscode
+	dune build @all --profile=release
+	yarn esbuild _build/default/src/vscode_ocaml_platform.bc.js --bundle --external:vscode --outdir=dist --platform=node --target=es6 --minify-whitespace --minify-syntax
```
smorimoto, line 3: `--minify-identifiers` breaks our extension, so it cannot be used at this time.

.vscodeignore
```diff
@@ -1,18 +1,5 @@
...
-.merlin
-**/*.map
-azure-pipelines.yaml
-esy.json
-Makefile
+**
+!assets/
+!dist/
+!snippets/
+!syntaxes/
```
mnxn, line 5: Missing `languages/`.

smorimoto: Oops, indeed.

mnxn:
Seems to work fine, but doesn't this mess with source maps/stack traces? That could make it harder to debug if a user has a problem with the release version.

smorimoto:
I made debugging easier by generating a source map. This will increase package size slightly, but still, eliminate unnecessary files.

mnxn (approved):

PR #593, opened 2021-04-24
Core_kernel -> Base

smorimoto:
Core_kernel is too heavy for us. This PR greatly reduces the total number of packages to be installed and also reduces the CI time.


src/import.ml
```diff
@@ -25,7 +25,7 @@ module Option = struct
 end
 
 module Or_error = struct
-  type 'a t = ('a, string) result
+  type 'a t = ('a, string) Result.t
```
mnxn, line 28: I think it'd be better to just add the alias to this import module ourselves:
```ocaml
type ('a, 'b) result = ('a, 'b) Result.t
```

mnxn (approved):

src/import.ml
smorimoto: Changed

PR #594, opened 2021-04-24
Use esy action instead

smorimoto:


.github/workflows/main.yml
```diff
@@ -47,26 +47,10 @@ jobs:
       - name: Use latest esy
         run: npm install --global esy
 
-      # This step is required because esy's home directory doesn't exist at this point
       - name: Install esy packages
```
mnxn, line 50: This step should probably be renamed since its doing more than just installing esy packages.

smorimoto: I agree with part of it, but it's actually the same as what opam does except for the cache part.

mnxn (approved):

PR #599, opened 2021-05-02
Fix sourcemap generation

mnxn:
Previously the sources were not mapped correctly in stack traces because sourcemaps are disabled by default in js_of_ocaml release mode.

Changes:
- `dune build src/vscode_ocaml_platform.bc.js` instead of `dune build @all` because `@all` unnecessarily copies the entire directory to `_build`.
- `--sourcemap` instead of `--sourcemap=external` because `external` does not include the `sourceMappingURL` in the generated JS.
- `--sources-content=false` because the generated JS is not being used in a debugger.
-  Clear the dist folder in `make clean`.
- `--source-map` in Js_of_ocaml flags because it is disabled for `--release`.
- `--pretty` so proper function names appear in stack traces.
   It seems that this has to be enabled because other combinations of flags result in `TypeError: s.toUtf16 is not a function` at runtime.

Before this patch:
```
  dist\vscode_ocaml_platform.bc.js      691.3kb
  dist\vscode_ocaml_platform.bc.js.map    1.6mb
```

After:
```
  dist\vscode_ocaml_platform.bc.js      921.8kb
  dist\vscode_ocaml_platform.bc.js.map  549.0kb
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 build:
-	dune build @all --profile=release
+	dune build src/vscode_ocaml_platform.bc.js --profile=release
 	yarn esbuild _build/default/src/vscode_ocaml_platform.bc.js \
 		--bundle \
 		--external:vscode \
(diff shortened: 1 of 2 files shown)


rgrinberg (approved):
Looks good.  I believe this deserves a release as well.

mnxn:
> I believe this deserves a release as well.

I will create a release after I merge #600

PR #617, opened 2021-05-18
Use ocaml/setup-ocaml@v2

smorimoto:
This includes a postponed mocha upgrade.


smorimoto:
As for Windows failures, it is not a problem with `avsm/setup-ocaml`.

smorimoto:
Windows failure will be resolved once this is merged: https://github.com/fdopen/opam-repository-mingw/pull/94

PR #618, opened 2021-05-19
Update bindings

tmattio:
This PR updates the bindings to use the ones at [`js-bindings`](https://github.com/tmattio/js-bindings).

The bindings cover the entire APIs of ECMAScript standard library (from ES5 to ES2020), NodeJS and VSCode. I also added the `vscode-test` bindings that will be useful to migrate the tests to OCaml.

They have initially been generated using `ts2ocaml` and I've been manually refactoring them over the past weeks.

The motivation for this update is the following:

- Streamline development by providing complete bindings. The bindings creation is arguably the most time-consuming part of adding a new feature at the moment. Although the bindings are not perfect and will require some work, I expect that new features will be experimented with more easily if we don't need to bind the required APIs from scratch.
- Support the upcoming work of automatic binary installation. The next version of the automatic binary installation will come with an automated Opam installation so that we have a zero system depedency for the extension. This requires multiple APIs that are not binded currently, in particular, Node's `fs` and `https` APIs.
- Support the upcoming AST explorer work. @arozovyk is working on an implementation of this and depends on multiple bindings that are not currently implemented, in particular the WebView part of VSCode API.

Note that the bindings try to be as close as possible to their JS counterparts. Manual functions in the bindings are rare. To stay compatible with `promise_jsoo` and `jsonoo` while using the `es5` standard library, I wrapped them with conversions functions in the packages `src-bindings/ocaml-promise` and `src-bindings/ocaml-json` respectively.

In addition to the complete coverage of the API, the bindings now have the following properties:

- Use the `snake_case` syntax
- Use `include module type of` for inheritance to stay somehow close to the JS semantics.
- The implementation are generated from the `mli` files. We don't duplicate the types anymore.

The diff is seemingly huge, but no code has been changed apart from the function renaming (most of it is just `camelCase` to `snake_case` conversion) and the refactoring when the API changed.

`ts2ocaml` is still very new. It's been immensely helpful to bootstrap the bindings, but there are shortcomings that I didn't address (given the amount of code, it would take too much time to do everything in one shot). We should give these in mind:

- Some functions are generated with the same name and different signatures. When this happens, only the last definition will be usable. If you need another signature, you should manually refactor these to only include the relevant signatures with different names.
- The builder functions are not generated. I added them manually for the modules we needed, but chances are, you'll need to add them yourself if you need a module that is not used at the moment

Note: the code compiles but there still are some runtime errors. This needs to be fixed before merging, hence the PR being opened in Draft mode.


mnxn:
> To stay compatible with promise_jsoo and jsonoo while using the es5 standard library, I wrapped them with conversions functions in the packages src-bindings/ocaml-promise and src-bindings/ocaml-json respectively.

This would add extra maintenance work to make sure `ocaml-promise` and `ocaml-json` keep up with the published packages.

Can't you go the other way? That is, to make the es5 types point to the promise_jsoo and jsonoo types?

Here: https://github.com/ocamllabs/vscode-ocaml-platform/blob/f05172fdd33f0cd7b60fb69e3c1c2af63999ab12/src-bindings/es5/es5.mli#L3192

```ocaml
module Promise : sig
  type +'T t = 'T Promise_jsoo.t
  ...
end
```

and here: https://github.com/ocamllabs/vscode-ocaml-platform/blob/f05172fdd33f0cd7b60fb69e3c1c2af63999ab12/src-bindings/es5/es5.mli#L3074

```ocaml
module Json : sig
  type t = Jsonoo.t
  ...
end
```

mnxn:
> Use the snake_case syntax

I can't find the discussion, but the previous decision was that we should maintain the original naming conventions because it makes it easier to consult official documentation.

rgrinberg:
Max Lantas ***@***.***> writes:

> I can't find the discussion, but the previous decision was that we
> should maintain the original naming conventions because it makes it
> easier to consult official documentation.

Yes, this is my preference. Bindings are often very poorly documented
and one must constantly refer to the original docs. Let's keep that as
easy as possible.

tmattio:
Oups, didn't want to start a bikeshedding conversation :sweat_smile:

I don't think using `camelCase` really produces the intended benefits.

For one, mixing the two styles is an annoyance I have with the current codebase. Having snake_case and camelCase in the same function is awkward, and having to remember whether the function is in camelCase or snake_case depending on where it comes from is frustrating.

I'm not convinced not having to search for `showInformationMessage` to get documentation for `show_information_message` outweighs the cost of the above.

If the reason for keeping the camelCase syntax is to ease doc searching, I'd prefer to spend some time adding docstring that redirects to the function documentations instead, what do you think?

rgrinberg:
> For one, mixing the two styles is an annoyance I have with the current codebase. Having snake_case and camelCase in the same function is awkward, and having to remember whether the function is in camelCase or snake_case depending on where it comes from is frustrating.
When does one need to remember this distinction? If you're creating an identifier, you're always using snake case.

src-bindings/es2015/es2015.mli
```diff
@@ -0,0 +1,37 @@
+include module type of struct
+  include Es5
+end
```
mnxn, line 3: What's the purpose of wrapping the module in a `struct ... end`?
```ocaml
include module type of Es5
```

tmattio:
> If you're creating an identifier, you're always using snake case.

Sure, I meant when using a function that already exists. So functions that come from `Node` and `VSCode` use `camelCase`, and functions from import or other modules that might be opened use `snake_case`, so you have to remember where every function comes from.

What do you think of my suggestion of generating docstrings with a link to the official function documentation? I think it addresses the concerns you have regarding doc searching (even offer a better experience than currently), while bypassing the drawbacks of mixing the two syntaxes.

rgrinberg:
Thibaut Mattio ***@***.***> writes:

> Sure, I meant when using a function that already exists. So functions
> that come from `Node` and `VSCode` use `camelCase`, and functions from
> import or other modules that might be opened use `snake_case`, so you
> have to remember where every function comes from.

I don't see the issue. To use the API of VSCode and Node, the case of
the identifiers needs to be known in advance anyway.

> What do you think of my suggestion of generating docstrings with a
> link to the official function documentation? I think it addresses the
> concerns you have regarding doc searching (even offer a better
> experience than currently), while bypassing the drawbacks of mixing
> the two syntaxes.

It seems like a good feature on its own, but it doesn't really address
my concerns.

I'm open minded to the change, but it needs to be argued in an objective
way as an improvment. Right now, it seems more like a personal
preference, and that's not enough to change our conventions.

rgrinberg:
> ts2ocaml is still very new. It's been immensely helpful to bootstrap the bindings, but there are shortcomings that I didn't address (given the amount of code, it would take too much time to do everything in one shot). We should give these in mind:

Where do I download and test this tool out? Google is failing me.

tmattio:
> It seems like a good feature on its own, but it doesn't really address my concerns.

Why not? I understood your concern is that having `snake_case` will make it more difficult to search documentation because the official function names are different. If you have a docstring with a link to the official documentation, this is not a problem anymore, right?

> I'm open minded to the change, but it needs to be argued in an objective way as an improvment. Right now, it seems more like a personal preference, and that's not enough to change our conventions.

I'm not sure how to objectively argue about syntax 😄 

Joke aside, I thought I gave relatively objective arguments, but if I'm the only one being annoyed by the mix of the syntaxes, I agree this is not enough to change the convention.

I just hope to understand a bit better why the docstring suggestion does not address your concerns because it seems to me like a much better option for doc accessibility (putting the syntax debate aside). If indeed I am missing something and it's not a good solution, I don't think I have a better idea so I'll revert this part of the PR 🙂

tmattio:
> Where do I download and test this tool out? Google is failing me.

`ts2ocaml` is not public yet, unfortunately. @smorimoto invited me to the repository hosted at https://github.com/ocsigen so I could beta test it.

However, you can look into the originally generated files in the `_gen` directories of the `js-bindings` repo. So for instance here are the generated bindings for `node`: https://github.com/tmattio/js-bindings/tree/main/lib/node/_gen

tmattio:
> This would add extra maintenance work to make sure ocaml-promise and ocaml-json keep up with the published packages.
> 
> Can't you go the other way? That is, to make the es5 types point to the promise_jsoo and jsonoo types?

To give some additional context, I intend to release these bindings outside of VSCode in the future.
It'll be useful to build tools that work on both VSCode and with a CLI. So for instance, the automatic platform tool installation will share code between the VSCode integration and the `opam-tools` plugin (with a node backend for Jsoo and a native backend for the CLI).

I'm also hoping to add more bindings to `js-bindings` in the future.

With this in mind, I'd like to keep `es5` as slim as possible, as it'll be used by all of the bindings.

`es5` provides raw bindings, that directly map to the JS functions, whereas `jsonoo` and `promise_jsoo` provide additional functions on top of the original JS functions, so they are higher-level libraries than the raw bindings.

I agree this will bring additional maintenance work in the event that there is a change in these libraries, but I think this should be rare and given the amount of code, it shouldn't take too long to upstream the changes.

mnxn:
> With this in mind, I'd like to keep `es5` as slim as possible, as it'll be used by all of the bindings.

Okay, I think I understand your reasoning now.

In that case, I still don't think copying the source code of those libraries to make a few changes is the best idea. Especially since the modified version doesn't have any tests but the originals do.

I believe a better solution would be to add functors to the original two libraries so they can be parameterized by different types.

rgrinberg:
Thibaut Mattio ***@***.***> writes:

> I'm not sure how to objectively argue about syntax 😄 

I know you are joking, but I actually try to think about this as
objectively as possible.

Choosing your names correctly relieves the programmer of some (likely
minor) cognitive load. In this case, the cognitive load imposed is
maintaining a map of binding library to original API names. I see no
downsides to this apart from making the code look "non uniform". Which
IMHO is a subjective and hence less important consideration.

When I look at other projects, I see a similar trade-off being made. For
example:

* js_of_ocaml uses the original naming convention of the DOM even if
  it's camelCase.

* OCaml's Unix library uses traditional Unix names for
  functions/constants even though they don't necessarily follow OCaml's
  naming convention (EINTR vs Error_interrupted)

I'm pretty confident in following the example of these two projects
here.

rgrinberg:
Thibaut Mattio ***@***.***> writes:

> `ts2ocaml` is not public yet, unfortunately. @smorimoto invited me to
> the repository hosted at https://github.com/ocsigen so I could beta
> test it.

I see. I'd prefer to hold off merging this PR until the tool is public
then. I hope that doesn't come off as controversial.

tmattio:
> I see. I'd prefer to hold off merging this PR until the tool is public then. I hope that doesn't come off as controversial.

With the update to `gen_js_api.1.0.8` from @mnxn and given that I work on a native implementation for the tools installation, this PR is not blocking any of the initiatives that motivated this work. I still think it's a good addition, but we can certainly hold off merging.

I'd mention again, however, that the bindings have been extensively refactored after their generation. If you compare the current bindings with the generated ones, you probably couldn't tell the former comes from the latter. The release of `ts2ocaml` wouldn't really impact this PR, because I'm not intending to keep track with the tool updates (it would be too time-consuming to redo all the modifications for every release).

That being said, I opened issues on `ts2ocaml` for the different issues I had with the generated bindings that required me to work on them. If most of these issues are addressed, we could imagine using the generated bindings directly, but I don't know how much time this would take for this to happen, might be years..

rgrinberg:
Thibaut Mattio ***@***.***> writes:

> I'd mention again, however, that the bindings have been extensively
> refactored after their generation. If you compare the current bindings
> with the generated ones, you probably couldn't tell the former comes
> from the latter. The release of `ts2ocaml` wouldn't really impact this
> PR, because I'm not intending to keep track with the tool updates (it
> would be too time-consuming to redo all the modifications for every
> release).

I see. I missed that subtlety. In that case it's definitely not
required.

mnxn:
@tmattio: I've pushed some updates to the `trunk` branch of [promise_jsoo](https://github.com/mnxn/promise_jsoo) and [jsonoo](https://github.com/mnxn/jsonoo) that should make it possible to use custom type representations.

So now, it should be possible to do the following in `import.ml` and remove the `ocaml-promise`/`ocaml-json` libraries:

```ocaml
module Promise = Promise.Make (Es5.Promise)
module Jsonoo = Jsonoo.Make (Es5.Json)
```

I haven't had a chance to test it with this PR yet, so please let me know if anything else is needed before I publish new library versions.

hackwaly:
Hi @tmattio, what status being there? Is it stable to depend on it to implement other features such as debugger integration?

tmattio:
Quick update on this PR: I don't think we will merge this as-is. @cannorin has been working on `ts2ocaml` extensively since I opened this PR, and most of the work I did manually to improve the generated bindings is now taken care of by `ts2ocaml`.

This PR is not on the critical path for anything, so I'd like to wait for an initial release of `ts2ocaml` and will open a PR to switch to the generated bindings (without any manual modification).

I'm closing this in the meantime to avoid any confusion.

PR #622, opened 2021-05-27
Nix package management

rgrinberg:
`nix-shell` in the repository now brings up a repository that can build
the project.


Makefile
```diff
@@ -38,3 +38,7 @@ pkg: build
 # builds, packages, and installs the extension to your VS Code
 install: pkg
 	code --force --install-extension test_extension.vsix
+
+.PHONY: nix/opam-selection.nix
+nix/opam-selection.nix:
```
tmattio, line 43: Why not `nix-shell`, or even `nix`? It's a bit shorter

tmattio:
While I'm not against adding nix setups, I'm wondering what benefits you get by using nix for `vscode-ocaml-platform`. I understood it was useful for `dune` and `ocaml-lsp` because of their dependency profile, but there's nothing really specific about `vscode-ocaml-platform` from what I can tell.

My fear is that it makes the repository unnecessarily convoluted. We'd have files to support an Opam workflow, an Esy workflow (which I am not sure any maintainer use?), and now a Nix workflow.

rgrinberg:
Thibaut Mattio ***@***.***> writes:

> While I'm not against adding nix setups, I'm wondering what benefits
> you get by using nix for `vscode-ocaml-platform`. I understood it was
> useful for `dune` and `ocaml-lsp` because of their dependency profile,
> but there's nothing really specific about `vscode-ocaml-platform` from
> what I can tell.

vscode-ocaml-platform has plenty of non OCaml dependencies that are
currently not managed by opam. For example, yarn, vscode itself are
needed for actually testing the plugin. Moreover, it simplifies testing
the plugin with packages still in development - pins are not adequate
for me.

> My fear is that it makes the repository unnecessarily convoluted. We'd
> have files to support an Opam workflow, an Esy workflow (which I am
> not sure any maintainer use?), and now a Nix workflow.

Yeah, I suppose that can be a problem. I certainly use nix and plan to
maintain but we should delete stuff that is unmaintained otherwise. So
if you aren't aware of anyone using esy, we can delete the lcok files.
They can be brought back easily anyway.

Makefile
rgrinberg: Well it's named after the target it produces. It doesn't need a short name as it's not something you should run often. It regenerates the build plan from opam-repository which is something that only needs to be done to if you're updating dependencies.

mnxn (approved):
I have very little Nix experience, so it took me a while to get a NixOS VM set up, but it looks like everything works correctly.

rgrinberg:
@mnxn It should be possible to run nix on WSL or Darwin (as I do) without a VM. NixOS is an OS that is baed on Nix but one can also use it as a standalone package manager.

mnxn:
@rgrinberg: I'm aware. I didn't want to mess with my existing systems so I used a VirtualBox image from https://nixos.org/download.html#nixos-virtualbox.

I only had to upgrade to the unstable channel (to get OCaml 4.12) and every worked fine after that.

rgrinberg:
> I only had to upgrade to the unstable channel (to get OCaml 4.12) and everything worked fine after that.

Good point. I switched this PR to use unstable directly until ocamlformat is in stable.

PR #637, opened 2021-06-14
Support `ocamllsp/nextHole` custom request

ulugbekna:
It is best to review all changes at once instead of commit-by-commit since there is some churn (but I don't yet squash to be able to revert some stuff if controversial)

https://user-images.githubusercontent.com/16353531/121881399-abd23c80-cd28-11eb-98cf-fe8cc9d56956.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -227,6 +227,10 @@
         "command": "ocaml.evaluate-selection",
         "key": "Shift+Enter",
         "when": "editorTextFocus && editorLangId == ocaml || editorTextFocus && editorLangId == ocaml.interface || editorTextFocus && editorLangId == reason || editorTextFocus && editorLangId == ocaml.ocamllex || editorTextFocus && editorLangId == ocaml.menhir"
+      },
+      {
+        "command": "ocaml.next-hole",
+        "key": "Alt+L"
       }
     ],
     "menus": {
--- a/package.json
+++ b/package.json
@@ -183,7 +183,7 @@
       {
         "command": "ocaml.next-hole",
         "category": "OCaml",
-        "title": "Next Hole"
+        "title": "Jump to Next Typed Hole"
       },
       {
         "command": "ocaml.current-dune-file",
--- a/package.json
+++ b/package.json
@@ -181,6 +181,11 @@
         }
       },
       {
+        "command": "ocaml.next-hole",
+        "category": "OCaml",
+        "title": "Next Hole"
+      },
+      {
         "command": "ocaml.current-dune-file",
         "category": "OCaml",
         "title": "Open Dune File (located in the same folder)"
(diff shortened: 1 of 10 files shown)


src/import.ml
```diff
@@ -43,6 +43,12 @@ let show_message kind fmt =
       ())
     fmt
 
+let show_err fmt = show_message `Error fmt
+
+let show_warn fmt = show_message `Warn fmt
+
+let show_info fmt = show_message `Info fmt
```
mnxn, line 50: These seem unnecessary. `show_message` is already a shortcut to the `Window.show...Message` functions.

src/extension_consts.ml
```diff
@@ -32,6 +32,20 @@ module Commands = struct
   let evaluate_selection = ocaml_prefixed "evaluate-selection"
 
   let open_repl = ocaml_prefixed "open-repl"
+
+  let next_hole = ocaml_prefixed "next-hole"
+end
+
+module Command_errors = struct
+  let avail_when_editor_open ~expl cmd_name =
```
mnxn, line 40: You might as well make this `available_when_editor_open`.

Omitting 4 characters doesn't help much with brevity here, instead it makes it harder to read.

package.json
```diff
@@ -222,6 +227,10 @@
         "command": "ocaml.evaluate-selection",
         "key": "Shift+Enter",
         "when": "editorTextFocus && editorLangId == ocaml || editorTextFocus && editorLangId == ocaml.interface || editorTextFocus && editorLangId == reason || editorTextFocus && editorLangId == ocaml.ocamllex || editorTextFocus && editorLangId == ocaml.menhir"
+      },
+      {
+        "command": "ocaml.next-hole",
+        "key": "Alt+L"
```
mnxn, line 233: I think this entry needs some `when` qualifiers since a comment says that the command is only available when a file is open.

src/import.ml
```diff
@@ -86,3 +92,27 @@ let with_confirmation message ~yes ?(no = "Cancel") f =
...
+(** Builds application-specific functionality around [Vscode.Position] *)
+module Position = struct
+  include Vscode.Position
+
+  let t_of_jsonoo json =
+    let line = Jsonoo.Decode.(field "line" int) json in
+    let character = Jsonoo.Decode.(field "character" int) json in
+    Position.make ~line ~character
+
+  let t_to_jsonoo t = t |> Position.t_to_js |> Jsonoo.t_of_js
```
mnxn, line 105: `Jsonoo.t_of_js` isn't meant for encoding values.

You should use the `Jsonoo.Encode.*` functions.

mnxn (changes requested):

ulugbekna:
Hi @mnxn, thanks for the review. I'm turning this PR to a draft, since it maybe a more lean on the ocamllsp side if we use a single new custom request to jump to prev/next typed hole (vs two custom requests that this PR would require in future). I tried to embed all of your comments in #643 :-)

PR #642, opened 2021-06-24
Add some bindings to vscode diagnostics manipulation

ulugbekna:
I had this commit as part of #640 , but since I'm not sure we're going further with the approach of #640, decided to create another PR. 

The bindings aren't full, but were somewhat tricky to get right (at least for me as a newbie to ojs). 

Fixed a "deprecated" warning along the way.

### DIFF
--- a/vscode/vscode/vscode.ml
+++ b/vscode/vscode/vscode.ml
@@ -2753,7 +2753,7 @@ module Languages = struct
     val getDiagnostics : Uri.t -> Diagnostic.t list
       [@@js.global "vscode.languages.getDiagnostics"]
 
-    val get_diagnostics_all : unit -> (Uri.t * Diagnostic.t list) list
+    val getDiagnostics_all : unit -> (Uri.t * Diagnostic.t list) list
       [@@js.global "vscode.languages.getDiagnostics"]]
 end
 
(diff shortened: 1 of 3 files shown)


mnxn:
I added support for the `relatedInformation` and `tags` fields as well as properly handling the record type for `code`.

mnxn (approved):

vscode/vscode/vscode.ml
```diff
@@ -1182,6 +1182,99 @@ module ProgressOptions = struct
...
+     | `Int of int
+     | `Targeted of code_target
+     ]
+    [@js.union])
+  [@@js]
+
+  let code_of_js js_val =
+    match Ojs.type_of js_val with
+    | "object" -> `Targeted ([%js.to: code_target] js_val)
+    | _ -> string_or_int_of_js js_val
```
ulugbekna, line 1247: doesn't it make sense to inline `string_or_int_of_js` unless we want to put it in a larger scope to be reusable?

ulugbekna:
btw, feel free to merge when you consider it ready

vscode/vscode/vscode.ml
mnxn: `string_or_int_of_js` is being used (by gen_js_api) for the `value` field of `code_target`.
So the function is already being reused locally.

mnxn: We could add these kind of types to the `interop` module, but I don't see the need for it yet.

PR #643, opened 2021-06-24
Jump to holes ocamllsp typed holes

ulugbekna:
Uses a custom request in ocamllsp to jump around holes. This should allow to eliminate too tight coupling with diagnostics, which caused problems when we want to jump to the first hole in a destructed code.

### DIFF
--- a/src/custom_requests.ml
+++ b/src/custom_requests.ml
@@ -1,0 +1,22 @@
+open Import
+
+let ocamllsp_prefixed s = "ocamllsp/" ^ s
+
+module Typed_holes : sig
+  (** [send_request client ~for_doc] will fetch ranges of holes in the file at
+      the URI [for_doc] *)
+  val send_request : LanguageClient.t -> for_doc:Uri.t -> Range.t list Promise.t
+end = struct
+  let meth = ocamllsp_prefixed "typedHoles"
+
+  let encode_params uri =
+    Jsonoo.Encode.(object_ [ ("uri", string @@ Uri.toString uri ()) ])
+
+  let decode_response json = Jsonoo.Decode.list Range.t_of_json json
+
+  let send_request client ~for_doc:uri =
+    let data = encode_params uri in
+    let open Promise.Syntax in
+    let+ response = LanguageClient.sendRequest client ~meth ~data () in
+    decode_response response
+end
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -112,10 +112,12 @@ let _open_current_dune_file =
     | None ->
       (* this command is available (in the command palette) only when a file is
          open *)
-      show_message `Error
-        "The command \"OCaml: Open Dune File\" should be run only with a file \
-         open in the editor, so the command can look for a dune file in the \
-         same folder as the open file."
+      show_message `Error "%s"
+      @@ Extension_consts.Command_errors.text_editor_must_be_active
+           "Open Dune File"
+           ~expl:
+             "The command can look for a dune file in the same folder as the \
+              open file."
     | Some text_editor ->
       let doc = TextEditor.document text_editor in
       let uri = TextDocument.uri doc in
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -133,6 +133,141 @@ let _open_current_dune_file =
   in
   command Extension_consts.Commands.open_current_dune_file handler
 
+let _jump_to_prev_hole, _jump_to_next_hole =
+  let pick_next_hole current_pos ~sorted_non_empty_holes_list:holes =
+    (* we want to find such a range that starts after the current position *)
+    match holes with
+    | [] -> assert false
+    | [ r ] -> r
+    | default :: _ ->
+      let next_hole =
+        List.find holes ~f:(fun range ->
+            match Position.compare current_pos (Range.start range) with
+            | Ordering.Less -> true
+            | Greater -> false
+            | Equal ->
+              (* we don't want the same range that we're in now; we need the
+                 next one *)
+              not
+                (Range.contains range ~positionOrRange:(`Position current_pos)))
+      in
+      (* if the current position is larger than all other ranges, we cycle back
+         to first hole in the file *)
+      Option.value next_hole ~default
+  in
+
+  let pick_prev_hole current_pos ~sorted_non_empty_holes_list:holes =
+    (* iterate through all hole ranges and pick the one that has start position
+       after the current cursor position *)
+    match holes with
+    | [] -> assert false
+    | [ r ] -> r
+    | fst_range :: rest ->
+      let start = Range.start fst_range in
+      if
+        Ordering.is_less (Position.compare current_pos start)
+        || Range.contains fst_range ~positionOrRange:(`Position current_pos)
+      then
+        (* [holes] is sorted from ranges in the beginning of the file to the
+           end; if the current position comes before or inside the first range,
+           then pick the last hole in file *)
+        List.last_exn rest
+      else
+        let rec find_prev_hole prev_range = function
+          | [] -> prev_range
+          | range :: rest -> (
+            if Range.contains range ~positionOrRange:(`Position current_pos)
+            then
+              prev_range
+            else
+              let start = Range.start range in
+              match Position.compare current_pos start with
+              | Ordering.Less -> prev_range
+              | Greater -> find_prev_hole range rest
+              | Equal ->
+                (* assert false because we check whether the current pos is in
+                   this range in the if-expr above *)
+                assert false)
+        in
+        find_prev_hole fst_range rest
+  in
+
+  (* We need to be able to specify the position from which we want to look for
+     the next typed hole.
+
+     If the args is empty, use the current position in the active text editor;
+     if given a specific position, use that position.
+
+     args := [] | [ { "position" : { "line": <int>; "character": <int> } } ] *)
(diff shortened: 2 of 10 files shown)


src/custom_requests.ml
```diff
@@ -0,0 +1,23 @@
...
+module Typed_holes : sig
+  (** [send_request client ~for_doc] will fetch ranges of holes in the file at
+      the URI [for_doc] *)
+  val send_request : LanguageClient.t -> for_doc:Uri.t -> Range.t list Promise.t
+end = struct
+  let meth = ocamllsp_prefixed "typedHoles"
+
+  let encode_params uri =
+    Jsonoo.Encode.(
+      object_ [ ("uri", Jsonoo.Encode.string @@ Uri.toString uri ()) ])
```
mnxn, line 14: You've already opened `Jsonoo.Encode` locally.

```suggestion
    Jsonoo.Encode.(object_ [ ("uri", string @@ Uri.toString uri ()) ])
```

src/import.ml
```diff
@@ -86,3 +93,47 @@ let with_confirmation message ~yes ?(no = "Cancel") f =
...
+(** Builds application-specific functionality around [Vscode.Position] *)
+module Position = struct
+  let compare p1 p2 =
+    let r = Position.compareTo p1 ~other:p2 in
+    if r < 0 then
+      Ordering.Less
+    else if r = 0 then
+      Equal
+    else
+      Greater
```
mnxn, line 106: I think `compare` should return an integer based on existing `compare` functions in the base library and this project.

It would also remove the need for `List.sort'`.

src/extension_commands.ml
```diff
@@ -131,6 +133,131 @@ let _open_current_dune_file =
   in
   command Extension_consts.Commands.open_current_dune_file handler
 
+let _jump_to_prev_hole, _jump_to_next_hole =
+  let pick_next_hole current_pos (`Sorted_non_empty (holes : Range.t list)) :
```
mnxn, line 137: I'm not sure how useful a `` `Sorted_non_empty`` tag is.
It's still possible to call it with an empty list: `` `Sorted_non_empty []``

A `non_empty_list` type would be typesafe, but might be too much for a few functions.

Maybe it would be better to have it as a label name? `~sorted_non_empty_holes`
It's not any more typesafe than the polymorphic variant, but it would have to be present at the calling site.

mnxn (changes requested):

src/import.ml
ulugbekna: I agree with your other comments, but let's keep this here: 
- it's better than dealings with ints
- it's the standard in ocamllsp; I'd hope we accept this idiom in this codebase as well

src/extension_commands.ml
```diff
@@ -131,6 +133,124 @@ let _open_current_dune_file =
   in
   command Extension_consts.Commands.open_current_dune_file handler
 
+let _jump_to_prev_hole, _jump_to_next_hole =
+  let pick_next_hole current_pos ~sorted_non_empty_holes_lst:holes =
```
mnxn, line 137: You can make this `~sorted_non_empty_holes_list`.

Omitting a single letter doesn't change much.

ulugbekna:
I have better things to do sorry

mnxn (approved):

tmattio:
> I have better things to do sorry

I can think of better ways to convey your point @ulugbekna 🙂 

Let's assume best intentions here: Max is a valuable maintainer of `vscode-ocaml-platform` and his reviews are equally valuable.
He took the time to review the PR, so let's respect his time as much as I'm sure he respects the time of other contributors.

In particular, his suggestion (I'm sure he didn't feel strongly about it and meant it more as a suggestion than a blocker) on the function naming follows previous guidelines of the codebase, which ensure that the code stays readable for newcomers.
An example of a codebase that does not follow similar guidelines is [Opam](https://github.com/ocaml/opam/blob/master/src/client/opamSwitchCommand.ml) which I personally find very hard to get into.
So, surely, there are some solid arguments that can be made on the importance of function and variable naming.

I don't have a specific opinion on the above, but if there is a disagreement on a review, let's keep it friendly and professional. We all have the same goal here.

ulugbekna:
I apologize to @mnxn if that sounded offensive. I assumed that his tone conveyed a suggesting character, which I turned down. Putting/omitting that character didn't change anything imho. I fully addressed the first round of comments, if necessary putting Max's suggestions/preferences above mine, e.g., using a named arg vs a type. I do, however, prefer to avoid bikeshed changes. I restrain myself from commenting on naming unless it would substantially make code easier to read. Nevertheless, I'm sorry if my comment above came off unprofessional. 

I very much appreciate Max's work and how supportive he is of this repo. For example, I recall offering to implement a feature (reason config option that is?) Max brought up in one of the meetings if he himself needed that feature because I do care. And I will surely be happy to review any of Max's PRs or contribute to them.

--- 

The PR should be ready but still dependent on https://github.com/ocaml/ocaml-lsp/pull/467. We should probably wait. I will comment in this PR when it's merged.

mnxn:
Thank you for your apology, but I had no hard feelings. I understand that even though we are communicating in English, there can still be language differences. I know that I've had difficulties communicating my points before. I was more confused by the comment than anything, so I appreciate your clarification.

src/import.ml
```diff
@@ -14,11 +14,22 @@ let property_exists json property =
 
 include Base
 
+module List = struct
+  let sort' lst ~compare =
+    List.sort lst ~compare:(fun a b -> compare a b |> Ordering.to_int)
```
rgrinberg, line 19: Why can't we use the original `List.sort`?

src/extension_commands.ml
```diff
@@ -131,6 +133,124 @@ let _open_current_dune_file =
...
+          | Equal ->
+            (* we don't want the same range that we're in now; we need the next
+               one *)
+            not (Range.contains range ~positionOrRange:(`Position current_pos)))
+    in
+    (* - if the current position is larger than all other ranges, we cycle back
+       to first hole in the file
+
+       - [hd_exn] because [holes] is non-empty *)
+    Option.value_or next_hole ~default:(fun () -> List.hd_exn holes)
```
rgrinberg, line 153: You could avoid the comment and the use of `hd_exn` by special casing the one element list:

```ocaml
let pick_next_hole current_pos ~sorted_non_empty_holes:holes =
  match holes with
  | [] -> assert false
  | [x] -> x
  | default :: _ -> let next_hole = ... in Option.value ~default holes
```

src/extension_commands.ml
```diff
@@ -131,6 +133,124 @@ let _open_current_dune_file =
...
+          match Position.compare current_pos start with
+          | Ordering.Less -> prev_range
+          | Greater -> find_prev_hole (Some range) rest
+          | Equal ->
+            (* assert false because we check whether the current pos is in this
+               range in the if-expr above *)
+            assert false)
+    in
+    Option.value_or (find_prev_hole None holes) ~default:(fun () ->
+        List.last_exn holes)
```
rgrinberg, line 175: Again, I think that avoiding `last_exn` is both cleaner and avoids extra assumptions:
```ocaml
let rec find_prev_hole prev_range = function
  | [] -> prev_range
  | [hole] -> hole
  ...
```

rgrinberg:
I only briefly glance through the code, but given that Max reviewed this, it should be good to go. @mnxn  did you try using this feature yet? We need to make sure the UI is solid too.

src/extension_commands.ml
ulugbekna: I think you misunderstood what the code is trying to do (or I misunderstood what you mean). We traverse all holes looking a hole such that its range comes after the current position. If there is no such hole, then we pick the first hole in the file, assuming that all holes came before the current position, ie cycle back to the first hole. Does this make sense? 

A similar algo used in `pick_prev_hole`.

rgrinberg: > Does this make sense?

It makes sense. I think the snippet above has this exact behavior. The full traversal of holes would still occur in the `...` of `let next_hole = ...`.

src/import.ml
ulugbekna: The `List.sort` that comes with Base uses an `int`-based compare, so I'm adding a sort that uses `Ordering.t`. 

Do you want me to shadow int-based `List.sort` of `Base`? (there are other `int`-compare-based functions in Base, which either need to be also shadowed or we should live with inconsistencies)

rgrinberg: I see. IMO we should live with the inconsistencies and just report them to the base bug tracker.

ulugbekna: I meant that if I shadow a single `List.sort` with our Ordering-based sort, all other Base functions would still use int-based compares. This is an inconsistency we create ourselves by deviating from Base's standard int-based compare. I can still do the shadowing.

re: traversal comments

I got what you meant and will fix them :-)

mnxn:
> did you try using this feature yet?

I have and it works well. I have not noticed any problems while using it.

ulugbekna:
Jumping around holes using `alt+l` (for next hole) and `alt+shift+l` (for previous hole)

https://user-images.githubusercontent.com/16353531/123912091-0957c180-d996-11eb-8247-e881ab0c29d3.mov

rgrinberg:
Do you want to clean up the history yourself or shall I just squash everything?

ulugbekna:
Right, I will do that manually.

Meanwhile, I don't know why tests are failing because of esy.

PR #666, opened 2021-07-19
AST exploration features for ppx developers. 

arozovyk:
Screenshot 2021-07-19 at 17 44 31
This PR adds a set of features aimed at simplifying the process of developing (and using) OCaml preprocessors.
Namely,

**Exploring the Abstract Syntax Tree**
- Live visualization of the OCaml AST inside a separate editor tab.
- The means of exploring the AST both ways:  from standard text editor, choosing a piece of code and revealing the corresponding AST node inside the custom UI; Highlighting the corresponding source code when mouse cursor is moved onto a node inside the UI.

This is done by using [Custom Editor API](https://code.visualstudio.com/api/extension-guides/custom-editors) that allows defining an arbitrary output for a certain type of files. Under the hood it uses the [Webview](https://code.visualstudio.com/api/extension-guides/webview), that can render (almost) any custom HTML, and it communicates with the extension via message passing.
 
For the UI itself, i adapted the popular open source option - [astexplorer.net](https://astexplorer.net/). Notably, besides other minor modifications (redefining the communications, adding a separate tab for preprocessed AST),  I reduced the application to only contain the `Tree` view, which is the AST output part.

**Working with preprocessed code and its AST.**
- Preview of the source code after it has been transformed by a preprocessor.
- A separate mode to navigate the transformed AST (keeping track of the original code).

Since modern ppxes are (and should be) developed using [ppxlib](https://github.com/ocaml-ppx/ppxlib), this PR introduces it as a dependency. I use ppxlib for almost all AST operations, and most importantly for its version of AST, because it's the one that will also be used by ppx developers.


rgrinberg:
Regarding the ast visualization component extracted from code mirror, do we plan to update it with patches from upstream? If so, we need to keep it as a fork and include it in this repository as a submodule.

arozovyk:
I don't think we need to do that since almost all of the recent updates (for the last year or so) to the the project are related to specific language features and the common core that we use remained pretty much the same.

ulugbekna:
A couple of quick notes from yesterday: 

- re: command name 

`OCaml: AST Preview` -> `OCaml: Preview AST` to align with other vscode commands where verb comes first

- ignored values should be type annotated like `let (_ : bool) = ...` 
- it'd be cool to have more contrast between background and checkboxes text

image

- iirc, the current shortcut for "reveal ast" uses `d`, which somewhat overlaps with `alt+d`, which is used for `Destruct`. I think it makes sense to (1) reserve use of `d` for destruct related functionality (2) use `alt+...` to follow convention of `alt+d`, `alt+o`, `alt+p` in the extension

- I assume that's already part of your plan, but having all the new functionality described (as a manual) in the README would be very helpful to the extension users 

- Question 1: we saw yesterday that one needs to rebuild the project to get updated AST explorer custom editor. How does it work with dune building in a watch mode? I have latency and usability in mind here. (I also wonder whether we need to come up with some term to denote this ast explorer custom editor, which will be used in the manual in the README and which should make it more precise/easier for us and users to discuss this custom editor in writing -- I'd propose "AST explorer")

Makefile
```diff
@@ -23,6 +23,10 @@ release-build:
 		--sources-content=false
 .PHONY: release-build
 
+view-build:
+	npm run --prefix astexplorer build
```
rgrinberg, line 27: Could you switch to `yarn` - we use it elsewhere in the file.

package.json
```diff
@@ -862,6 +906,18 @@
...
       }
+    ],
+    "customEditors": [
+      {
+        "viewType": "ast-editor",
+        "displayName": "OCaml AST preview",
+        "priority": "option",
+        "selector": [
+          {
+            "filenamePattern": "*.ml"
```
rgrinberg, line 917: I'm not familiar with this option, but why is `mli` excluded here?

src/ast_editor.ml
```diff
@@ -0,0 +1,549 @@
...
+
+let original_mode = ref true
+
+let hover_disposable_ref = ref None
+
+let doc_string_uri ~document = Uri.toString (TextDocument.uri document) ()
+
+let pp_doc_to_changed_origin_map = ref (Map.empty (module String))
+
+let origin_to_pp_doc_map = ref (Map.empty (module String))
```
rgrinberg, line 14: All of these global variables should be moved to a record. The record should then be attached to the global state of the plugin. This will guarantee that this is all cleaned up correctly once the plugin is unloaded.

src/ast_editor.ml
```diff
@@ -0,0 +1,549 @@
...
+  | None -> ()
+
+let read_html_file () =
+  let filename = Node.__dirname () ^ "/../astexplorer/dist/index.html" in
+  Fs.readFile filename
+
+let document_eq a b =
+  String.equal
+    (Uri.toString (TextDocument.uri a) ())
+    (Uri.toString (TextDocument.uri b) ())
```
rgrinberg, line 70: Can we add `Uri.equal` instead?

src/ast_editor.ml
```diff
@@ -0,0 +1,549 @@
...
+        let pp_code = get_reparsed_code_from_pp_file ~document in
+        let reparsed_signature =
+          pp_code |> Lexing.from_string |> Parse.interface
+        in
+        let reparsed_json =
+          Dumpast.reparse_signature signature reparsed_signature
+        in
+        reparsed_json
+    with
+    | _ -> null
```
rgrinberg, line 106: We should avoid doing a catch all on exceptions like this. Aren't there specific exceptions you can catch here?

Also, use the `match` to catche the exception:

```ocaml
match .. with
| Impl .. ->
| Intf .. ->
| exception _ -> null
```

src/ast_editor.ml
```diff
@@ -0,0 +1,549 @@
...
+  let astpair = object_ [ ("ast", origin_json); ("pp_ast", pp_value) ] in
+  send_msg "parse" (Jsonoo.t_to_js astpair) ~webview
+
+let onDidChangeTextDocument_listener event ~(document : TextDocument.t)
+    ~(webview : WebView.t) =
+  let changed_document = TextDocumentChangeEvent.document event in
+  if document_eq document changed_document then
+    transform_to_ast ~document ~webview
+  else
+    ()
```
rgrinberg, line 117: `else ()` does nothing and can be removed.

src/ppx_utils.ml
```diff
@@ -0,0 +1,51 @@
...
+      root ^ "/_build/default/"
+      ^ String.sub ~pos:0 ~len:(String.length relative - 2) relative
+      ^ "pp.ml"
+    | Signature ->
+      root ^ "/_build/default/"
+      ^ String.sub ~pos:0 ~len:(String.length relative - 3) relative
+      ^ "pp.mli"
+    | Unknown -> failwith "Unknown file extension"
+  with
+  | Failure errorMsg -> errorMsg
```
rgrinberg, line 39: who is raising `Failure` here? This error should never really be caught.

rgrinberg (changes requested):

rgrinberg:
We need @mnxn's help to review the bindings.

src/dumpast.mli
```diff
@@ -0,0 +1,9 @@
+val transform : string -> Ppx_utils.kind -> Jsonoo.t
+
+val from_structure : Parsetree.structure -> Jsonoo.t
+
+val reparse :
+  Parsetree.structure_item list -> Parsetree.structure_item list -> Jsonoo.t
+
+val reparse_signature :
+  Parsetree.signature_item list -> Parsetree.signature_item list -> Jsonoo.t
```
rgrinberg, line 9: Could you move this and other ppx related modules to a sub library.

arozovyk:
Thanks for excellent points @ulugbekna, 


> `OCaml: AST Preview` -> `OCaml: Preview AST` to align with other vscode commands where verb comes first

Updated by 2fc5ebd2f54cfe2a72e86b44670d2243f824cacf

> * ignored values should be type annotated like `let (_ : bool) = ...`

Done in 287e495baa84926fb5937c5e288a56ad8f73a865



> it'd be cool to have more contrast between background and checkboxes text

Fixed in  a289816878b3f0d8be5a951b4cb99ad0f7a9d8a9
Also remouved unused checkboxes in 1ed017891a9b6e1d90f498faea2a3834140ec8de
Screenshot 2021-07-28 at 18 03 36

> * iirc, the current shortcut for "reveal ast" uses `d`, which somewhat overlaps with `alt+d`, which is used for `Destruct`. I think it makes sense to (1) reserve use of `d` for destruct related functionality (2) use `alt+...` to follow convention of `alt+d`, `alt+o`, `alt+p` in the extension

Reserved Alt+N, for Node reveal and Alt+H for Hover mode in a26aa18df25d760d03160794924b62b30b50e95a :) (can't seem to use Mac thought. I think separate keybindings are needed for it)


> * I assume that's already part of your plan, but having all the new functionality described (as a manual) in the README would be very helpful to the extension users

Will do once you confirm the keybindings are correct. 

> * Question 1: we saw yesterday that one needs to rebuild the project to get updated AST explorer custom editor. How does it work with dune building in a watch mode? I have latency and usability in mind here. (I also wonder whether we need to come up with some term to denote this ast explorer custom editor, which will be used in the manual in the README and which should make it more precise/easier for us and users to discuss this custom editor in writing -- I'd propose "AST explorer")

Just tried it. It does seem to track AST explorer file changes, but if i'm not mistaken, dune rebuilds with `dune build` so **AST explorer** project is not affected (since again, it has to be built with npm/yarn, at least for now...). Even if there will be a need for building it aswell, the project is not heavy and it takes a fraction of a second (well, i do admit having poweful pc, but still). 

For the last point i do agree on "AST explorer" for future references

Makefile
arozovyk: Done in fdec035cf52bfbf50580da4b1ef9711eb838f40d

package.json
arozovyk: I missed it since I added `mli` support only lately. Surprisingly, I didn't break when opening AST explorer with: 
```Ocaml 
Vscode.Commands.executeCommand ~command:"vscode.openWith"
      ~args:
        [ Uri.t_to_js uri
        ; Ojs.string_to_js "ast-editor"
        ; ViewColumn.t_to_js ViewColumn.Beside
        ]
``` 
Added with 8066285d1e8a3a41ea89b6eb3f59712d074da8e0 so it's there.

src/ast_editor.ml
arozovyk: Added in 22f72393c5d7beb12713a2097a78bd25c5c8a828

src/dumpast.mli
arozovyk: Done in 59f461ea9985c9a23becf3b34f63d1e3e38ad7cd

src/ast_editor.ml
arozovyk: The only exception here was a `Failure` when reading binary file returned some kind of error. It has already been changed both in this function (d68aac57c09f3a65452aaa5ed6864eb1fad90ab3) and in [ppxlib#270](https://github.com/ocaml-ppx/ppxlib/pull/270). More changes are probably incoming though.

src/ppx_utils.ml
arozovyk: the `project_root_path` did in case where root wasn't found (when is that even possible?). Anyway, changed it to managing `string option`  returned by `Workspace.rootPath` in 3a57c06e207942b7bca2f7c88d0471cce3e70a5b

package.json
arozovyk: Oops, i somehow messed it up, it's actually in 2fc5ebd2f54cfe2a72e86b44670d2243f824cacf

vscode/vscode/vscode.ml
```diff
@@ -1898,6 +1920,48 @@ module DocumentFormattingEditProvider = struct
...
+  include Interface.Make ()
+
+  include
+    [%js:
+    val provideHover :
+         t
+      -> document:TextDocument.t
+      -> position:Position.t
+      -> token:CancellationToken.t
+      -> Hover.t list ProviderResult.t
```
mnxn, line 1952: Isn't this is supposed to return a single `Hover` value? There is no `[]` or `Array< >` in the TypeScript code

https://github.com/microsoft/vscode/blob/1.58.2/src/vs/vscode.d.ts#L2674

vscode/vscode/vscode.ml
```diff
@@ -2107,10 +2171,57 @@ module Progress = struct
...
+  include
+    [%js:
+    val onDidChange : t -> OnDidChange.t [@@js.get]
+
+    val provideTextDocumentContent :
+      t -> uri:Uri.t -> token:CancellationToken.t -> string ProviderResult.t
+      [@@js.call]
+
+    val create :
+         onDidChange:OnDidChange.t
```
mnxn, line 2212: Since `onDidChange` can be undefined, this should be `?onDidChange:OnDidChange.t`. Add a `unit` parameter at the end to support the optional argument.

vscode/vscode/vscode.ml
```diff
@@ -2140,9 +2254,26 @@ module Workspace = struct
...
+    val applyEdit : edit:WorkspaceEdit.t -> bool Promise.t
+      [@@js.global "vscode.workspace.applyEdit"]
+
+    val registerTextDocumentContentProvider :
+      scheme:string -> provider:TextDocumentContentProvider.t -> Disposable.t
+      [@@js.global "vscode.workspace.registerTextDocumentContentProvider"]
+
+    val asRelativePath :
+      pathOrUri:([ `String of string | `Uri of Uri.t ][@js.union]) -> string
+      [@@js.global "vscode.workspace.asRelativePath"]
```
mnxn, line 2275: Missing the `includeWorkspaceFolder` optional argument.

https://github.com/microsoft/vscode/blob/1.58.2/src/vs/vscode.d.ts#L10546

vscode/vscode/vscode.ml
```diff
@@ -2107,10 +2171,57 @@ module Progress = struct
...
+end
+
+module TextDocumentContentProvider = struct
+  include Interface.Make ()
+
+  module OnDidChange = Event.Make (Uri)
+
+  include
+    [%js:
+    val onDidChange : t -> OnDidChange.t [@@js.get]
```
mnxn, line 2205: Possible undefined. Should return `OnDidChange.t or_undefined`.

https://github.com/microsoft/vscode/blob/1.58.2/src/vs/vscode.d.ts#L1619

vscode/vscode/vscode.ml
```diff
@@ -2521,6 +2652,271 @@ module TreeView = struct
   end
 end
 
+module WebviewPanelOptions = struct
+  include Interface.Make ()
+
+  include
+    [%js:
+    val enableFindWidget : t -> bool [@@js.get]
```
mnxn, line 2660: Possible undefined. Should return `bool or_undefined`.

https://github.com/microsoft/vscode/blob/1.58.2/src/vs/vscode.d.ts#L7529

vscode/vscode/vscode.ml
```diff
@@ -2521,6 +2652,271 @@ module TreeView = struct
...
+
+    val retainContextWhenHidden : t -> bool [@@js.get]]
```
mnxn, line 2662: Possible undefined. Should return `bool or_undefined`.

https://github.com/microsoft/vscode/blob/1.58.2/src/vs/vscode.d.ts#L7547

vscode/vscode/vscode.ml
```diff
@@ -2521,6 +2652,271 @@ module TreeView = struct
...
+      ([%js.to: Uri.t] js_val) else if Ojs.has_property js_val "light" then
+      `LightDark ([%js.to: LightDarkIcon.t] js_val) else assert false*)
+
+    val active : t -> bool [@@js.get]
+
+    val options : t -> WebviewPanelOptions.t [@@js.get]
+
+    val title : t -> string [@@js.get]
+
+    val viewColumn : t -> ViewColumn.t [@@js.get]
```
mnxn, line 2835: Possibly undefined. Should return `ViewColumn.t or_undefined`.

https://github.com/microsoft/vscode/blob/1.58.2/src/vs/vscode.d.ts#L7583

vscode/vscode/vscode.ml
```diff
@@ -2521,6 +2652,271 @@ module TreeView = struct
...
+    -> dispose:Js.Any.t
+    -> reveal:(?preserveFocus:bool -> ?viewColumn:ViewColumn.t -> unit -> unit)
+    -> t
+end = struct
+  include Interface.Make ()
+
+  module LightDarkIcon = struct
+    type t =
+      { light : ([ `Uri of Uri.t ][@js.union])
+      ; dark : ([ `Uri of Uri.t ][@js.union])
```
mnxn, line 2796: Making these polymorphic variants seems unnecessary. Can't it just be `: Uri.t`? It would remove the need for the custom `t_of_js` function.

vscode/vscode/vscode.ml
```diff
@@ -2521,6 +2652,271 @@ module TreeView = struct
...
+
+and WebviewPanelOnDidChangeViewStateEvent : sig
+  include Js.T
+
+  val webviewPanel : t -> WebviewPanel.t
+end = struct
+  include Interface.Make ()
+
+  include [%js: val webviewPanel : t -> WebviewPanel.t [@@js.get]]
+end
```
mnxn, line 2876: If you place this module inside the `WebviewPanel` module, it would remove the need for using recursive modules.

src/ast_editor.ml
arozovyk: Did i do it right? 692b39a1d0a5c5fa7bad7c99864a1cf187309f71

.ocamlformat
```diff
@@ -1,4 +1,4 @@
-version=0.18.0
+version=0.19.0
```
rgrinberg, line 1: Is this change really needed? if it is, move it to a separate PR.

src/ast_editor.ml
```diff
@@ -0,0 +1,514 @@
...
+    Vscode.Commands.executeCommand ~command:"vscode.openWith"
+      ~args:
+        [ Uri.t_to_js uri
+        ; Ojs.string_to_js "ast-editor"
+        ; ViewColumn.t_to_js ViewColumn.Beside
+        ]
+  in
+  ()
+
+let replace_document_content ~document ~content =
```
rgrinberg, line 202: I don't see the return value of this function used anywhere. Therefore it should return just return unit

src/ast_editor.ml
```diff
@@ -0,0 +1,514 @@
...
+        ^ Ppx_tools.Pp_path.relative_document_path ~document
+        ^ "' haven't been preprocessed yet.")
+      ~choices:[ ("Run `dune build`", 0); ("Abandon", 1) ]
+      ()
+  in
+  manage_choice instance choice ~document
+
+let open_preprocessed_doc_to_the_side instance ~document =
+  try open_pp_doc instance ~document with
+  | _ -> manage_open_failure instance ~document
```
rgrinberg, line 321: Again, we shouldn't be catching all exceptions.

src/ast_editor_state.mli
```diff
@@ -0,0 +1,39 @@
...
+
+val set_original_mode : t -> bool -> unit
+
+val get_hover_disposable : t -> Disposable.t option
+
+val set_hover_disposable : t -> Disposable.t option -> unit
+
+val set_origin_changed : t -> data:bool -> key:string -> unit
+
+val entry_exists : t -> string -> string -> bool
```
rgrinberg, line 24: The arguments for this function need labels.

src/ast_editor.ml
rgrinberg: Yes, it looks much better.

src/ast_editor.ml
arozovyk: Changed to catching `Sys_error` when file is not found in b6f514d1400b2464909d6a0657899bf10b6cbcc5

vscode/vscode/vscode.ml
arozovyk: Is it ok to have `WebviewPanelOnDidChangeViewStateEvent .webviewPanel` have signature `: t -> t`? 
https://github.com/ocamllabs/vscode-ocaml-platform/commit/ef4ee16416ee5e3430cfcb0e8015b853895750a4#diff-f66b1e1ddae54b2c611b3c724af7f1c92c192c722a2bb03ea746cab601b9553bR2739,
Or should it be something else?

.ocamlformat
arozovyk: #679

vscode/vscode/vscode.ml
```diff
@@ -818,6 +820,26 @@ module WorkspaceConfiguration = struct
...
 end
 
+module WorkspaceEdit = struct
+  include Interface.Make ()
+
+  include
+    [%js:
+    val size : t -> int [@@js.get]
+
+    val set_size : t -> int -> unit [@@js.set]
```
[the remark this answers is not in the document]
arozovyk: Done in 9dc2a15300d90a829591139b1d25c3545f6e9567

vscode/vscode/vscode.ml
```diff
@@ -818,6 +820,26 @@ module WorkspaceConfiguration = struct
     [%js.to: T.t] (get_default this ~section ~defaultValue)
 end
 
+module WorkspaceEdit = struct
+  include Interface.Make ()
```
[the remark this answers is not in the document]
arozovyk: Done in 39fa91eb61020b3081f6b453415dd71092cea498

vscode/vscode/vscode.ml
```diff
@@ -1898,6 +1920,48 @@ module DocumentFormattingEditProvider = struct
...
 end
 
+module Hover = struct
+  include Interface.Make ()
+
+  include
+    [%js:
+    val contents : t -> MarkdownString.t [@@js.get]
+
+    val range : t -> Range.t [@@js.get]
```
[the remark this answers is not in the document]
arozovyk: Done in 1f0c59a4408f7e4244f1bb27961ce2863d372e53

vscode/vscode/vscode.ml
arozovyk: Done in 84ec3948dc80307b424fa225aebf385e1fe7e03e

vscode/vscode/vscode.ml
arozovyk: Done in 3c87042314c6be0e944087cc3cfc14c13ef9b642

vscode/vscode/vscode.ml
arozovyk: Done in db5b8077e4f4e909438c050c4b4b8813884fc6f9

vscode/vscode/vscode.ml
arozovyk: Also done in db5b8077e4f4e909438c050c4b4b8813884fc6f9

vscode/vscode/vscode.ml
arozovyk: Done in af072b6330ac64e439013b2f099033607d85d9f5

vscode/vscode/vscode.ml
arozovyk: Also done in af072b6330ac64e439013b2f099033607d85d9f5

vscode/vscode/vscode.ml
arozovyk: Done in b0f217bbb0a6f8a866299ee40e4122dee1f2a1dd

vscode/vscode/vscode.ml
arozovyk: It is indeed unnecessary, fixed in 7efe612f6bc6db5f6773d5b4fa8fdc0c089bce49.

src/ast_editor.ml
arozovyk: Changed in d17892794eb016a5a250c82484080217ed54c802

src/ast_editor_state.mli
arozovyk: Done in 001f9964951a21cccebf25092f417f31be9f28a7

vscode/vscode/vscode.ml
arozovyk: Thanks, 
I also had to add `[@@js]` to type alias 
 https://github.com/ocamllabs/vscode-ocaml-platform/pull/666/commits/5844d1b6545e2a15cdef8839aa9540638337036f#diff-f66b1e1ddae54b2c611b3c724af7f1c92c192c722a2bb03ea746cab601b9553bR2740 
Is this correct? 
Or should i add `let webviewPanel_of_js = t_of_js` right after it?

mnxn: Oops, forget about that.
The `[@@js]` is fine.

src/ast_editor_state.ml
```diff
@@ -0,0 +1,104 @@
...
+  ; original_mode
+  ; hover_disposable
+  ; pp_doc_to_changed_origin_map
+  ; origin_to_pp_doc_map
+  }
+
+let find_original_doc_by_pp_uri ~uri_string map =
+  let r =
+    Map.filteri ~f:(fun ~key:_ ~data -> String.equal uri_string data) map
+    |> Map.to_alist |> List.unzip |> fst
```
rgrinberg, line 29: What about `Map.find_first`?

Note: `find_first` is the name used in the stdlib, it could be different in base.

src/ast_editor_state.ml
```diff
@@ -0,0 +1,104 @@
...
+  in
+  match r with
+  | k :: _ -> Some k
+  | _ -> None
+
+let find_webview_by_doc t ~document =
+  let doc_uri = Uri.toString (TextDocument.uri document) () in
```
rgrinberg, line 36: Should be `to_string`.

src/ppx_tools/dumpast.mli
```diff
@@ -0,0 +1,9 @@
+val transform : string -> Pp_path.kind -> Jsonoo.t
```
rgrinberg, line 1: This mli needs a top level ocamldoc comment. ocamldoc comments for the individual functions would be welcome as well.

src/ppx_tools/dumpast.ml
```diff
@@ -0,0 +1,397 @@
...
+    match kind with
+    | Structure ->
+      let v = Parse.implementation (Lexing.from_string source) in
+      parse_ast#structure v
+    | Signature ->
+      let v = Parse.interface (Lexing.from_string source) in
+      parse_ast#signature v
+    | Unknown -> Jsonoo.Encode.string "Unknown file extension"
+  with
+  | _ -> Jsonoo.Encode.string "Syntax error"
```
rgrinberg, line 389: Can we stick to a concrete exception here?

src/ppx_tools/dumpast.ml
```diff
@@ -0,0 +1,397 @@
...
+
+let from_structure (structure : Parsetree.structure) =
+  try parse_ast#structure structure with
+  | _ -> Jsonoo.Encode.string "Syntax error"
```
rgrinberg, line 393: And here as well?

src/ppx_tools/dumpast.ml
```diff
@@ -0,0 +1,396 @@
...
+      (*FIXME*)
+      self#record "open_infos"
+        [ ("popen_expr", popen_expr)
+        ; ("popen_override", popen_override)
+        ; ("popen_loc", popen_loc)
+        ; ("popen_attributes", popen_attributes)
+        ]
+
+    method! directive_argument_desc x x' =
+      try super#directive_argument_desc x x' with
```
rgrinberg, line 172: Could you extract this try/catch pattern into a function:

```ocaml
val with_reparse : (unit -> unit) -> on_reparse_error:(unit -> unit)
```

src/ast_editor_state.ml
```diff
@@ -0,0 +1,103 @@
+open Import
+
+type t =
+  { mutable webview_map : (string, WebView.t, String.comparator_witness) Map.t
```
rgrinberg, line 4: These fields deserve comments.

src/ppx_tools/traverse_ast2.ml
```diff
@@ -0,0 +1,1936 @@
+open Ppxlib
+
+exception Reparse_error of string
```
rgrinberg, line 3: This exception needs a comment describing the error and the argument.

src/ast_editor.ml
```diff
@@ -0,0 +1,487 @@
...
+let resolveCustomTextEditor instance ~(document : TextDocument.t) ~webviewPanel
+    ~token:_ : CustomTextEditorProvider.ResolvedEditor.t =
+  let webview = WebviewPanel.webview webviewPanel in
+  (*persist the webview*)
+  let ast_editor_state = Extension_instance.ast_editor_state instance in
+  Ast_editor_state.set_webview_map ast_editor_state
+    (Map.set
+       (Ast_editor_state.get_webview_map ast_editor_state)
+       ~key:(doc_string_uri ~document) ~data:webview);
+  let (_ : Disposable.t) =
```
rgrinberg, line 138: Why is it ok to ignore this disposable? Shouldn't it be disposed when the plugin is disposed of?

src/ast_editor.ml
```diff
@@ -0,0 +1,487 @@
...
+    in
+    HoverProvider.create ~provideHover
+  in
+  Vscode.Languages.registerHoverProvider ~selector:(`String "ocaml") ~provider
+
+let activate_hover_mode instance ~document =
+  let ast_editor_state = Extension_instance.ast_editor_state instance in
+  match Ast_editor_state.find_webview_by_doc ast_editor_state ~document with
+  | Some webview -> on_hover document webview
+  | None -> failwith "Webview wasn't found while switching hover mode"
```
rgrinberg, line 127: Under what circumstances is this error possible?

src/ast_editor.ml
```diff
@@ -0,0 +1,487 @@
...
+        then
+          reload_pp_doc
+        else
+          open_pp_doc)
+          instance ~document
+      else
+        let* perror =
+          Window.showErrorMessage
+            ~message:"Building project failed, fix project errors and retry."
+            ~choices:[ ("Retry running `dune build`", 0); ("Abandon", 1) ]
```
rgrinberg, line 270: Instead of using integers, can we use polymorphic variant values for the different choices?

src/ast_editor.ml
```diff
@@ -0,0 +1,487 @@
...
+      let* res = build_cmd () in
+      if res.exitCode = 0 then
+        (if
+         Map.existsi
+           (Ast_editor_state.get_pp_doc_to_changed_origin_map ast_editor_state)
+           ~f:(fun ~key ~data:_ -> String.equal key (doc_string_uri ~document))
+        then
+          reload_pp_doc
+        else
+          open_pp_doc)
```
rgrinberg, line 264: Why aren't we handling Sys_error here?

src/ast_editor_state.ml
arozovyk: I can't seem to find another way to search keys by data in `Base`. I sould've probably chosen some other structure like bidirectional map, but I'm not sure it's really needed here since it's the only time i search "the other way around".

src/ast_editor_state.ml
arozovyk: Do you mean changing the name of https://github.com/microsoft/vscode/blob/c3f126316369cd610563c75b1b1725e0679adfb3/src/vs/vscode.d.ts#L1413 ?

src/ast_editor_state.ml
ulugbekna: We could do `Map.filter map ~f:predicate_on_value |> Map.keys`?

src/ppx_tools/dumpast.ml
arozovyk: Changed in 9d311ed4ae0b788b2aa4dbf1ad4d67b6c07fb354 (not doing anything about the exception since users know already that there's syntax error via their editor)

src/ast_editor_state.ml
arozovyk: Right, thanks! I didn't process the fact that it operated on values for some reason. It does look cleaner 0b7301348fc95c6823040df24de11c7bd01ff2bb

src/ast_editor.ml
arozovyk: Actually I'm not sure what's the logic behind transforming the
```typescript
Event<void>
```
https://github.com/microsoft/vscode/blob/c3f126316369cd610563c75b1b1725e0679adfb3/src/vs/vscode.d.ts#L7608

to 
```OCaml 
Disposable.t 
```
But if we follow the similar [example](https://github.com/ocamllabs/vscode-ocaml-platform/blob/175a87213f9b024258dcffd4d7f6a6c3e89cbc59/src/repl.ml#L184) in `Repl` module, then we shouldn't ignore it. I did it just in case 6993d2647ebd03c9416e1086e0d593e02155a42f

src/ast_editor.ml
arozovyk: It happends when user tries to activate hover mode without opening the AST explorer for instance. 
I added an error message to clarify this 69aae11657ecddc8973ccd56ab8dfcf1a307fe4f .

src/ast_editor.ml
arozovyk: We should indeed handle it. a804ac893a2a1db146d988787ce8caaf35b7a333

src/ast_editor.ml
arozovyk: 395d876e4c5f1b797623894fe507e6e549a6ed9b

src/ast_editor_state.ml
rgrinberg: Oh sorry - never mind. I thought Uri was our own module.

rgrinberg:
After reviewing the workflow again, I think we should get rid of the dune build prompt. The way it suggests to run dune is flawed and isn't appropriate for all workflows. After erroring out we should suggest tell the user to build the project, but without doing it ourself.

src/ast_editor.ml
```diff
@@ -0,0 +1,492 @@
...
+  let origin_json =
+    Dumpast.transform
+      (TextDocument.getText document ())
+      (Pp_path.get_kind ~document)
+  in
+  let pp_value =
+    match Pp_path.get_pp_path ~document with
+    | None ->
+      show_message `Error "%s" "project root path wasn't found";
+      Jsonoo.Encode.null
```
rgrinberg, line 36: I find that returning `null` is suspicious. As far as I understand, this is a fatal error that signals that we can't proceed. To me, that means the right course of action is just to raise and abort the operation. Why do we need a return value at all?

src/ppx_tools/dumpast.ml
rgrinberg: Yeah, pretty much. I pushed a commit that does this and it does clean things up a little.

src/ppx_tools/dumpast.ml
```diff
@@ -0,0 +1,389 @@
...
+    match kind with
+    | Structure ->
+      let v = Parse.implementation (Lexing.from_string source) in
+      parse_ast#structure v
+    | Signature ->
+      let v = Parse.interface (Lexing.from_string source) in
+      parse_ast#signature v
+    | Unknown -> Jsonoo.Encode.string "Unknown file extension"
+  with
+  | Syntaxerr.Error _ -> Jsonoo.Encode.null
```
rgrinberg, line 381: Perhaps we need a wrapper around the parse functions that returns a result type? So that we don't forget checking this error. But to me, this looks like another case where we should just abort and let the error bubble up. The uninformative null is suspicious again.

src/ppx_tools/traverse_ast2.ml
rgrinberg: Thanks. This comment is quite helpful. So when this error occurs, we just abort and let it bubble up, right? i.e. there's no way to recover from this error.

PS You might want to review the syntax for [ocamldoc comments](https://ocamlverse.github.io/content/documentation_guidelines.html). It has a different convention for code, lists, etc.

arozovyk: On the contrary, we use one argument lift class to continue serializing the subtree that starts where the error occurs. The  subtree will simply not contain double locations (but the ones generated by ppx) which means that simultaneous navigation will not work for that subtree. 

It looks like the following (using `with_reparse` we saw earlier) : 
```OCaml 
method! directive_argument_desc x x' =
      with_reparse
        (fun () -> super#directive_argument_desc x x')
        ~on_reparse_error:(fun () -> parse_ast#directive_argument_desc x)
``` 

I'm reading about odoc, will add comments everywhere soon.

src/ppx_tools/dumpast.ml
arozovyk: I don't think your commit went through for some reason. Could you please recheck?

rgrinberg: Oops. Done.

src/ppx_tools/traverse_ast2.ml
rgrinberg: I see. In which case, this exception cannot be observed by callers then? Shall we remove it from the mli then?

src/ppx_tools/dumpast.ml
```diff
@@ -0,0 +1,362 @@
...
+      Jsonoo.Encode.string "serializing int64 is not supported yet"
+
+    method int32 _ _ =
+      Jsonoo.Encode.string "serializing int32 is not supported yet"
+
+    method int value _ = Jsonoo.Encode.int value
+
+    method list f l l' =
+      (*FIXME different length?*)
+      if List.length l > List.length l' then
```
rgrinberg, line 113: Could you clarify the FIXME?

Also, there's no need to compute the lengths of `l` and `l'` twice.

rgrinberg:
Regarding error handling, I suggest we include an exception:

```ocaml
User_error of string
```

This error should be caught and translated to an appropriate error message for the user. If there's additional information required, it can be written to the logs. This should simplify a lot of of code and get rid of some useless return values.

src/ppx_tools/traverse_ast2.ml
arozovyk: Pretty much, yes.

arozovyk: Wait, actually no. It has to be visible at least in the dumpast module, I don't think we can remove it from the mli.

rgrinberg: In that case, I would suggest the function to return `Result.t`. If the exception needs to be handled, it should just be reflected in the signature.

rgrinberg:
Why do we need Traverse_ast and Traverse_ast2 at all? As far as I can tell, they are only used once in dumpast. Is there any issue with inlining them?

src/ppx_tools/dumpast.ml
arozovyk: The issue was that reparsing sometimes gave additional (or less, I don't recall) information making lists of different sizes. It is resolved now by always considering the longest list. (So the FIXME still being here is an oversight on my part)

arozovyk:
> Why do we need Traverse_ast and Traverse_ast2 at all? As far as I can tell, they are only used once in dumpast. Is there any issue with inlining them?

I created separate modules just to not clutter (visually) dumpast module with 3k additional lines of code. If there's any good reason to put them into dumpast instead of having two modules - there is no issue. 
That being said, I'm hoping to upstream those `lift` and `lift2` functions to `ppxlib` one day to be derived for each new AST version.

src/ast_editor.ml
arozovyk: It is here because of the fact that I'm trying to send both AST (of the document, and the preprocessed one) to the Webview. If getting the preprocessed one resulted in an error, I still want to send the non preprocessed AST. 

I agree that it has to be changed. I will try to define 2 separate messages exchanges for both AST, and also only send the preprocessed one when the corresponding tab is chosen. There are also adjustments to do to update both AST on document change.

src/ppx_tools/dumpast.ml
arozovyk: Wrapped in result. 70503e6382b3377d2040697f280b4a9028039def
I think it would be fine to also indicate syntax error inside the AST explorer (at the moment nothing happends).

src/ppx_tools/traverse_ast2.ml
arozovyk: Changed the return type and removed exception fca04409f9d3371d4b799534253d42e1cdbd5631. 
Not sure I can avoid the fact that all `lift2`method return `('res, string) result`. Because this way i have to unwrap some elements with `Result.get_ok` which sould never cause an exception but it's still awkward.

rgrinberg:
> I created separate modules just to not clutter (visually) dumpast module with 3k additional lines of code. If there's any good reason to put them into dumpast instead of having two modules - there is no issue.
That being said, I'm hoping to upstream those lift and lift2 functions to ppxlib one day to be derived for each new AST version.

Okay, I don't mind if it's left how it is now.

src/ast_editor_state.ml
```diff
@@ -0,0 +1,109 @@
...
+        (* Indicates current mode inside AST explorer (true = Original, false =
+           Prerocessed) *)
+  ; mutable original_mode : bool
+        (* Contains Some value when hover mode is enabled, None otherwise *)
+  ; mutable hover_disposable : Disposable.t option
+        (* Mapping between string value of Uri.t of the original (non
+           preprocessed) document to boolean, indicating that the document has
+           changed since opening its Preprocessed Document *)
+  ; mutable pp_doc_to_changed_origin_map :
+      (string, bool, String.comparator_witness) Map.t
```
rgrinberg, line 16: Does this actually need to be a map? It seems like we only ever care about `false` values. Might as well make it a set then.

src/ast_editor.ml
```diff
@@ -0,0 +1,532 @@
...
+  | Ok x -> Promise.return x
+  | Error e -> manage_open_failure e instance ~document
+
+let open_both_ppx_ast instance ~document =
+  let open Promise.Syntax in
+  let* pp_doc_open = open_preprocessed_doc_to_the_side instance ~document in
+  if pp_doc_open = 0 then
+    open_ast_explorer ~uri:(TextDocument.uri document)
+  else
+    Promise.make (fun ~resolve:_ ~reject:_ ->
```
rgrinberg, line 344: Why do we need this `Promise.make` wrapper? It should be removed here (and elsewhere).

src/ast_editor_state.mli
```diff
@@ -0,0 +1,34 @@
...
+
+type t
+
+val make : unit -> t
+
+val find_original_doc_by_pp_uri : t -> Uri.t -> string option
+
+val find_webview_by_doc : t -> Vscode.Uri.t -> WebView.t option
+
+val set_changes_tracking : t -> TextDocument.t -> TextDocument.t -> unit
```
rgrinberg, line 14: I think we should change the API to take `Uri.t` everywhere. We only need `Uri.t` in all these functions so let's ask for only what we need.

rgrinberg:
Looking much better now! Just remove all those integer error codes in favor of result types and this should be good to go.

arozovyk:
> Regarding error handling, I suggest we include an exception:
> 
> ```ocaml
> User_error of string
> ```
> 
> This error should be caught and translated to an appropriate error message for the user. If there's additional information required, it can be written to the logs. This should simplify a lot of of code and get rid of some useless return values.

Do you still think it's a good idea to define this exception? If yes, what errors should it cover?

rgrinberg:
> Do you still think it's a good idea to define this exception? If yes, what errors should it cover?

Yes, I still think it's a goo idea. It should cover all errors caused by a misconfigured environment or bad user input. It should essentially result in an error message displayed to the user. The only problem is that I'm not sure where is the right place to install the uncaught exception handler to handle it.

arozovyk:
> > Do you still think it's a good idea to define this exception? If yes, what errors should it cover?
> 
> Yes, I still think it's a goo idea. It should cover all errors caused by a misconfigured environment or bad user input. It should essentially result in an error message displayed to the user. The only problem is that I'm not sure where is the right place to install the uncaught exception handler to handle it.

First of all, there are 2 ways I would like to output errors : 

1. Errors related to AST explorer, like syntax error, or Preprocessed mode not being available are shown in the AST explorer itself. (These types of errors are sent as a message to the `Webview`)  
2. Everything else (project updates, file not found on opening _Preprocessed Document_, etc.)  is a vscode prompt. 

(So 2 separate exceptions are needed?)

And in terms of the handler installation, I don't see any other way than handling it in every single command and listener. (Is there a common caller to all of that? I'm not sure)

Because this way https://github.com/arozovyk/vscode-ocaml-platform/commit/c4370cd689e29e1d886018f0bedda1e8558635df#diff-30cd2e06848ea70119bca78e584ea32e87f411b5203717096daba42410427c72 I don't feel like it simplifyies that much.

rgrinberg:
> So 2 separate exceptions are needed?

For 1. in your list, we need an exception. For 2., you should use whatever you find more convenient. So either a new exception, or result types.

src/ast_editor.ml
```diff
@@ -0,0 +1,578 @@
...
+    in
+    Ast_editor_state.associate_origin_and_pp ast_editor_state
+      ~origin_uri:(TextDocument.uri document)
+      ~pp_doc_uri:(TextDocument.uri doc);
+    replace_document_content ~content:pp_pp_str ~document:doc;
+    let+ (_ : TextEditor.t) =
+      Window.showTextDocument ~document:(`TextDocument doc)
+        ~column:ViewColumn.Beside ()
+    in
+    Ok 0
```
rgrinberg, line 302: Why 0 and not `()`?

src/ast_editor.ml
```diff
@@ -0,0 +1,578 @@
...
+                | Impl ppml_structure ->
+                  let reparsed_structure = Parse.implementation lex in
+                  Dumpast.reparse ppml_structure reparsed_structure
+                | Intf signature ->
+                  let reparsed_signature = Parse.interface lex in
+                  Dumpast.reparse_signature signature reparsed_signature)
+            in
+            make_value (Ok pp_json))
+      with
+      | Sys_error e -> raise (User_error e)
```
rgrinberg, line 120: Can we localize catching this `Sys_error`? It's not clear who is actually throwing this.

arozovyk: The `Sys_error` here occurs when you give inexistent path to `Ppxlib.Ast_io.read_binary`. What do you mean by localizing it, like print the call stack or maybe catch it earlier and add a clarifying message?

rgrinberg: I pushed a commit that made `Sys_error ` part of the result types. This makes it more obvious to see where this error is coming from.

src/ast_editor.ml
```diff
@@ -0,0 +1,585 @@
...
+          raise (User_error err_msg)
+        | Ok res ->
+          let pp_json =
+            let pp_code =
+              match fetch_pp_code ~document with
+              | Ok s -> s
+              | Error m -> raise (User_error m)
+            in
+            let lex = Lexing.from_string pp_code in
+            Result.ok_or_failwith
```
rgrinberg, line 117: Why is this `failwith`? Isn't this an error that should be displayed in the WebView? (parsing error of sorts).

src/ast_editor.ml
```diff
@@ -0,0 +1,585 @@
...
+let activate_hover_mode instance ~document =
+  let ast_editor_state = Extension_instance.ast_editor_state instance in
+  match
+    Ast_editor_state.find_webview_by_doc ast_editor_state
+      (TextDocument.uri document)
+  with
+  | Some webview -> on_hover document webview
+  | None ->
+    show_message `Error "Webview wasn't found while switching hover mode";
+    failwith "Webview wasn't found while switching hover mode"
```
rgrinberg, line 220: This is a fatal error that the user isn't responsible for and we cannot recover from, right? In that case we should introduce an exception type called `Code_error`. In general, we shouldn't be using `failwith` at all.

src/ast_editor.ml
```diff
@@ -0,0 +1,585 @@
...
+  let open Promise.Syntax in
+  let ast_editor_state = Extension_instance.ast_editor_state instance in
+  let visibleTextEditors = Window.visibleTextEditors () in
+  let origin_uri =
+    match
+      Ast_editor_state.find_original_doc_by_pp_uri ast_editor_state
+        (TextDocument.uri document)
+    with
+    | Some x -> x
+    | None -> failwith "Failed finding the original document URI."
```
rgrinberg, line 320: Again, this exception needs to be something else:

* User_error if the user is responsible or is part of the normal workflow
* Code_error if it's a bug in our extension or we don't know what happened.

src/ast_editor.ml
```diff
@@ -0,0 +1,585 @@
...
+      | Some path -> (
+        match Ppx_tools.get_preprocessed_ast path with
+        | Error (Io_error err_msg)
+        | Error (Read_error err_msg) ->
+          raise (User_error err_msg)
+        | Ok res ->
+          let pp_json =
+            let pp_code =
+              match fetch_pp_code ~document with
+              | Ok s -> s
```
rgrinberg, line 113: When `s = ""` as a result of not being able to parse the file, why are we trying to parse it again?

I think `fetch_pp_code` should return a more compound type and let the callers report errors.
```ocaml
val fetch_pp_code : TextDocument.t -> (string, Ppx_tools.read_error) result
```

This will also prevent some errors being displayed both in the webview and in the popup.

rgrinberg:
Final complaint about the UI:

The Retry/Abandon dialog seems somewhat suboptimal to me. If we had a refresh button/keybinding/action inside the web view that allowed us to refresh the view, that would be more than sufficient. This is less intrusive as it doesn't force the user's attention on this problem, and it also allows the user to restore the webview even if the menu was accidentally dismissed.

arozovyk (commented):
I'm changing this 9ba660b22f4b5410cca791307e6391b527d8735c to a result type since we don't know where to catch exceptions in an elegant way. ( `transform_to_ast` being a different story).

src/ast_editor.ml
arozovyk: This is not a fatal error, it just means that user tried to activate hover mode without opening an AST editor. The `failwith` is useless though 8e2c3055024505f21f2bea2b707ef23ee43cf61b

src/ast_editor.ml
arozovyk: Changed it to managing `Error` (which should never occur since all of the `Traverse_ast2.lift2`methods returning it are accounted for by `with_reparse` (except if I forgot something).

rgrinberg:
> I'm changing this 9ba660b to a result type since we don't know where to catch exceptions in an elegant way. ( transform_to_ast being a different story).

Sounds good

src/ast_editor.ml
arozovyk: Again, replaced `failwith` with just an `Error`. 064c5ab58fc14656ef46f80968decdc490acdfae

src/ast_editor.ml
arozovyk: 7a6c155dc2f8adae4940884cb1eb027e6a980815

rgrinberg: It seems like the `read_error` type isn't useful anymore. As far as I can tell, `Io_error` and `Read_error` are now handled exactly the same way. So we can drop the distinction if that's intended.

rgrinberg:
Let me  clarify the general strategy behind error handling. In hindsight, I'm giving out concrete review feedback that is piecemeal. I can see how it might be hard to understand what is the end state that I'm looking for.

1. Whenever we want to abort and give control back to the user, we should use `User_error`. The message in this exception should clear. Thus including prefixes like `"Read_error: "` isn't good because it is making it cryptic for the user. If something is invalid (e.g. filename), we should include the invalid data in the error message.

2. Whenever there's an error condition that we must handle differently depending on the caller, we should prefer result types (if possible). The type signature makes it easy to remember that the error case must be explicitly handled.

3. If our error handling code is different based on the kind of error, we should introduce a type to represent the various error conditions we intend to handle. If our error handling code is the same for different errors, this excessive level of detail is confusing and creates opportunities for bugs. We aren't writing library code, so we don't need to write fully general code.

Some consequences of the above are:

1. We never use `failwith`. Because it is not clear whether the error message is a bug in our code or intended for the user.

2. We should usually not catch `User_error` unless we are in the top level handler. As this error signifies that whatever operation that is being done is aborted.

Let me know if you have any questions. Thanks for being patient and polishing the code.

arozovyk:
> Final complaint about the UI:
> 
> The Retry/Abandon dialog seems somewhat suboptimal to me. If we had a refresh button/keybinding/action inside the web view that allowed us to refresh the view, that would be more than sufficient. This is less intrusive as it doesn't force the user's attention on this problem, and it also allows the user to restore the webview even if the menu was accidentally dismissed.

I agree that dialog shouldn't be the only way to update the Webview, so i added a button for `Preprocessed` mode. I could remove the prompt but It would mean either 

1. User has to close and reopen the Preprocessed Document to update it. 
2. We gotta have an additional command to reload it.

arozovyk:
> Let me clarify the general strategy behind error handling. In hindsight, I'm giving out concrete review feedback that is piecemeal. I can see how it might be hard to understand what is the end state that I'm looking for.
> 
> 1. Whenever we want to abort and give control back to the user, we should use `User_error`. The message in this exception should clear. Thus including prefixes like `"Read_error: "` isn't good because it is making it cryptic for the user. If something is invalid (e.g. filename), we should include the invalid data in the error message.
> 2. Whenever there's an error condition that we must handle differently depending on the caller, we should prefer result types (if possible). The type signature makes it easy to remember that the error case must be explicitly handled.
> 3. If our error handling code is different based on the kind of error, we should introduce a type to represent the various error conditions we intend to handle. If our error handling code is the same for different errors, this excessive level of detail is confusing and creates opportunities for bugs. We aren't writing library code, so we don't need to write fully general code.
> 
> Some consequences of the above are:
> 
> 1. We never use `failwith`. Because it is not clear whether the error message is a bug in our code or intended for the user.
> 2. We should usually not catch `User_error` unless we are in the top level handler. As this error signifies that whatever operation that is being done is aborted.
> 
> Let me know if you have any questions. Thanks for being patient and polishing the code.

Thanks for taking your time and helping me figure this out.

I agree with the points you are making, the only thing that seems kinda wierd is why we don't catch `User_error`. Because if we don't, the user is never informed about it (unless he opens Dev tools). Is that what we are looking for?

rgrinberg:
> I agree with the points you are making, the only thing that seems kinda wierd is why we don't catch User_error. Because if we don't, the user is never informed about it (unless he opens Dev tools). Is that what we are looking for?

I was looking for a good way to handle these errors, but it looks like it is currently not possible to install a handler in one place: https://github.com/microsoft/vscode/issues/45264

What we can do is a workaround like in this extension: https://github.com/julia-vscode/julia-vscode/pull/1276. 

So I suppose that means we need to:

* Add function that will install an exception handler (`.catch`) to handle User_error before ignoring
* Add try/catch in all providers etc. to catch synchronously thrown `User_error` and handle them the same way as above.

smorimoto:
There is no change log entry for this yet, so I think we can release 1.9.0 after adding it.

ulugbekna:
We shouldn't release before we add more principal handling of ocaml-lsp
versions in the extension; otherwise, we get false warnings about outdated
ocaml-lsp version because people on ocaml v < 4.12 cannot really update to
the latest ocaml-lsp. Another thing that should be handled before a release
is that jumping between holes keyboard shortcuts coincide with important
shortcuts on French keyboard.

ulugbekna:
Hi, I'm getting an unbound module error when I try to build the master, which version of ppxlib are we expected to have? both ppxlib master and latest release are failing

```
dune build src/vscode_ocaml_platform.bc.js
File "src/ppx_tools/ppx_tools.mli", line 5, characters 38-53:
5 | val get_preprocessed_ast : string -> (Ppxlib.Ast_io.t, string) result
                                          ^^^^^^^^^^^^^^^
Error: Unbound module Ppxlib.Ast_io
make: *** [build] Error 1
```

rgrinberg:
> We shouldn't release before we add more principal handling of ocaml-lsp
versions in the extension; otherwise, we get false warnings about outdated
ocaml-lsp version because people on ocaml v < 4.12 cannot really update to
the latest ocaml-lsp

How about we just make it possible for the user to disable such warnings per project? I'd rather not introduce some heavy weight solution to a problem that is fixed by upgrading.

> Another thing that should be handled before a release
is that jumping between holes keyboard shortcuts coincide with important
shortcuts on French keyboard.

Okay, let's get rid of the shortcuts for now then. We can think of better shortcuts later.

src/ast_editor.ml
```diff
@@ -0,0 +1,617 @@
...
+end = struct
+  type kind =
+    | Structure
+    | Signature
+    | Unknown
+
+  let relative_document_path ~document =
+    Workspace.asRelativePath ~pathOrUri:(`Uri (TextDocument.uri document)) ()
+
+  let project_root_path () = Workspace.rootPath ()
```
Khady, line 63: would it be possible to use dune to get the root of the project instead of vscode's workspace? The current code doesn't work in the context of a monorepo

rgrinberg: Hmm, but why not? Doesn't it make sense to open the monorepo such that its root matches the workspace root?

Khady: here I'm thinking about a monorepo in a larger sense than just a dune monorepo. A repo containing multiple projects written in multiple languages with different build systems. It's convenient to have the whole thing open (search in the whole repo, check that a change in one project doesn't break the other ones, ...)

This is not a total deal breaker. It would just make the experience better. I've created https://github.com/ocamllabs/vscode-ocaml-platform/issues/873 to keep track of the idea.

rgrinberg: I see. Well I'm definitely not against the idea.

PR #675, opened 2021-08-02
Support Local mode in Construct - user configuration option

ulugbekna:
1. I'm not very sure about how `InitializationOptions` module should look like -- let's see how it evolves, and we'll refactor it perhaps. 
2. Using `InitializationOptions` to compute specific things to pass to the language server vs just passing the OCaml Platform settings section (the whole json representation of ocaml platform related settings) (e.g., Rust Analyzer just takes the settings section `rust-analyzer` and passes it to the language server). I think it's better if we pass to ocaml-lsp only settings that it needs even though that results in extra code to write & maintain. Wdyt?

### DIFF
--- a/vscode/vscode_languageclient/vscode_languageclient.ml
+++ b/vscode/vscode_languageclient/vscode_languageclient.ml
@@ -81,6 +81,23 @@ end
 module ClientOptions = struct
   include Interface.Make ()
 
+  type any_or_thunk =
+    [ `Any of Ojs.t
+    | `Thunk of unit -> Ojs.t
+    ]
+
+  let any_or_thunk_of_js js_val =
+    match Ojs.type_of js_val with
+    | "object" -> `Any js_val
+    | "function" -> `Thunk ([%js.to: unit -> Ojs.t] js_val)
+    | unexpected_type ->
+      failwith @@ "any_or_thunk_of_js: expected 'object' or 'function' but got "
+      ^ unexpected_type
+
+  let any_or_thunk_to_js = function
+    | `Any ojs -> ojs
+    | `Thunk f -> [%js.of: unit -> Ojs.t] f
+
   include
     [%js:
     val documentSelector : t -> DocumentSelector.t or_undefined [@@js.get]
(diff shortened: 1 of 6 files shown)


rgrinberg:
I've thought about this a little more about your PR on the LSP side, and I think we should clarify our goals regarding stuff like this first. The ideal number of configuration options I imagine for the lsp server is 0. In general, I view configuration as a necessary evil rather than a feature. Clearly, some configuration is unavoidable, but this does not mean we should be adding more without strong justification. Therefore, I ask why is adding end user configuration is unavoidable in this case? Is this really the simplest way for you to continue developing and working on this feature? Do you have some group of users in mind that are willing to use this option that you need to make it easy for them to tweak this?

ulugbekna:
That's an experimental feature that users can try turning on to see for themselves if it's worth it, which seems reasonable and done in other LSP implementations. Nevertheless, there is also a question of whether a common user will know about this config option and understand it - that largely depends on how well we document it. 

I think that having opt-in experimental features as config helps to get user feedback on new features, especially if they are going to be in development for quite some time. For example, I found a couple of bugs as I was adding support for Construct in ocaml-lsp and used local mode on at the beginning.

About this particular config: I can replace this config with an env var for ocaml-lsp, but I would say _some_ users will be missing out -- I could even imagine a report where a user says that they have `Construct an expression` working but it doesn't propose locally defined values in the proposed expressions. 

If we keep this as a config, I should put this config under key `ocaml.experimental.construct.useLocalContext` I think.

Yes, the idea of this config came up as a way to help Ulysse to develop the feature faster, but I was also missing out on the feature as an end user (I don't work on `Construct`)

rgrinberg:
If this is indeed going to be useful to @voodoos, then I can accept this feature. Otherwise, I'm going to reject elaborate configuration that is useful for n=1 users.

vscode/vscode_languageclient/vscode_languageclient.ml
```diff
@@ -81,6 +81,23 @@ end
...
+    [ `Any of Ojs.t
+    | `Thunk of unit -> Ojs.t
+    ]
+
+  let any_or_thunk_of_js js_val =
+    match Ojs.type_of js_val with
+    | "object" -> `Any js_val
+    | "function" -> `Thunk ([%js.to: unit -> Ojs.t] js_val)
+    | unexpected_type ->
+      failwith @@ "any_or_thunk_of_js: expected 'object' or 'function' but got "
```
rgrinberg, line 94: Shouldn't this be `invalid_arg`?

PR #677, opened 2021-08-05
Improve working with REPL

### LINKED ISSUE #535: More REPL features
Borrowing from https://github.com/BetterThanTomorrow/calva, it would be nice to have more REPL features. There are many great commands there, but the two most important for OCaml I believe are:
- Evaluate current file
- Evaluate current form
- Evaluate current top-level-form

The first is fairly self explanatory and maps directly to "#use "path/to/current/file". The second two are slightly more tricky in OCaml than in Clojure. I think the thing to do for the for "Evaluate current form" is to "grow" the selection upwards until you reach the nearest `let` node in the AST and send all that to the REPL. For "Evaluate current top-level-form", I think it should grow until it reaches a "root" `let` node in that current module. Some examples, using `|||` to indicate the position of the cursor:

Evaluate current form
```ocaml
let foo = ||| "bar";
(* sends: let foo = "bar"; *)

module Foo = struct
let foo x =
  let bar = |||"bam" in
  bar
(* sends: let bar = "bam" *)
end
```
Evaluate current top-level form

```ocaml
let foo = ||| "bar";
(* sends: let foo = "bar"; *)

module Foo = struct
let foo x =
  let bar = |||"bam" in
  bar
(* sends: 
let foo x =
  let bar = |||"bam" in
  bar
*)
end
```

Does ocamllsp make it easy to find parent nodes of the current node in the AST?

ulugbekna:
This PR has several improvements to how we work with REPLs: 

1. Improved workflow - now if the user doesn't have indicated `ocaml.repl.path` and `ocaml.repl.args`, we don't launch a default REPL, but prompt the user to enter a command to launch the REPL, e.g., `dune utop my_library -- -implicit-bindings`, which is more useful than the default `dune utop`. Read more about workflow improvement in the Changelog (and code). Update: user now can also pick to launch the REPL either in a new terminal or an existing one.
2. Add support for evaluating "toplevel" expression enclosing the cursor. See the Changelog for a detailed description of how that works. Internally, it uses a custom request `ocamllsp/wrappingAstNode` implemented in (https://github.com/ocaml/ocaml-lsp/pull/482). (closes #535) 
3. Rename the command `Evaluate Selection` to `Evaluate Expression in REPL`:
- because of (2), we now evaluate not only a selection, but also expressions enclosing the cursor 
- indicates that the evaluation happens in a REPL
4. Now evaluation doesn't steal focus from the editor to the REPL

_How to review_ 

Commits following logical steps, but sometimes a later commit overwrites a part of what a previous one did. So commit-by-commit review should be easier but may be more time-consuming. Any comment/feedback is welcome because I took some liberty in making changes that are not directly related to the goal of this PR -- I wouldn't call those changes cosmetic, but they just move around code (though make code easier IMHO)

_TODO_

- [x] improve prompt in inputbox to launch REPL
- [x] needs a fix in utop: when passing a large piece of code parts of the code are repeated -- needs investigation & a fix preferably
    - entering a large piece of code results in duplication, so that should be fixed in utop 
- [x] support opening the REPL in an existing terminal (important)
- [x] add support for "evaluate a file" (should this happen in another PR?)
- [ ] clean up code
- [ ] determine whether `Exit` cmd works in windows shell

Should this PR be broken into several?

### DIFF
--- a/src/import.ml
+++ b/src/import.ml
@@ -193,3 +193,14 @@ module Range = struct
 end
 
 let sprintf = Printf.sprintf
+
+module Node = struct
+  let set_timeout milliseconds =
+    Promise.make (fun ~resolve ~reject:_ ->
+        let (_ : Node.Timeout.t) =
+          Node.setTimeout (fun () -> resolve ()) milliseconds
+        in
+        ())
+
+  include Node
+end
--- a/src/import.ml
+++ b/src/import.ml
@@ -142,6 +142,18 @@ module Position = struct
     let character = Position.character t in
     Jsonoo.Encode.(object_ [ ("line", int line); ("character", int character) ])
 
+  type position =
+    { line : int
+    ; character : int
+    }
+
+  let position_of_t t =
+    let line = Position.line t in
+    let character = Position.character t in
+    { line; character }
+
+  let t_of_position { line; character } = Position.make ~line ~character
+
   include Vscode.Position
 end
 
@@ -161,6 +173,21 @@ module Range = struct
     let start = Range.start t |> Position.json_of_t in
     let end_ = Range.end_ t |> Position.json_of_t in
     Jsonoo.Encode.(object_ [ ("start", start); ("end", end_) ])
+
+  type range =
+    { start : Position.position
+    ; end_ : Position.position
+    }
+
+  let range_of_t t =
+    let start = Range.start t |> Position.position_of_t in
+    let end_ = Range.end_ t |> Position.position_of_t in
+    { start; end_ }
+
+  let t_of_range { start; end_ } =
+    let start = Position.t_of_position start in
+    let end_ = Position.t_of_position end_ in
+    Range.makePositions ~start ~end_
 
   include Vscode.Range
 end
--- a/src/import.ml
+++ b/src/import.ml
@@ -174,21 +174,6 @@ module Range = struct
     let end_ = Range.end_ t |> Position.json_of_t in
     Jsonoo.Encode.(object_ [ ("start", start); ("end", end_) ])
 
-  type range =
-    { start : Position.position
-    ; end_ : Position.position
-    }
-
-  let range_of_t t =
-    let start = Range.start t |> Position.position_of_t in
-    let end_ = Range.end_ t |> Position.position_of_t in
-    { start; end_ }
-
-  let t_of_range { start; end_ } =
-    let start = Position.t_of_position start in
-    let end_ = Position.t_of_position end_ in
-    Range.makePositions ~start ~end_
-
   include Vscode.Range
 end
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,18 +56,31 @@
 
   - Improved workflow for opening a terminal:
 
-    If the `ocaml.repl.path` and `ocaml.repl.args` configuration options are
-    given, then they are used to open a terminal and launch a REPL.
+    - Picking which REPL to use
 
-    Otherwise, the user is prompted to write a string command to launch the REPL
-    in an input box; for example, `dune utop my_library`.
+      If the `ocaml.repl.path` and `ocaml.repl.args` configuration options are
+      given, then they are used to open a terminal and launch a REPL.
 
-    If the user doesn't enter anything, a default (\*) REPL is created:
+      Otherwise, the user is prompted to write a string command to launch the
+      REPL in an input box; for example, `dune utop my_library`.
 
-    - If you have both `dune` and `utop` and the project _seems_ "buildable",
-      use `dune utop` command to create a REPL;
-    - If only `utop` is installed, then use command `utop` to launch REPL;
-    - Else, use the default OCaml REPL `ocaml`.
+      If the user doesn't enter anything, a default (\*) REPL is created:
+
+      - If you have both `dune` and `utop` and the project _seems_ "buildable",
+        use `dune utop` command to create a REPL;
+      - If only `utop` is installed, then use command `utop` to launch REPL;
+      - Else, use the default OCaml REPL `ocaml`.
+
+    - Picking to launch REPL in a new terminal or existing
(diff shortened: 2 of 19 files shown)


src/import.ml
```diff
@@ -162,5 +174,31 @@ module Range = struct
...
+  let range_of_t t =
+    let start = Range.start t |> Position.position_of_t in
+    let end_ = Range.end_ t |> Position.position_of_t in
+    { start; end_ }
+
+  let t_of_range { start; end_ } =
+    let start = Position.t_of_position start in
+    let end_ = Position.t_of_position end_ in
+    Range.makePositions ~start ~end_
+
```
mnxn, line 191: I don't think this type is used anywhere? It seems unnecessary.

src/import.ml
```diff
@@ -162,5 +174,31 @@ module Range = struct
...
   include Vscode.Range
 end
+
+module Node = struct
+  let set_timeout milliseconds =
+    Promise.make (fun ~resolve ~reject:_ ->
+        let (_ : Node.Timeout.t) =
+          Node.setTimeout (fun () -> resolve ()) milliseconds
+        in
+        ())
```
mnxn, line 201: I would move this to the actual `node.ml` file with a name that references the fact that it returns a promise. Something like `setTimeoutPromise`.

ulugbekna:
I'm stuck at making a decision and would love your input: 

In the current `eval expr in repl` command, we wait for a couple of seconds before writing the expression being evaluated into the terminal: the wait is for opening/creating terminal and running the user-provided code to launch the REPL. The wait time is too long when user simply wants to open a new terminal and run `utop` in it and pass the code into utop AND the wait is too short when user wants to run `dune utop my_lib` to open the terminal (too short means that the code being evaluated is sent into terminal before the repl launches:

image

The terminal API in vscode is very limited, by which we can launch a terminal and send keystrokes -- that's it. So there are two options: 

1. We make working with REPL more limited: allow user to "attach" and "detach" to a terminal running a repl: when user wants to evaluate some code in repl, they first "attach" to a running terminal with already launched REPL; the extension remembers that terminal, and all future evaluations happen in that REPL. However, if the user exits REPL (and don't close the terminal), they have to "detach" the extension from that terminal; otherwise, code will be sent to the shell. 

2. The other option is to keep our current workflow, in which code can be accidentally sent to the shell because the REPL hasn't been launched. The other problem is that we append `; exit` to the user's REPL launching command to kill the terminal automatically & thus automate "detach"ing. But that results in the REPL launch error being dismissed. If we do `&& exit`, user again need to detach terminal themselves.

tmattio:
Thanks @ulugbekna!

> Improved workflow - now if the user doesn't have indicated ocaml.repl.path and ocaml.repl.args, we don't launch a default REPL, but prompt the user to enter a command to launch the REPL, e.g., dune utop my_library -- -implicit-bindings, which is more useful than the default dune utop. Read more about workflow improvement in the Changelog (and code). Update: user now can also pick to launch the REPL either in a new terminal or an existing one.

I don't think this is improving the workflow. As a new user, I might not know the commands to run the REPL. I think custom REPL commands should remain an advanced feature, provided by the extension settings. I agree, however, that implicit bindings are more useful by default. Maybe an alternative is to run the toplevel with `--implicit-bindings` by default? IIRC, I didn't do this because the REPL would fail if the project does not build. @rgrinberg Is there a way to make `dune utop --implicit-bindings` use the latest successful build state if the project does not build?

The rest of the improvements make a lot sense to me 🙂

rgrinberg:
> Is there a way to make dune utop --implicit-bindings use the latest successful build state if the project does not build?

I don't see an easy way to do it, but why would this be good anyhow? It would be confusing to show the user an out of date repl.

dragomirtitian:
I am testing out this extension for use in an academic setting. The lack of full file eval in REPL is a deal breaker for us ATM.

Do you have any idea if this PR will make it in anytime in the next 2 weeks ?

ulugbekna:
@dragomirtitian, sure, I will try to work on this (at least for full-file eval) and a new release next week :-) 

Just a heads up: Running `#use "path/to/file.ml"` to evaluate the whole file in UTop can be a good alternative that doesn't involve the extension

dragomirtitian:
@ulugbekna  Thank you so much. That would be very useful.

rgrinberg:
Please re-open once you make progress. I'm pretty convinced that we should be able to get rid of all the timeout hacks as well.

OscarLahaie:
@ulugbekna is the project still being actively developed? I would like to implement the same features but I prefer if possible not to start from scratch.

ulugbekna:
Hi @OscarLahaie , this work is currently stale. If you'd like to work on this, feel free to checkout my branch and make updates to it. 

If you describe what particularly you'd like to do, I can advise on things. The current code on this branch is somewhat more complicated than it should be, I think.

OscarLahaie:
Hello @ulugbekna,
Do you think the use of `#use "..."` is sufficient for a whole file evaluation command?

ulugbekna:
Hi Oscar,
Yes that should be enough

OscarLahaie:
Nice !
As for the evaluation of an expression.
If I understand correctly, you use ocaml-lsp to define the code to be executed?

ulugbekna:
Code can be evaluated in repl in different ways:

- evaluate selection - works by getting from vscode the selection as a
string and passes it to the repl
- evaluate a definition (AST node) at cursor - send the position to
ocamllsp to get the node location and passes it to the repl
- evaluate a file gets the file path and just sends a #use command to repl
with the file path

ulugbekna:
Not sure what you mean by “implement ast node” but this branch already includes code that does that

OscarLahaie:
> Not sure what you mean by “implement ast node” but this branch already includes code that does that

No sorry i was talking about implementing `evaluate expression` with Ast-node

PR #682, opened 2021-08-13
Windows case-insensitivity

jonahbeckford:
* External Opam switches (aka local switches) often were not visible to
  the extension because of the case mismatch between the switch path as
  reported by `opam switch list` and the workspace folder path reported by
  VS Code
* ${pathSeparator} adds a standard VS Code variable which lets VS Code
  projects open portably on Windows or Unix

---

Also included a DEVELOPING.md containing slightly reverse-engineered instructions.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -1,0 +1,23 @@
+# Developing
+
+## Quick Start
+
+FIRST, install the OCaml dependencies and generate the OCaml -> Javascript
+artifacts:
+
+```bash
+opam install . --deps-only --with-test
+make build
+```
+
+SECOND, press F5 in VS Code to launch the "Extension Development Host".
+
+## Code Hygiene
+
+```bash
+esy
+yarn
+opam install ocamlformat
+make fmt
+make test
+```
--- a/src/settings.ml
+++ b/src/settings.ml
@@ -53,8 +53,14 @@ let resolve_workspace_vars setting =
   Interop.Regexp.replace setting ~regexp ~replacer
 
 let substitute_workspace_vars setting =
+  (* Windows paths are case-insensitive *)
+  let folder_replace_all =
+    match Platform.t with
+    | Win32 -> String.Caseless.substr_replace_all
+    | _ -> String.substr_replace_all
+  in
   List.fold (Workspace.workspaceFolders ()) ~init:setting ~f:(fun acc folder ->
-      String.substr_replace_all acc
+      folder_replace_all acc
         ~pattern:(workspace_folder_path folder)
         ~with_:(workspace_folder_var folder))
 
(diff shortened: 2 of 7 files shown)


jonahbeckford:
(The Permission Denied CI failure for `windows-latest` does not seem to be related at all to the PR)

DEVELOPING.md
```diff
@@ -0,0 +1,22 @@
+# Developing
+
+## Quick Start
+
+FIRST, install the OCaml dependencies and generate the OCaml -> Javascript
+artifacts:
+
+```bash
+opam install gen_js_api js_of_ocaml jsonoo opam-file-format promise_jsoo
```
mnxn, line 9: The dependencies can automatically be read from the opam files (like how it's done in CI).
```suggestion
opam install . --deps-only --with-test
```

src/settings.ml
```diff
@@ -50,13 +51,34 @@ let resolve_workspace_vars setting =
...
+          ~pattern:(workspace_folder_path folder)
+          ~with_:(workspace_folder_var folder)
+      in
+      if String.equal acc' acc then
+        (* no ${workspaceFolder:folder_name}, so pointless and verbose-y using
+           cross-platform ${pathSeparator} *)
+        acc
+      else
+        String.substr_replace_all acc' ~pattern:(Char.to_string Path.sep)
+          ~with_:"${pathSeparator}")
```
mnxn, line 81: Replacing every separator with `${pathSeparator}` makes the path quite messy.
Is there any benefit to this over replacing all separators with the forward slash?

I haven't had a chance to check, but I think Windows still reads a forward-slash as a path separator.
If it doesn't, the `resolve_workspace_vars` function could just replace forward slashes with `Path.sep`.

If the paths come from detected sandboxes, I think it's unlikely for there to be mixed separators.

jonahbeckford: > Replacing every separator with ${pathSeparator} makes the path quite messy.

It is slightly messy, which is why the PR originally only did it for relative paths rooted at `${workspaceFolder}`. But that ${pathSeparator} change is not as critical as the case-insensitivity change, so I reduced this PR to just the case-insensitivity.

> Is there any benefit to this over replacing all separators with the forward slash?
> I think Windows still reads a forward-slash as a path separator

I don't think it was Windows that had the problem, but some parsing code which split path parts by whatever the OS path separator was (I think the node.js `Node.path.dirname`, etc. was the problem, but I'll revisit it after the Windows case-insensitivity issue). Anyway, changing the path parsing code seemed like a huge blast radius, so I didn't go down that route. Your suggestion of replacing in `resolve_workspace_vars` looks safer.

jonahbeckford:
Rebased and resubmitted the now-narrowed PR with just Windows case-insensitivity. Pending a maintainer to approve the workflow. Thanks!

mnxn (approved):
Looks good.

Thanks for contributing!

PR #726, opened 2021-10-05
Improve setup

tmattio:
In anticipation of the Outreachy application period, this PR does quite a few things:

- Move prettier configuration to `packages.json`
- Comment `gitignore` rules
- Remove unused metafiles to clean up the project top level
- Move CHANGELOG.md to CHANGES.md to stay in line with OCaml conventions.
- Add make commands to create the switch and install dependencies
- Remove the Esy setup
- Remove yarn files in favor of npm files.
- Rename the extension to `OCaml` and update the logo
- Ensure AST structures are converted properly to allow compilations on compilers other than 4.12

Once this is merged, I'll add a CONTRIBUTING guide as a follow up PR.


tmattio:
If anything seems controversial here, I can separate it to another PR.

rgrinberg:
Is the npm setup just as fast a the yarn setup? If it is, the same setup should be done for the ocamllsp repo.

.github/workflows/main.yml
```diff
@@ -37,60 +37,51 @@ jobs:
           node-version: ${{ matrix.node-version }}
 
       - name: Install ocaml-lsp-server
-        run: |
-          opam pin ocaml-lsp-server https://github.com/ocaml/ocaml-lsp.git --no-action
-          opam install ocaml-lsp-server
+        run: opam install ocaml-lsp-server
```
rgrinberg, line 40: How does this work if there's no functional version of ocamllsp for 4.13?

rgrinberg:
The lock files for ast explorers have been removed. How is building ast explorer with the same dependency versions going to work?

rgrinberg:
I'm also against renaming the extension. There's very little to gain from renaming and the churn to developers and users is non trivial.

.github/workflows/main.yml
tmattio: There is though: https://v3.ocaml.org/p/ocaml-lsp-server/1.9.0~4.13preview. Or is it what you meant by "functionnal"? The preview version isn't functionnal?

tmattio:
> Is the npm setup just as fast a the yarn setup?

As far as I know, there's no major performance difference.

> The lock files for ast explorers have been removed.

They haven't been removed, just replaced with `package-lock.json`, or do you mean something else?

> I'm also against renaming the extension. There's very little to gain from renaming and the churn to developers and users is non trivial.

There is a lot of value in renaming the extension and making it clear that this is the official extension by changing the logo and updating the description. When you type Rust, Scala, Python, Erlang, etc. on the marketplace, which extension is the "official" one is very clear. When you type OCaml, however...
image

This is a pain point that is expressed by beginners, as seen by the number of people asking "Which extension should I install on VSCode?".

I understand that renaming the extension will create a bit of friction, but this can be mitigated by some communication to the community and will only be a temporary situation, whereas the friction beginners experience will be permanent if we don't make it obvious that the extension is the official one.

tmattio:
Merging now as the application period for Outreachy has already started, but I'll address any review in follow-up PRs.

PR #747, opened 2021-10-18
Update keybindings for holes

ulugbekna:
Fixing clashes with important keybindings in French keyboard

Note: is based on #746 because that PR fixes some extension installation problems; see the single commit relevant to this PR to review

I would completely remove the keybindings but that'd be annoying workflow breaker IMHO

@tmattio , could you please confirm the new keybindings work well with French layout?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


tmattio:
> @tmattio , could you please confirm the new keybindings work well with French layout?

I don't have a French keyboard on hand unfortunately 😕

Issue #423, opened 2020-11-01
Highlight current switch in sandbox selection


rgrinberg:
The switch returned by `opam switch show` should be distinguished somehow in our sandbox selection box.

ulugbekna:
yeah, it should probably be displayed first in the list of sandboxes available and should be pre-selected.

ulugbekna:
Fixed in #751

PR #751, opened 2021-10-21
Highlight current switch

panglesd:
As a response to #423, this PR improves UX for switching switches.

1) The currently VS-code selected switch is highlighted in the list of switches:
image

2) The result of `opam switch show` is shown as the first item in the switch selection box: 
image


panglesd:
I'm a little bit confused though. Isn't `Global` already supposed to be the result of `opam switch show`?
If yes, wouldn't it be better to have a signle entry, displaying the name of the switch as well as `Global` ?

I think that I also should deal with the case where the current swith is `Global` in item 1. of my first message... (I currently do nothing)

src/treeview_switches.ml
```diff
@@ -34,15 +42,33 @@ module Dependency = struct
 
   let icon = function
     | Switch _ ->
-      TreeItem.LightDarkIcon.
-        { light = `String (Path.asset "dependency-light.svg" |> Path.to_string)
-        ; dark = `String (Path.asset "dependency-dark.svg" |> Path.to_string)
-        }
+      fun b ->
```
tmattio, line 45: We can give this variable a full name to make the code clearer:

```suggestion
      fun is_current_sandbox ->
```

src/treeview_switches.ml
```diff
@@ -34,15 +42,33 @@ module Dependency = struct
...
+                     "")
+                 ^ ".svg"
+                |> Path.to_string)
+          }
     | Package _ ->
-      TreeItem.LightDarkIcon.
-        { light = `String (Path.asset "number-light.svg" |> Path.to_string)
-        ; dark = `String (Path.asset "number-dark.svg" |> Path.to_string)
-        }
+      fun _ ->
```
tmattio, line 67: Same thing as above:

```suggestion
      fun _is_current_sandbox ->
```

src/treeview_switches.ml
```diff
@@ -34,15 +42,33 @@ module Dependency = struct
...
+      fun b ->
+        TreeItem.LightDarkIcon.
+          { light =
+              `String
+                (Path.asset @@ "dependency-light"
+                 ^ (if b then
+                     "-selected"
+                   else
+                     "")
+                 ^ ".svg"
```
tmattio, line 54: We can put this in a function maybe since it's reused below.

src/sandbox.ml
```diff
@@ -291,13 +291,15 @@ let of_settings_or_detect () =
...
 let save_to_settings sandbox =
+  let open Promise.Syntax in
   let to_setting = function
     | Esy (_, root) -> Setting.Esy root
     | Opam (_, switch) -> Setting.Opam switch
     | Global -> Setting.Global
     | Custom template -> Setting.Custom template
   in
-  Settings.set ~section:"ocaml" Setting.t (to_setting sandbox)
+  let+ () = Settings.set ~section:"ocaml" Setting.t (to_setting sandbox) in
```
tmattio, line 301: This change does not seem to be necessary.

tmattio (commented):
Looks very good, thanks a lot @panglesd. I left a few comments, and this is ready to merge no my side once they are addressed.

src/sandbox.ml
panglesd: I think it is necessary as we need to wait for the change to be effective before going on to the next line

panglesd: Yes you are right. In commit 98a1b6e I moved an instruction from some place `A` to some place `B`, and forgot to remove the now useless `let*` in place `A`. I though you were referring to place `B`. Anyway, I removed the useless line :)

src/treeview_switches.ml
```diff
@@ -5,6 +5,14 @@ module Dependency = struct
     | Package : Opam.Package.t -> t
     | Switch : Opam.t * Opam.Switch.t -> t
 
+  let dependency_is_sandbox = function
+    | Sandbox.Opam (opam1, switch1) -> (
+      function
+      | Switch (opam2, switch2) ->
+        Opam.equal opam1 opam2 && Opam.Switch.equal switch1 switch2
+      | _ -> false)
+    | _ -> fun _ -> false
```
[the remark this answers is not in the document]
panglesd: You are right, and the code is clearer now.

ulugbekna:
Thanks @panglesd !

PR #756, opened 2021-10-31
Improve odoc syntax highlighting

### LINKED ISSUE #470: Improve odoc syntax highlighting
https://ocaml.github.io/odoc/odoc/using-odoc.html

SaySayo:
fixes #470. This PR is still a draft. @mnxn can you please review to see if i'm on the right path. I tried to write regexes that would match inline-texts which I assumed would be random sentences. Also made the parent scope `comment.other.odoc` instead of ocamldoc due to the file name change


mnxn:
Since you changed the name from ocamldoc to odoc, you need to update it in `package.json` too.

> I tried to write regexes that would match inline-texts which I assumed would be a random sentences. 

You do not need to match `inline-text`. It will be everything between `begin` and `end`.

syntaxes/odoc.json
```diff
@@ -0,0 +1,81 @@
...
+      "begin": "\\{\\s\\{[0-9]\\}\\s\\:\\s[label]\\s[[\\w+\\s+]+[\\W]]*\\b",
+      "end": "\\}",
+      "beginCaptures": {
+        "1": { "name": "variable.other.declaration.odoc" }
+      },
+      "patterns": [{ "include": "#markup" }]
+    },
+    {
+      "comment": "text in bold",
+      "begin": "\\{[b]\\s[[\\w+\\s+]+[\\W]?]*\\b",
```
mnxn, line 32: I think the brackets around `b` in `[b]` are unnecessary.

syntaxes/odoc.json
```diff
@@ -0,0 +1,81 @@
...
+  "patterns": [
+    {
+      "name": "comment.other.odoc",
+      "begin": "",
+      "end": "",
+      "patterns": [{ "include": "#markup" }]
+    },
+    {
+      "comment": "section header",
+      "begin": "\\{\\s\\{[0-9]\\}\\s[[\\w+\\s+]+[\\W]]*\\b",
```
mnxn, line 14: In the documentation `{ 0 … 9 }+` means one or more digits.

Instead of `\\{[0-9]\\}`, you should do `[[:digit:]]+`.

syntaxes/odoc.json
```diff
@@ -0,0 +1,81 @@
...
+  "scopeName": "source.ocaml.odoc",
+  "name": "comment.other.odoc",
+  "fileTypes": ["mld"],
+  "patterns": [
+    {
+      "name": "comment.other.odoc",
+      "begin": "",
+      "end": "",
+      "patterns": [{ "include": "#markup" }]
+    },
```
mnxn, line 11: I think everything except this and `odoc documentation tag` should be in `markup`.

syntaxes/odoc.json
```diff
@@ -0,0 +1,81 @@
...
+        },
+        {
+          "comment": "center text",
+          "begin": "\\{C\\s[[\\w+\\s+]+[\\W]?]*\\b",
+          "end": "\\}",
+          "patterns": [{ "include": "#markup" }]
+        },
+        {
+          "comment": "left align text",
+          "begin": "\\{[L]\\s[[\\w+\\s+]+[\\W]?]*\\b",
```
mnxn, line 68: The `[L]` and `[R]` don't need brackets either.

mnxn:
> You do not need to match `inline-text`. It will be everything between `begin` and `end`.

To be clear, you can just remove the `\\s[[\\w+\\s+]+[\\W]?]*` from the `begin` pattern.

syntaxes/odoc.json
```diff
@@ -0,0 +1,81 @@
...
+        {
+          "comment": "embedded ocaml source",
+          "begin": "(?<!\\\\)\\[",
+          "end": "(?<!\\\\)\\]",
+          "contentName": "source.embedded.odoc",
+          "patterns": [{ "include": "source.ocaml" }]
+        },
+        {
+          "comment": "section header",
+          "begin": "\\{\\s\\{[[:digit:]]+\\}\\b",
```
mnxn, line 30: There shouldn't be `\\{` and `\\}` around the `[[:digit:]]+`.

mnxn: and there should be parenthesis around the `[[:digit:]]+` for the capture to work correctly.

syntaxes/odoc.json
```diff
@@ -27,17 +27,17 @@
...
             "1": { "name": "constant.language.odoc" }
           },
           "patterns": [{ "include": "#markup" }]
         },
         {
           "comment": "label",
-          "begin": "\\{\\s\\{[[:digit:]]+\\}\\s\\:\\s[label]\\b",
-          "end": "\\}",
+          "begin": "[[:digit:]]+\\b",
+          "end": "",
```
mnxn: You removed too much stuff. Only the `\\{` and `\\}` that were directly surrounding the `[[:digit:]]+` should have been removed (not all of them). And you also removed the label from the second part.

SaySayo:
I reverted the new changes and just omitted the `\\{`, I assumed you mean `begin: ""` and `end: ""` should be empty string to match everything in between

mnxn:
The `\\{` and `\\}` are correct now.

It might be hard to see, but the curly brackets in https://ocaml.org/manual/ocamldoc.html#inline-text-element are actually two different colors. The brown curly brackets are what should be included in the grammar. The black curly brackets are [EBNF syntax](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form#Table_of_symbols) for "repetition" and the `+` means "one or more".


If you weren't already aware, you can use <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> (or kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> on macOS) to open the VSCode Command Pallete and search for `Developer: Inspect Editor Tokens and Scopes`.
That feature will let you see which how a file is being recognized by the syntax:

image

Before pushing a commit, try to make sure that everything is being highlighted correctly. You can look in this repository for some example odoc files: https://github.com/ocaml/odoc/tree/master/doc

SaySayo:
Noted. Thank you!

syntaxes/odoc.json
```diff
@@ -0,0 +1,169 @@
...
+          "patterns": [{ "include": "#markup" }]
+        },
+        {
+          "comment": "section header",
+          "begin": "\\{([[:digit:]]+)\\b",
+          "end": "\\}",
+          "beginCaptures": {
+            "1": { "name": "constant.language.odoc" }
+          },
+          "contentName": "constant.language.odoc",
```
mnxn, line 45: Why did you add a `contentName` to these two rules?

They should only have highlighting in the captures.

SaySayo: I thought the texts within the braces also had to be highlighted. Will take it out.

syntaxes/odoc.json
```diff
@@ -0,0 +1,169 @@
...
+          "begin": "\\{ol\\b",
+          "end": "\\}",
+          "patterns": [{ "include": "#markup" }]
+        },
+        {
+          "comment": "link address text",
+          "begin": "\\{\\{\\:\\b",
+          "end": "\\}",
+          "contentName": "string.other.odoc",
+          "patterns": [{ "include": "#markup" }]
```
mnxn, line 104: The second left curly bracket is unmatched and the text for the link is not highlighted:

image

I think a combination of [regex lookahead](http://www.rexegg.com/regex-lookarounds.html) for the `{:` and adding a rule in the `patterns` will work

SaySayo: Hi @mnxn, I've made the requested changes in your review but I have a bit of issue matching the link address text. Here's the regex I wrote for it, its still matches the string but doesn't match the link text.
Screenshot 2021-11-12 at 11 17 27

mnxn: I was thinking something more like this

```json
"begin": "\\{(?=\\{\\:)",
"end": "\\}",
```

Then you can add a pattern to `patterns`:
```json
"patterns": [
    {
        "begin": "\\{\\:",
        "end": "\\}"
    },
    { "include":" #markup" }
]
```


I'll leave the highlighting names to you.

This way the parent pattern will match the outer `{` and `}` without matching the inner `{:`. The inner `{:` and `}` can be matched by the inner pattern.

syntaxes/odoc.json
```diff
@@ -0,0 +1,169 @@
...
+        "comment": "string in source code",
+        "begin": "\\[\\b",
+        "end": "\\]",
+        "contentName": "string.other.odoc",
+        "patterns": [{ "include": "#markup" }]
+      },
+      {
+        "comment": "string in preformatted source code",
+        "begin": "\\{\\[\\b",
+        "end": "\\}",
```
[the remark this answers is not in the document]
mnxn: To be clear, when `begin` ends in a letter, digit, or underscore, there should still be a `\\b`.

You removed too many `\\b`s.

SaySayo:
Oh didnt realize that. Thank you for your patient reviews.

syntaxes/odoc.json
```diff
@@ -0,0 +1,171 @@
...
+        "comment": "string in preformatted source code",
+        "begin": "\\{\\[",
+        "end": "\\]\\}",
+        "contentName": "string.other.odoc",
+        "patterns": [{ "include": "#markup" }]
+      },
+      {
+        "comment": "string in verbatim",
+        "begin": "\\{v\\b",
+        "end": "v\\}",
```
mnxn, line 125: Since this regex pattern beings with a letter it needs a boundary too:
```suggestion
        "end": "\\bv\\}",
```

syntaxes/odoc.json
```diff
@@ -0,0 +1,171 @@
...
+          "contentName": "source.embedded.odoc",
+          "patterns": [{ "include": "source.ocaml" }]
+        },
+        {
+          "comment": "label",
+          "begin": "\\{([[:digit:]]+)\\:[[:space:]]?[a-z]+",
+          "end": "\\}",
+          "beginCaptures": {
+            "1": { "name": "variable.other.declaration.odoc" }
+          },
```
mnxn, line 34: This needs two separate captures and colors: one for the number and one for the label.

You are currently highlighting the number as a variable.

syntaxes/odoc.json
```diff
@@ -0,0 +1,171 @@
...
+        },
+        { "include":" #markup" }]
+        },
+        {
+        "comment": "string in source code",
+        "begin": "\\[",
+        "end": "\\]",
+        "contentName": "string.other.odoc",
+        "patterns": [{ "include": "#markup" }]
+      },
```
mnxn, line 114: This is already recognized by the `embedded ocaml source` rule.

syntaxes/odoc.json
```diff
@@ -0,0 +1,171 @@
...
+      {
+        "comment": "string in preformatted source code",
+        "begin": "\\{\\[",
+        "end": "\\]\\}",
+        "contentName": "string.other.odoc",
+        "patterns": [{ "include": "#markup" }]
```
mnxn, line 120: I think since this is considered _preformatted_ source code, it shouldn't be able to include and markup syntax.
You should remove this line.

syntaxes/odoc.json
mnxn: Actually I think this rule is actually unnecessary since the comment colors comes from the OCaml syntax.

You should be to replace this with just

```suggestion
    { "include": "#markup" },
```

syntaxes/ocaml.json
```diff
@@ -126,12 +126,12 @@
...
         },
         {
-          "comment": "ocamldoc comment",
+          "comment": "odoc comment",
           "name": "comment.doc.ocaml",
           "begin": "\\(\\*\\*",
           "end": "\\*\\)",
           "patterns": [
-            { "include": "source.ocaml.ocamldoc#markup" },
+            { "include": "source.ocaml.odoc#markup" },
```
mnxn, line 134: If you apply the change from https://github.com/ocamllabs/vscode-ocaml-platform/pull/756/files#r756447787, you should make this just

```suggestion
            { "include": "source.ocaml.odoc" },
```

This way the documentation tags will also be highlighted in OCaml odoc comments.

syntaxes/odoc.json
```diff
@@ -0,0 +1,171 @@
...
+      {
+        "comment": "cross-reference to an element",
+        "begin": "\\{\\!",
+        "end": "\\}",
+        "contentName": "string.other.odoc",
+        "patterns": [{ "include": "#markup" }]
+      },
+      {
+        "comment": "index table for given modules",
+        "begin": "\\{\\![a-z]+\\:",
```
mnxn, line 145: I think this is supposed to be a literal `modules`

```suggestion
        "begin": "\\{\\!modules\\:",
```

syntaxes/odoc.json
```diff
@@ -0,0 +1,171 @@
...
+        "end": "\\]",
+        "contentName": "string.other.odoc",
+        "patterns": [{ "include": "#markup" }]
+      },
+      {
+        "comment": "insert a table of links",
+        "begin": "\\{\\!",
```
mnxn, line 152: Missing the `indexlist` keyword.

```suggestion
        "begin": "\\{\\!indexlist",
```

mnxn:
A few more general comments:

- The syntax is currently missing rules for [`-`, `+`, and `li` elements for lists](https://ocaml.org/manual/ocamldoc.html#sss:ocamldoc-list). You can add these to the `markup` section because of the shortcut mentioned in the documentation. Make sure you add the `+` list elements too.

- You can improve the highlighting of cross references by adding the [Cross-reference formatting](https://ocaml.org/manual/ocamldoc.html#sss:ocamldoc-crossref) inside `{! ... }`

- It would be easier to notice the formatting elements if there is keyword highlighting for the symbols. For example, the `b` in `{b ...}` and the `^` in `{^ ... }` should be highlighted. 
  You will need to add parentheses around the symbols in the `begin` regex patterns and add an entry with a `name` in `beginCaptures`. I think the name `keyword.operator.odoc` would be appropriate.

Issue #393, opened 2020-10-07
Port tests to OCaml


rgrinberg:
Now that we have a full blown bindings for vscode, it would be nice to add bindings for the test package as well. This will allow us to get rid of the remaining JS in our tests

smorimoto:
This definitely seems to be worth doing.

PR #762, opened 2021-11-02
convert test runners to ocaml

patricoferris:
This PR converts the three vscode-test runners to OCaml by adding a few extra bindings to the various JS packages. This is in partial fulfilment of #393 (it doesn't bind `mocha` and the like, this could be a follow-up PR perhaps or I can revisit this PR next week to add that). 

cc: @tmattio


src-bindings/fs-extra/fsextra_stub.js
```diff
@@ -0,0 +1 @@
+joo_global_object.fsExtra = require('fs-extra')
```
tmattio, line 1: It's a shame we have to add a dependency and bindings to the library for a single function. Would it be easy to implement copySync with of standard NodeJS functions?

tmattio:
Thanks @patricoferris, this is great!

> This is in partial fulfilment of #393 (it doesn't bind mocha and the like, this could be a follow-up PR perhaps or I can revisit this PR next week to add that).

Up to you, we can leave it until next week open if you want to continue working on it, but I'm fine with merging as-is.

src-bindings/fs-extra/fsextra_stub.js
patricoferris: Thanks for the quick review :))

https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy/copy-sync.js 🤷 

To reduce the number of files perhaps bundling it into the existing `FS` bindings might be a good idea seeing as they're already an altered form of the "true" Node FS module ?

tmattio: > To reduce the number of files perhaps bundling it into the existing FS bindings might be a good idea seeing as they're already an altered form of the "true" Node FS module ?

Sure, that sounds good 🙂

ulugbekna:
> To reduce the number of files perhaps bundling it into the existing FS
bindings might be a good idea seeing as they're already an altered form of
the "true" Node FS module

Could you, please, then leave a comment explaining why you're doing so
to avoid confusion in the future?

src-bindings/fs-extra/fsextra_stub.js
patricoferris: Actually, unless I'm mistaken, wouldn't this require an extra dependency for the package as `fs-extra` is only a `dev` dependency at the moment https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/package.json#L966 ? Maybe not worth it then?

test/run_esy_tests.ml
```diff
@@ -0,0 +1,44 @@
...
+  Path.join [ root; ".."; ".."; "test"; "fixtures"; "sample-esy" ]
+
+let project_path = Path.join [ Os.tmpdir (); "sample-esy" ]
+
+let esy_sandbox_config root =
+  let config =
+    Ojs.obj
+      [| ("root", Ojs.string_to_js root); ("kind", Ojs.string_to_js "esy") |]
+  in
+  let t = Ojs.obj [| ("ocaml.sandbox", config) |] in
```
mnxn, line 19: You should use `Jsonoo.Encode.object_` instead of the `Ojs` functions.

src-bindings/fs-extra/fsextra_stub.js
mnxn: Do we even need this dependency?

I noticed you added `Fs.Sync.copy`, can't that be used instead of `Fs_extra.copy_sync`?

.github/workflows/main.yml
```diff
@@ -69,7 +69,7 @@ jobs:
       - name: Test extension
         uses: GabrielBB/xvfb-action@v1
         with:
-          run: npm run test
+          run: make test
```
smorimoto, line 72:
```suggestion
          run: opam exec -- make test
```

src-bindings/fs-extra/fsextra_stub.js
patricoferris: IIUC `Fs_extra.copy_sync` is used because it copied directories (not just that it is synchronous).

mnxn: Okay in that case I personally think we should leave it as-is.

> It's a shame we have to add a dependency and bindings to the library for a single function.

@tmattio: It's not _adding_ a dependency since the JS version of the tests also used this library. 
We can remove the dependency later in another PR if necessary.

> unless I'm mistaken, wouldn't this require an extra dependency for the package as fs-extra is only a dev dependency at the moment

That's right. That means that keeping the bindings to `fs-extra` separate would not require us to add it as a runtime dependency.

rgrinberg:
@patricoferris is this PR ready?

patricoferris:
Sorry for the delay doing some more testing locally first to debug https://github.com/ocamllabs/vscode-ocaml-platform/runs/4082355161?check_suite_focus=true

rgrinberg:
Please re-open if you ever get around to do this.

PR #769, opened 2021-11-10
More principled handling of ocaml-lsp version in the extension

### LINKED ISSUE #460: Improve ocaml-lsp version handling by the extension
**Problem:** 

The extension currently warns the user that their ocaml-lsp version may be outdated and recommends installing the new version, when the user triggers a feature already present in the extension (because vscode updates extensions automatically) but not in ocaml-lsp, which isn't updated automatically. For example, if user does `alt+o` to switch impl/intf AND their ocaml-lsp doesn't have `handleSwitchImplIntf` (IIRC) capability, there is a warning. 

However, the extension comes up with warnings based on ocaml-lsp's capabilities, which may be cumbersome and error-prone (for example, if I understand correctly, with the new code action "infer interface", user on "alt+p" gets just "no infer interface code action available" instead of a warning message "update to get that feature" with a button to update, preferably).

**Solution:**

Can be done now: 

- [ ] Update code that handles warnings to depend on ocaml-lsp's version (and automatically fix "infer interface" code action problem) 

Can be done in future:

- [ ] after every ocaml-lsp release on opam, release the extension with some code that warns the user on ocaml-lsp launch to update if their version is not the latest (or even update automatically if there are no package manager conflicts, which should be the case, since we vendor everything)

ulugbekna:
We looked whether ocaml-lsp supports a certain experimental/custom capability, if missing, we suggested to upgrade. However, suggesting upgrading is incorrect in cases when a user cannot upgrade because they use the latest ocamllsp available for their ocaml version. 

I would appreciate a closer look since it is related to the extension startup and some mutability and mutation order.

Closes #460 and removes a blocker for a release (there are still problems with the ppx viewer though)


src/extension_instance.ml
```diff
@@ -3,6 +3,8 @@ open Import
 type t =
   { mutable sandbox : Sandbox.t
   ; mutable repl : Terminal_sandbox.t option
+  ; mutable ocaml : Ocaml.t option
+        (** invariant: must be set before initializing the language server *)
```
rgrinberg, line 7: Do we check this somewhere in the code?

src/ocaml.ml
```diff
@@ -0,0 +1,17 @@
+open Import
+
+type t = Ocaml_version.t
```
rgrinberg, line 3: do we really need this type? and this entire module for that matter. It is only used once so just inline it there.

src/ocaml_lsp.ml
```diff
@@ -1,65 +1,105 @@
...
+    ; handleSwitchImplIntf = false
+    ; handleInferIntf = false
+    ; handleTypedHoles = false
+    }
+
+  (** Creates [t] given a JSON of form [{ 'handleSwitchImplIntf' : true, .... }] *)
+  let t_of_json (json : Jsonoo.t) =
+    let experimental_caps_tbl = Jsonoo.Decode.(dict bool) json in
+    let has_capability cap =
+      Stdlib.Hashtbl.find_opt experimental_caps_tbl cap
```
rgrinberg, line 22: Why don't we use base here?

src/ocaml_lsp.ml
```diff
@@ -1,65 +1,105 @@
 open! Import
 
+module Experimental_capabilities = struct
+  type t =
+    { interfaceSpecificLangId : bool
+    ; handleSwitchImplIntf : bool
+    ; handleInferIntf : bool
+    ; handleTypedHoles : bool
+    }
```
rgrinberg, line 9: This module is a good idea. Eventually, we need to move it to ocamllsp and then just use it here.

src/ocaml_lsp.ml
```diff
@@ -1,65 +1,105 @@
...
+      | None -> Error `ServerInfo_version_missing
+      | Some v -> Ok v)
 
-let default_field key decode default json =
-  let open Jsonoo.Decode in
-  try_default default (field key decode) json
+let get_version_semver t =
+  match get_version_from_serverInfo t with
+  | Ok v -> (
+    match String.split v ~on:'-' |> List.hd with
```
rgrinberg, line 72: lsplit2

src/import.ml
```diff
@@ -165,4 +165,62 @@ module Range = struct
   include Vscode.Range
 end
 
+module Promise = struct
+  let don't_wait_for (_p : _ Promise.t) = ()
```
rgrinberg, line 169: is this being used anywhere?

src/import.ml
```diff
@@ -165,4 +165,62 @@ module Range = struct
...
+    fmt
+
+module Ocaml_version = struct
+  let ( > ) v1 v2 = Ocaml_version.compare v1 v2 = 1
+
+  let ( < ) v1 v2 = Ocaml_version.compare v1 v2 = -1
+
+  let ( <= ) v1 v2 =
+    let r = Ocaml_version.compare v1 v2 in
+    r = -1 || r = 0
```
rgrinberg, line 213: `r <= 0` and inline the `r` variable.

src/import.ml
```diff
@@ -165,4 +165,62 @@ module Range = struct
...
+
+  let ( >= ) v1 v2 =
+    let r = Ocaml_version.compare v1 v2 in
+    r = 1 || r = 0
```
rgrinberg, line 217: ditto

src/ocaml_lsp.ml
```diff
@@ -1,65 +1,105 @@
...
-  { interfaceSpecificLangId = false
-  ; handleSwitchImplIntf = false
-  ; handleInferIntf = false
-  ; handleTypedHoles = false
-  }
+let get_version_from_serverInfo { serverInfo; experimental_capabilities = _ } =
+  match serverInfo with
+  | None -> Error `Missing_serverInfo (* ocamllsp should have [serverInfo] *)
+  | Some { name; version } -> (
+    if not (String.equal name "ocamllsp") then (
```
rgrinberg, line 59: this `if` can be inverted

src/ocaml_lsp.mli
```diff
@@ -4,7 +4,19 @@ type t
...
+val is_version_up_to_date :
+     t
+  -> Ocaml_version.t
+  -> ( bool
+     , [> `Ocaml_version_not_supported of Ocaml_version.t
+       | `Unexpected of
+         [> `Language_server_isn't_ocamllsp
+         | `Missing_serverInfo
+         | `Unable_to_parse_version
+         | `ServerInfo_version_missing
```
rgrinberg, line 16: Some of these error constructors are only used for error construction. In these cases, we shouldn't even have these constructors as all they accomplish is leaking implementation information. I haven't looked closely but I'm pretty sure at least the last 3 constructors can be eliminated into `Message of string`.

rgrinberg:
Looks OK. @mnxn can you take a glance as well? it's a rather large PR.

src/extension_instance.ml
ulugbekna: We access that field with `Option.value_exn`, which is kind of a check. Do you have a preferred place you'd like it checked?

src/ocaml_lsp.ml
ulugbekna: `Jsonoo.Decode.dict` returns a `Stdlib.Hashtbl.t`, which is incompatible with `Base.Hashtbl.t`

ulugbekna:
@rgrinberg 

1. Currently, we compare versions by `String.equal ocamllsp_v "1.9.0"`, but will results in "out-of-date ocamllsp" for dev versions of olsp such as `1.9.0-56-g4d8b660`. I was thinking of doing `String.is_prefix ocamllsp_v "1.9.0"` instead, wdyt? 
2. Currently, when I `make install` ocaml-lsp, its version is still `1.8.3-...` . As far as I understand, dune-build-info runs `git describe --always --dirty --abbrev=7`, which "finds the most recent tag that is reachable from a commit." Is it the way we tag commits prevents from the version being `1.9.0-...`?

rgrinberg:
1. What about `1.9.x`? What does `1.9.0 <= 1.9.0-56-g4d8b660 < 1.10.0` according to the rules of semver? Perhaps we can use one of the many semver libraries  out there for this?

2. Yes, tagging a commit removes the `-...` junk.

src/extension_instance.ml
rgrinberg: I see. I suppose that's fine then.

src/ocaml_lsp.ml
ulugbekna: I'm afraid:
`String.lsplit2 "1.9.0" ~on:'-'` evaluates to `None` (because `~on` doesn't occur there), but we want 
`String.split "1.9.0" ~on:'-' |> List.hd` which evaluates to `Some "1.9.0`

ulugbekna:
> What about 1.9.x? What does 1.9.0 <= 1.9.0-56-g4d8b660 < 1.10.0 according to the rules of semver? 

If I understand what you mean by `1.9.x` correctly, the extension won't warn about new patch versions of OLSP. I'd prefer to warn about patch releases as well. 

I'm not sure the current versioning we use for dev versions follows semver spec for dev versions: https://semver.org/#spec-item-10

I'll replace with string prefix matching for now. Let me know if you prefer another way of handling versions.

ulugbekna:
Uh, sorry for the noise, `get_version_semver` is already splitting, so we _can_ actually do `String.equal`. I was thrown off by warnings on the latest OLSP, but it still has a version `1.8.3-...`

ulugbekna:
I've addressed all of the comments

rgrinberg (approved):

PR #771, opened 2021-11-11
Offline doc generation

tatchi:
**Opening that is one as draft to get some feedback and discuss the way to move forward (see some of my questions below). I also still need to write some tests.**

The purpose of this PR is to make it possible to generate and view the documentation for installed packages in the opam switch right from the VSCode editor. It uses [odig](https://github.com/b0-system/odig) to generate the HTML documentation for a given package and rely on the [vscode live preview](https://github.com/microsoft/vscode-livepreview) extension to display the generated documentation files. Both need to be installed to make it work.

A new icon (next to the existing one) added to the tree view allows to generate+show the documentation for a package:

https://user-images.githubusercontent.com/5595092/141298715-aaeb9234-29d2-47be-b8c5-6344a747d1dc.mov

Like I said above, `odig` and the `vscode live preview` extension have to be installed. We check that both are indeed available with an error message being displayed if that's not the case. Also, only **opam** sandboxes are supported for now.

I think it works quite well excepts for two things I would like to improve (and for which I would like to have your input):

1. The `vscode live preview` extension seems to freeze when the documentation for a package is generated while it's already showing documentation for another package (basically when we switch between package's doc). The extension probably has a file watcher to [reload the preview](https://github.com/microsoft/vscode-livepreview#live-refreshing) in live while you edit the code. Maybe that watcher is having some trouble when a couple of files are being added? 

https://user-images.githubusercontent.com/5595092/141302096-30603c8b-30d8-43b0-aa38-223798feea29.mp4

2. The `vscode live preview` internal server can't display files that are outside of the currently opened workspace.

It means that it won't work with the default opam switches since the documentation will live somewhere under `~/.opam/...`. As suggested by @ulugbekna, we could try to add the documentation folder in the workspace. There's actually a function `updateWorkspaceFolders` that allows adding folders to the current workspace. That's what I use but it doesn't work well because if we're not yet in a workspace, the extension is restarted. So the code that triggers the command to show the documentation via the live preview extension is not even executed (the extension is killed before).

From the doc:

> If the first workspace folder is added, removed or changed, the currently executing extensions (including the one that called this method) will be terminated and restarted so that the (deprecated) rootPath property is updated to point to the first workspace folder.

A possible solution could be to set some state before we're about the add a new folder to the workspace (and the extension being restarted) and then in the `onDidChangeWorkspaceFolders` we can check that state and eventually show the documentation.

Another solution I've experimented with is to not use the live preview extension. Instead, we could start our own server that serves the documentation files and display them via the built-in `Simple Browser` extension. I've used [polka](https://github.com/lukeed/polka) (lightweight server) together with [sirv](https://github.com/lukeed/sirv) (middleware to serve static files) for my experiment.

The cool thing is that it would not require any external extension to be installed and it would also solve both 1 and 2. See the video below (I'm on the default opam switch)

https://user-images.githubusercontent.com/5595092/141309271-28b8fbb8-4aae-4e49-956e-1fdd3826a933.mov

The `Simple Browser` has fewer features though. For instance, we can't search (CTRL+F) across the page, which might be handy when you're looking up something.

### DIFF
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -22,11 +22,7 @@ module Server = struct
     | `Error f -> on t "error" @@ [%js.of: err:Node.JsError.t -> unit] f
 end
 
-type polka = Ojs.t
-
-let polka_of_js = Ojs.t_of_js
-
-let polka_to_js = Ojs.t_to_js
+type polka = Ojs.t [@@js]
 
 module Middleware = struct
   module Request = struct
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -13,7 +13,7 @@ module Server = struct
     [%js:
     val close : t -> t [@@js.call]
 
-    val address : t -> Address.t [@@js.call]
+    val address : t -> Address.t or_undefined [@@js.call]
 
     val on : t -> string -> Ojs.t -> unit [@@js.call]]
 
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -1,9 +1,5 @@
 open Interop
 
-module Middleware = struct
-  include Interface.Make ()
-end
-
 module Server = struct
   include Interface.Make ()
 
@@ -18,31 +14,48 @@ module Server = struct
     | `Error f -> on t "error" @@ [%js.of: err:Node.JsError.t -> unit] f
 end
 
-module Sirv = struct
-  include
-    [%js:
-    type t = Middleware.t
+type polka = Ojs.t
 
-    val serve : string -> t [@@js.global "sirv"]]
+let polka_of_js = Ojs.t_of_js
+
+let polka_to_js = Ojs.t_to_js
+
+module Middleware = struct
+  module Request = struct
+    include Interface.Make ()
+  end
+
+  module Response = struct
+    include Interface.Make ()
+  end
+
+  type t =
+    request:Request.t -> response:Response.t -> next:(unit -> polka) -> polka
+  [@@js]
 end
 
 include
   [%js:
-  type t
+  val create : unit -> polka [@@js.global "polka"]
 
-  val create : unit -> t [@@js.global "polka"]
-
-  val listen_ : t -> int -> ?callback:(unit -> unit) -> unit -> t
+  val listen_ : polka -> int -> ?callback:(unit -> unit) -> unit -> polka
     [@@js.call "listen"]
 
-  val get_ : t -> string -> (unit -> unit) -> t [@@js.call "get"]
+  val get_ : polka -> string -> (unit -> unit) -> polka [@@js.call "get"]
 
-  val use_ : t -> (Middleware.t list[@js.variadic]) -> t [@@js.call "use"]
+  val use_ : polka -> (Middleware.t list[@js.variadic]) -> polka
+    [@@js.call "use"]
 
-  val server : t -> Server.t [@@js.get]]
+  val server : polka -> Server.t [@@js.get]]
 
 let get path callback t = get_ t path callback
 
 let listen port ?callback t = listen_ t port ?callback
 
 let use middlewares t = use_ t middlewares
+
+module Sirv = struct
+  include
+    [%js:
+    val serve : string -> (Middleware.t[@js.dummy]) [@@js.global "sirv"]]
+end
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -7,7 +7,15 @@ end
 module Server = struct
   include Interface.Make ()
 
-  include [%js: val close : t -> t [@@js.call]]
+  include
+    [%js:
+    val close : t -> t [@@js.call]
+
+    val on : t -> string -> Ojs.t -> unit [@@js.call]]
+
+  let on t = function
+    | `Close f -> on t "close" @@ [%js.of: unit -> unit] f
+    | `Error f -> on t "error" @@ [%js.of: err:Node.JsError.t -> unit] f
 end
 
 module Sirv = struct
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -1,0 +1,40 @@
+open Interop
+
+module Middleware = struct
+  include Interface.Make ()
+end
+
+module Server = struct
+  include Interface.Make ()
+
+  include [%js: val close : t -> t [@@js.call]]
+end
+
+module Sirv = struct
+  include
+    [%js:
+    type t = Middleware.t
+
+    val serve : string -> t [@@js.global "sirv"]]
+end
+
+include
+  [%js:
+  type t
+
+  val create : unit -> t [@@js.global "polka"]
+
+  val listen_ : t -> int -> ?callback:(unit -> unit) -> unit -> t
+    [@@js.call "listen"]
+
+  val get_ : t -> string -> (unit -> unit) -> t [@@js.call "get"]
+
+  val use_ : t -> (Middleware.t list[@js.variadic]) -> t [@@js.call "use"]
+
+  val server : t -> Server.t [@@js.get]]
+
+let get path callback t = get_ t path callback
+
+let listen port ?callback t = listen_ t port ?callback
+
+let use middlewares t = use_ t middlewares
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -59,7 +59,15 @@ let listen port ?callback t = listen_ t port ?callback
 let use middlewares t = use_ t middlewares
 
 module Sirv = struct
+  module Options = struct
+    include Interface.Make ()
+
+    include [%js: val create : dev:bool -> t [@@js.builder]]
+  end
+
   include
     [%js:
-    val serve : string -> (Middleware.t[@js.dummy]) [@@js.global "sirv"]]
+    val serve :
+      string -> ?options:Options.t -> unit -> (Middleware.t[@js.dummy])
+      [@@js.global "sirv"]]
 end
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -7,6 +7,8 @@ module Server = struct
     include Interface.Make ()
 
     include [%js: val port : t -> int [@@js.get]]
+
+    include [%js: val address : t -> string [@@js.get]]
   end
 
   include
@@ -44,7 +46,13 @@ include
   [%js:
   val create : unit -> polka [@@js.global "polka"]
 
-  val listen_ : polka -> int -> ?callback:(unit -> unit) -> unit -> polka
+  val listen_ :
+       polka
+    -> int
+    -> ?hostname:string
+    -> ?callback:(unit -> unit)
+    -> unit
+    -> polka
     [@@js.call "listen"]
 
   val get_ : polka -> string -> (unit -> unit) -> polka [@@js.call "get"]
@@ -56,7 +64,7 @@ include
 
 let get path callback t = get_ t path callback
 
-let listen port ?callback t = listen_ t port ?callback
+let listen port ?hostname ?callback t = listen_ t port ?hostname ?callback
 
 let use middlewares t = use_ t middlewares
 
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -11,7 +11,9 @@ module Server = struct
 
   include
     [%js:
-    val close : t -> t [@@js.call]
+    val close :
+      t -> ?callback:(Node.JsError.t or_undefined -> unit) -> unit -> t
+      [@@js.call "close"]
 
     val address : t -> Address.t or_undefined [@@js.call]
 
--- a/src-bindings/polka/polka.ml
+++ b/src-bindings/polka/polka.ml
@@ -3,9 +3,17 @@ open Interop
 module Server = struct
   include Interface.Make ()
 
+  module Address = struct
+    include Interface.Make ()
+
+    include [%js: val port : t -> int [@@js.get]]
+  end
+
   include
     [%js:
     val close : t -> t [@@js.call]
+
+    val address : t -> Address.t [@@js.call]
 
     val on : t -> string -> Ojs.t -> unit [@@js.call]]
 
--- a/src-bindings/vscode/vscode.mli
+++ b/src-bindings/vscode/vscode.mli
@@ -1809,7 +1809,6 @@ module Workspace : sig
        start:int
     -> deleteCount:int or_undefined
     -> workspaceFoldersToAdd:(workspaceFolderToAdd list[@js.variadic])
-    -> unit
     -> bool
 end
 
--- a/src-bindings/vscode/vscode.mli
+++ b/src-bindings/vscode/vscode.mli
@@ -1106,8 +1106,6 @@ end
 
 module Extensions : sig
   val getExtension : string -> Extension.t or_undefined
-
-  val is_extension_installed : string -> bool
 end
 
 module TerminalExitStatus : sig
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -28,5 +28,13 @@ let port t =
 let on_close t ~f = Polka.Server.on t (`Close f)
 
 let stop t =
-  let (_ : Polka.Server.t) = Polka.Server.close t in
+  Promise.make @@ fun ~resolve ~reject:_ ->
+  let (_ : Polka.Server.t) =
+    Polka.Server.close t
+      ~callback:(fun error ->
+        match error with
+        | None -> resolve (Ok ())
+        | Some error -> resolve (Error error))
+      ()
+  in
   ()
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -7,11 +7,11 @@ type t = Polka.Server.t
     scheduled timeout is overridden *)
 let debouncing_middleware ~timeout_in_ms f =
   let timeout = ref None in
-  fun ~request:_ ~response:_ ~next ->
-    (* on every request, we clear previous timeout and set a new one *)
-    Option.iter !timeout ~f:Node.clearTimeout;
-    timeout := Some (Node.setTimeout f timeout_in_ms);
-    next ()
+  Staged.stage (fun ~request:_ ~response:_ ~next ->
+      (* on every request, we clear previous timeout and set a new one *)
+      Option.iter !timeout ~f:Node.clearTimeout;
+      timeout := Some (Node.setTimeout f timeout_in_ms);
+      next ())
 
 let start ~path ?(port = 0) () =
   Promise.make @@ fun ~resolve ~reject:_ ->
@@ -26,7 +26,7 @@ let start ~path ?(port = 0) () =
   let serve =
     polka
     |> Polka.use
-         [ debouncing_server_termination_mdlw
+         [ Staged.unstage debouncing_server_termination_mdlw
          ; Polka.Sirv.serve (path |> Path.to_string) ~options ()
          ]
     |> Polka.listen port ~callback:(fun () ->
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -1,6 +1,11 @@
 open Import
 
-type t = Polka.Server.t option ref
+type t' =
+  { server : Polka.Server.t
+  ; path : Path.t
+  }
+
+type t = t' option ref
 
 let start ~path =
   Promise.make @@ fun ~resolve ~reject:_ ->
@@ -11,19 +16,21 @@ let start ~path =
     |> Polka.use [ Polka.Sirv.serve (path |> Path.to_string) ~options () ]
     |> Polka.listen 0 ~callback:(fun () ->
            let server = Polka.server polka in
-           resolve (Ok (ref (Some server))))
+           resolve (Ok (ref (Some { server; path }))))
   in
   let polka = serve () in
   let server = Polka.server polka in
   Polka.Server.on server (`Error (fun ~err -> resolve (Error err)))
 
-let get t =
+let get (t : t) =
   match !t with
   | None -> failwith "document server: already disposed"
   | Some t -> t
 
+let path (t : t) = (get t).path
+
 let port t =
-  match Polka.Server.address (get t) with
+  match Polka.Server.address (get t).server with
   | None ->
     (* if it's [None], server must not be listening, but we aim to have server
        only if it's listening *)
@@ -32,17 +39,17 @@ let port t =
 
 (* TODO extract this from the server somehow *)
 let host t =
-  ignore (get t : Polka.Server.t);
+  ignore (get t : t');
   "localhost"
 
-let dispose t =
+let dispose (t : t) =
   match !t with
   | None -> Disposable.make ~dispose:(fun () -> ())
   | Some server ->
     t := None;
     Disposable.make ~dispose:(fun () ->
         let (_ : Polka.Server.t) =
-          Polka.Server.close server
+          Polka.Server.close server.server
             ~callback:(fun error ->
               match error with
               | None -> ()
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -22,11 +22,12 @@ let start ~path ?(port = 0) () =
       (fun () ->
         ignore @@ (Polka.Server.close (Polka.server polka) : Polka.Server.t))
   in
+  let options = Polka.Sirv.Options.create ~dev:true in
   let serve =
     polka
     |> Polka.use
          [ debouncing_server_termination_mdlw
-         ; Polka.Sirv.serve (path |> Path.to_string)
+         ; Polka.Sirv.serve (path |> Path.to_string) ~options ()
          ]
     |> Polka.listen port ~callback:(fun () ->
            let server = Polka.server polka in
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -16,6 +16,7 @@ let debouncing_middleware ~timeout_in_ms f =
 let start ~path ?(port = 0) () =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let polka = Polka.create () in
+  (* Closes the server after some idle time *)
   let debouncing_server_termination_mdlw =
     debouncing_middleware
       ~timeout_in_ms:(60 * 10 * 1000)
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -28,14 +28,16 @@ let port t =
 (* TODO extract this from the server somehow *)
 let host _ = "localhost"
 
-let stop t =
-  Promise.make @@ fun ~resolve ~reject:_ ->
-  let (_ : Polka.Server.t) =
-    Polka.Server.close t
-      ~callback:(fun error ->
-        match error with
-        | None -> resolve (Ok ())
-        | Some error -> resolve (Error error))
-      ()
-  in
-  ()
+let dispose t =
+  Disposable.make ~dispose:(fun () ->
+      let (_ : Polka.Server.t) =
+        Polka.Server.close t
+          ~callback:(fun error ->
+            match error with
+            | None -> ()
+            | Some error ->
+              show_message `Warn "Error closing server: %s"
+                (Ojs.string_of_js (Promise.error_to_js error)))
+          ()
+      in
+      ())
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -1,9 +1,6 @@
 open Import
 
-type t =
-  { server : Polka.Server.t
-  ; port : int
-  }
+type t = Polka.Server.t
 
 let start ~path ?(port = 0) () =
   Promise.make @@ fun ~resolve ~reject:_ ->
@@ -34,19 +31,22 @@ let start ~path ?(port = 0) () =
          [ timeout_middleware; Polka.Sirv.serve (path |> Path.to_string) ]
     |> Polka.listen port ~callback:(fun () ->
            let server = Polka.server polka in
-           let address = Polka.Server.address server in
-           let port = Polka.Server.Address.port address in
-           resolve (Ok { server; port }))
+           resolve (Ok server))
   in
   let polka = serve () in
   let server = Polka.server polka in
-  let () = Polka.Server.on server (`Error (fun ~err -> resolve (Error err))) in
-  ()
+  Polka.Server.on server (`Error (fun ~err -> resolve (Error err)))
 
-let port t = t.port
+let port t =
+  match Polka.Server.address t with
+  | None ->
+    (* if it's [None], server must not be listening, but we aim to have server
+       only if it's listening *)
+    assert false
+  | Some a -> Polka.Server.Address.port a
 
-let on_close t ~f = Polka.Server.on t.server (`Close f)
+let on_close t ~f = Polka.Server.on t (`Close f)
 
 let stop t =
-  let (_ : Polka.Server.t) = Polka.Server.close t.server in
+  let (_ : Polka.Server.t) = Polka.Server.close t in
   ()
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -1,6 +1,6 @@
 open Import
 
-type t = Polka.Server.t
+type t = Polka.Server.t option ref
 
 let start ~path =
   Promise.make @@ fun ~resolve ~reject:_ ->
@@ -11,14 +11,19 @@ let start ~path =
     |> Polka.use [ Polka.Sirv.serve (path |> Path.to_string) ~options () ]
     |> Polka.listen 0 ~callback:(fun () ->
            let server = Polka.server polka in
-           resolve (Ok server))
+           resolve (Ok (ref (Some server))))
   in
   let polka = serve () in
   let server = Polka.server polka in
   Polka.Server.on server (`Error (fun ~err -> resolve (Error err)))
 
+let get t =
+  match !t with
+  | None -> failwith "document server: already disposed"
+  | Some t -> t
+
 let port t =
-  match Polka.Server.address t with
+  match Polka.Server.address (get t) with
   | None ->
     (* if it's [None], server must not be listening, but we aim to have server
        only if it's listening *)
@@ -26,18 +31,24 @@ let port t =
   | Some a -> Polka.Server.Address.port a
 
 (* TODO extract this from the server somehow *)
-let host _ = "localhost"
+let host t =
+  ignore (get t : Polka.Server.t);
+  "localhost"
 
 let dispose t =
-  Disposable.make ~dispose:(fun () ->
-      let (_ : Polka.Server.t) =
-        Polka.Server.close t
-          ~callback:(fun error ->
-            match error with
-            | None -> ()
-            | Some error ->
-              show_message `Warn "Error closing server: %s"
-                (Ojs.string_of_js (Promise.error_to_js error)))
-          ()
-      in
-      ())
+  match !t with
+  | None -> Disposable.make ~dispose:(fun () -> ())
+  | Some server ->
+    t := None;
+    Disposable.make ~dispose:(fun () ->
+        let (_ : Polka.Server.t) =
+          Polka.Server.close server
+            ~callback:(fun error ->
+              match error with
+              | None -> ()
+              | Some error ->
+                show_message `Warn "Error closing server: %s"
+                  (Ojs.string_of_js (Promise.error_to_js error)))
+            ()
+        in
+        ())
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -1,0 +1,28 @@
+open Import
+
+type t =
+  { server : Polka.Server.t
+  ; port : int
+  }
+
+let start ~port ~path =
+  Promise.make @@ fun ~resolve ~reject:_ ->
+  let polka = Polka.create () in
+  let serve =
+    polka
+    |> Polka.use [ Polka.Sirv.serve (path |> Path.to_string) ]
+    |> Polka.listen port ~callback:(fun () ->
+           resolve (Ok { server = Polka.server polka; port }))
+  in
+  let polka = serve () in
+  let server = Polka.server polka in
+  let () = Polka.Server.on server (`Error (fun ~err -> resolve (Error err))) in
+  ()
+
+let port t = t.port
+
+let on_close t ~f = Polka.Server.on t.server (`Close f)
+
+let stop t =
+  let (_ : Polka.Server.t) = Polka.Server.close t.server in
+  ()
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -25,6 +25,9 @@ let port t =
     assert false
   | Some a -> Polka.Server.Address.port a
 
+(* TODO extract this from the server somehow *)
+let host _ = "localhost"
+
 let stop t =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let (_ : Polka.Server.t) =
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -25,8 +25,6 @@ let port t =
     assert false
   | Some a -> Polka.Server.Address.port a
 
-let on_close t ~f = Polka.Server.on t (`Close f)
-
 let stop t =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let (_ : Polka.Server.t) =
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -2,14 +2,14 @@ open Import
 
 type t = Polka.Server.t
 
-let start ~path ?(port = 0) () =
+let start ~path =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let polka = Polka.create () in
   let options = Polka.Sirv.Options.create ~dev:true in
   let serve =
     polka
     |> Polka.use [ Polka.Sirv.serve (path |> Path.to_string) ~options () ]
-    |> Polka.listen port ~callback:(fun () ->
+    |> Polka.listen 0 ~callback:(fun () ->
            let server = Polka.server polka in
            resolve (Ok server))
   in
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -2,33 +2,32 @@ open Import
 
 type t = Polka.Server.t
 
+(** [debouncing_middleware ~timeout_in_ms f] on every request schedules [f] to
+    be run in [timeout_in_ms] milliseconds; on every new request the last
+    scheduled timeout is overridden *)
+let debouncing_middleware ~timeout_in_ms f =
+  let timeout = ref None in
+  fun ~request:_ ~response:_ ~next ->
+    (* on every request, we clear previous timeout and set a new one *)
+    Option.iter !timeout ~f:Node.clearTimeout;
+    timeout := Some (Node.setTimeout f timeout_in_ms);
+    next ()
+
 let start ~path ?(port = 0) () =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let polka = Polka.create () in
-
-  let timeout_ref = ref None in
-
-  let timeout_middleware ~request:_ ~response:_ ~next =
-    let () =
-      Option.iter !timeout_ref ~f:Node.clearTimeout;
-      timeout_ref :=
-        Some
-          (Node.setTimeout
-             (fun () ->
-               let (_ : Polka.Server.t) =
-                 Polka.Server.close (Polka.server polka)
-               in
-               ())
-             (* 10 minutes (in ms) *)
-             (60 * 10 * 1000))
-    in
-    next ()
+  let debouncing_server_termination_mdlw =
+    debouncing_middleware
+      ~timeout_in_ms:(60 * 10 * 1000)
+      (fun () ->
+        ignore @@ (Polka.Server.close (Polka.server polka) : Polka.Server.t))
   in
-
   let serve =
     polka
     |> Polka.use
-         [ timeout_middleware; Polka.Sirv.serve (path |> Path.to_string) ]
+         [ debouncing_server_termination_mdlw
+         ; Polka.Sirv.serve (path |> Path.to_string)
+         ]
     |> Polka.listen port ~callback:(fun () ->
            let server = Polka.server polka in
            resolve (Ok server))
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -14,7 +14,7 @@ let start ~path =
   let serve =
     polka
     |> Polka.use [ Polka.Sirv.serve (path |> Path.to_string) ~options () ]
-    |> Polka.listen 0 ~callback:(fun () ->
+    |> Polka.listen 0 ~hostname:"localhost" ~callback:(fun () ->
            let server = Polka.server polka in
            resolve (Ok (ref (Some { server; path }))))
   in
@@ -29,18 +29,17 @@ let get (t : t) =
 
 let path (t : t) = (get t).path
 
-let port t =
+let address t =
   match Polka.Server.address (get t).server with
   | None ->
     (* if it's [None], server must not be listening, but we aim to have server
        only if it's listening *)
     assert false
-  | Some a -> Polka.Server.Address.port a
+  | Some a -> a
 
-(* TODO extract this from the server somehow *)
-let host t =
-  ignore (get t : t');
-  "localhost"
+let port t = Polka.Server.Address.port (address t)
+
+let host t = Polka.Server.Address.address (address t)
 
 let dispose (t : t) =
   match !t with
--- a/src/documentation_server.ml
+++ b/src/documentation_server.ml
@@ -8,9 +8,30 @@ type t =
 let start ~port ~path =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let polka = Polka.create () in
+
+  let timeout_ref = ref None in
+
+  let timeout_middleware ~request:_ ~response:_ ~next =
+    let () =
+      Option.iter !timeout_ref ~f:Node.clearTimeout;
+      timeout_ref :=
+        Some
+          (Node.setTimeout
+             (fun () ->
+               let (_ : Polka.Server.t) =
+                 Polka.Server.close (Polka.server polka)
+               in
+               ())
+             (* 10 minutes (in ms) *)
+             (60 * 10 * 1000))
+    in
+    next ()
+  in
+
   let serve =
     polka
-    |> Polka.use [ Polka.Sirv.serve (path |> Path.to_string) ]
+    |> Polka.use
+         [ timeout_middleware; Polka.Sirv.serve (path |> Path.to_string) ]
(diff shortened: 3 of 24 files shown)


src-bindings/vscode/vscode.mli
```diff
@@ -1761,6 +1771,18 @@ module Workspace : sig
...
     scheme:string -> provider:TextDocumentContentProvider.t -> Disposable.t
+
+  type workspaceFolderToAdd =
+    { name : string
+    ; uri : Uri.t
+    }
+
+  val updateWorkspaceFolders :
+       start:int
+    -> deleteCount:int or_undefined
```
tatchi, line 1782: Is it correct? It can actually be `null`, `undefined` or `number`

ulugbekna: I think so? @mnxn would know better 

Ideally, `deleteCount` would be an optional argument, ie `... -> ?deleteCount:int -> ...`

mnxn: When going from JS to OCaml, `or_undefined` will turn the `number` into a `Some` `int` and `null` | `undefined` into `None`.

When going from OCaml to JS, `or_undefined` will turn a `Some` `int` into a `number` and `None` into `undefined`.

https://github.com/ocamllabs/vscode-ocaml-platform/blob/7870a8ecaaac8a81bd59764956880f54e6e095e6/src-bindings/interop/interop.ml#L7-L17

This means it's not possible to pass a `null` for this argument.

That being said, I don't think there's any reason to be able to pass both. The [VSCode documentation](https://code.visualstudio.com/api/references/vscode-api#workspace.updateWorkspaceFolders) doesn't say that it handles `null`s any differently than `undefined` here.

> Ideally, deleteCount would be an optional argument

@ulugbekna: This can't be done because [according to gen_js_api](https://github.com/LexiFi/gen_js_api/blob/master/TYPES.md#arrow-types):

> If no default value is provided and the argument is missing, the argument is dropped from the list of arguments passed to the JS call

If that argument was dropped, it would mess up the variadic arguments that follow it. Additionally, the VSCode documentation doesn't mark this parameter as optional like in the other places where we use OCaml optional arguments.

I think this should be left as-is.

mnxn: Oh and since there are no optional arguments in this function signature, you can remove the `unit` parameter.

tmattio:
Thanks for the work @tatchi, this looks really nice! 🎉 

I'd be inclined to use the built-in browser to serve the documentation. If I'm not mistaken, we don't have any dependency on other extensions at the moment, and it's a nice feature to keep if possible (the fewer third parties to interact with, the better for maintainability and stability). In particular, it seems that the live preview extension brings its own set of problems. For the limitations you're mentioning, we could see if we have other solutions. For instance, a search feature is something that's been requested a lot for odoc-generated documentation, so we could address it there.

I'll try to review the code next week, but it looks good at first glance!

ulugbekna:
I can take the lead on reviewing and merging this feature in collaboration with Corentin, @tmattio, no worries. 

---

As for depending on another extension, the alternative is to write our own server and depend on even more things, AFAIU. I'd imagine that to be more error-prone than using an extension doing that from the vscode team, given extensions have public API that I'd imagine to not be broken too often. 

---

> For instance, a search feature is something that's been requested a lot for odoc-generated documentation, so we could address it there.

What does "search feature" mean here?

tmattio:
> As for depending on another extension, the alternative is to write our own server and depend on even more things

If that's the alternative, then you're probably right. I'm confused as to why we'd need all of that though? What's blocking us from serving the HTML page with the built in browser?

> What does "search feature" mean here?

I imagine it would allow you to input a string, and get redirected to a result page with the matching signatures and docstrings.

tatchi:
Thanks for all the feedbacks 🤗  

> i) Have you talked to the vscode live preview extension authors about the limitations you're facing, especially the freezes? Freezes seem like a bug or inefficiency that should be eliminated. They were quite responsive last time I filed an issue in that repo.

Good idea, I opened an [issue](https://github.com/microsoft/vscode-livepreview/issues/174) with a possible fix. Good news is that there's actually a [configuration](https://github.com/microsoft/vscode-livepreview/blob/main/package.json#L186) `livePreview.autoRefreshPreview` that determines how often the webview is refreshed. Changing that value to `Never` or `On Changes to Saved Files` fixes the issue.

> > It means that it won't work with the default opam switches since the documentation will live somewhere under ~/.opam/....
> 
> I think that if we can't get this to work without lengthy/bizarre workarounds, we stick to being simple stupid and
> 
> when we can open a documentation in vscode, we open it
> when we can't, we open it in browser (I assume we can do that) and let user know that we couldn't open the documentation inside the vscode because we're limited by the workspace folders. If the user then chooses to add their out-of-workspace switch to the workspace, then we support showing the documentation inside vscode. Does this make sense?

I agree with what you propose and that should definitely be doable

> iii) We should explore whether we can make our dependence on vscode live preview explicit to vscode, so that it automatically installs it along with OCaml Platform

I guess what you're referring to is [extension packs](https://code.visualstudio.com/api/references/extension-manifest#extension-packs). I'll check if it can be used to have `vscode-livepreview` automatically installed with the extension.

> iv) If going from a documentation page for a single package to the all packages documentations list is impossible, it may make sense to have two commands:
> 
> "Show documentation for packages in the current sandbox" - this shows the list of packages page by odoc, from which a user go to a particular package, go back to the list, go to another package.
> "Show documentation for a specific package in the current sandbox" - on trigger, shows a list of packages available in the sandbox as a vscode menu (quick picks - akin to the sandboxes list to pick from), a user picks a package to generate the docs for; ideally, we'd support generating docs for several packages, but I'm not sure vscode live preview can do that -- one package doc at a time is also fine for now I guess

Going from a documentation page for a single package to the all packages documentation list is possible, but you'll only see the packages you've already generated the doc for. I think what you propose makes sense. The first option "Show documentation for packages in the current sandbox" would generate the doc for all the packages in the current sandbox. It would allow users to navigate between packages' doc.

As for the second option, odig/odoc can generate documentation for multiple packages `odig odoc pkgName1 pkgName2 ...`. I'm just not sure if UI a vscode menu would allow selecting multiple packages (I'll check what's feasible).

> What does "search feature" mean here?

I was referring to the same feature you generally have in the browser when you type `CMD+F`:


https://user-images.githubusercontent.com/5595092/141746641-9edb77f8-9fcd-46d2-b166-55709cb5ee6d.mov

> As for depending on another extension, the alternative is to write our own server and depend on even more things, AFAIU. I'd imagine that to be more error-prone than using an extension doing that for us from the vscode team, given extensions have public APIs that I'd imagine to not be broken too often.

I'm not sure I understand what you mean here by "building our own server". The example I showed in my first post use a library to launch a server that serves the documentation files under a given folder. It just required a few lines of code (although we would probably have to write some proper bindings if we want to go with that approach):

```ocaml
let _ =
  Js_of_ocaml.Js.Unsafe.eval_string
    "
    const polka = require('polka');
    // path should not be harcoded ofc
    const serve = require('sirv')('/Users/corentin/.opam/default/var/cache/odig/html/')
    polka()
      .use(serve)
      .listen(5000, () => {
        console.log(`> Running on localhost:5000`);
      });"
```

I'm not pushing for one or the other solution, they both have their pros and cons. I just would like to make sure that you understand both of them.

> I'm confused as to why we'd need all of that though? What's blocking us from serving the HTML page from the filesystem with the built-in browser?

You can't give the path to a file to the built-in `Simple Browser`. You have to provide an URL (AFAIK)

tatchi:
Btw and now that I think about it, this PR has 2 "features":

* Generating the documentation so that it can be viewed in an offline mode (doesn't require an internet connection)
* Viewing the documentation right in the VSCode IDE

These 2 features are independent of each other. For instance, we could show the "online" documentation right in the IDE (through the `Simple Browser`) instead of opening it up in the user's external browser. Same for the offline documentation, we could open it both right in the IDE or in the browser.

Would that make sense to have a configuration for the preferred way of showing the documentation (i.e either inside or outside the IDE)? Both the online/offline documentation would adapt to that config. Eventually with a way to change the preferred option for a one time use case.

ulugbekna:
When I invoke `OCaml: Generate documentation`, nothing happens and I have this in the developer tools:

image

I guess it has to do with the fact that my sandbox is a global OPAM switch. It says some promise is not handled

---

> As for the second option, odig/odoc can generate documentation for multiple packages odig odoc pkgName1 pkgName2 .... I'm just not sure if UI a vscode menu would allow selecting multiple packages (I'll check what's feasible).

I meant more like having several tabs with documentation for different packages as you would be able to have in a browser. If we can only look up a single package documentation at a time -- that's quite limited compared to the experience a user can get using browser.

src/odig.ml
```diff
@@ -0,0 +1,62 @@
...
+    in
+    List.map lines ~f:(String.rsplit2 ~on:':')
+    |> List.filter_opt
+    |> List.find_map ~f:(fun (k, v) ->
+           if String.equal k key then
+             Some
+               (String.strip v ~drop:is_whitespace_or_single_quote
+               |> Path.of_string)
+           else
+             None)
```
ulugbekna, line 37: We may not need `conf` fn at all since we have `odig cache path`, but if we do:
we don't need to rsplit2 each line to find a particular key-value pair, I'd split on-demand and minimize unnecessary computations. Or if we want to query `odig conf` several times, I'd create a map from k-v pairs and return it

ulugbekna (changes requested):
I went through code quickly left some comments -- not a big deal since we're still in experimenting phase with these features

ulugbekna:
As long as we keep the amount of code minimal, let's try an implementation with polka if that's going to make working with documentation in global switches easier

src/odig.ml
```diff
@@ -0,0 +1,62 @@
...
+
+let cache_dir t =
+  let opam, switch = t in
+  let default_cache_dir = Path.(Opam.path opam switch / "/var/cache/odig/") in
+  let open Promise.Syntax in
+  let+ cache_dir = conf t "cache-dir" in
```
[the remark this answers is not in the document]
tatchi: Good point. As long as we only need the `cache_dir` config it's probably better since it simplifies the code. https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/c4620028547d15a004867a25f138af189e676357

src/opam.ml
```diff
@@ -305,3 +310,13 @@ let root_packages t switch =
...
+
+let has_package t switch package_name =
+  let open Promise.Syntax in
+  let+ packages = packages t switch in
+  match packages with
+  | Ok packages ->
+    packages
+    |> List.exists ~f:(fun package ->
+           String.equal (Package.name package) package_name)
+  | Error _ -> false
```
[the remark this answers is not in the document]
tatchi: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/20aa8f22bd3a5adf474c88f984b6745e12498871

src/treeview_sandbox.ml
```diff
@@ -79,6 +79,133 @@ module Command = struct
...
+          | Sandbox.Esy (_, _)
+          | Sandbox.Global
+          | Sandbox.Custom _ ->
+            let message =
+              Printf.sprintf
+                "This functionality is not supported for %s sandbox"
+                (Sandbox.to_string sandbox)
+            in
+            let+ _ = Window.showErrorMessage ~message () in
+            ()
```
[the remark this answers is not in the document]
tatchi: Ah yes, you're right. That definitely makes sense, Thanks! See: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/7811d9607cdc30fdb914d7bb8fc9cbb3ec5c37f7

src/odig.ml
tatchi: I made the changes to use `odig cache path` (see above) so that code doesn't exist anymore

tatchi:
> When I invoke `OCaml: Generate documentation`, nothing happens and I have this in the developer tools:
> 
> image
> 
> I guess it has to do with the fact that my sandbox is a global OPAM switch. It says some promise is not handled

Ah yes, I forgot to remove it from the `commandPalette`. It should be fixed in https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/6815484763c3bf5a1ba02d3cfb59d8b4915e3233

We can have some commands related to that feature like you suggested above but I'll prefer to implement that in a separate PR if you don't mind?


> Hmm, when I click on an icon to get the offline documentation in the Sandbox tree view, it works like a charm (I get a warning from livepreview saying the shown documentation is not at the server root, but it's minor -- it would of course be nice to not have it but the end result of having fully navigable docs is amazing)

Did it work the first time? For me the first time I generate the documentation for a package in a global switch, the live preview opens but it's kind of broken (probably because the extension restarted as I described in my first post). After that if I regenerate the doc it works fine.

> I meant more like having several tabs with documentation for different packages as you would be able to have in a browser. If we can only look up a single package documentation at a time -- that's quite limited compared to the experience a user can get using browser.

I don't think the `live preview` extension supports that, unfortunately.

tatchi:
Thanks so much for your reviews 🤗 I should have addressed all the requested changes.

I guess the next step is to give the following a try:

> As long as we keep the amount of code minimal, let's try an implementation with polka if that's going to make working with documentation in global switches easier

I don't expect it in itself to be hard but writing the bindings definitely will 😅 I've no experience in writing bindings with `js_of_ocaml` and AFAIK there's not much documentation. I'll let you know if I face any issues.

In the meantime, please let me know if you still spot anything that should be improved in the code or if I missed something that needs my attention.

src/treeview_sandbox.ml
```diff
@@ -79,6 +79,134 @@ module Command = struct
...
+          | Sandbox.Opam (opam, switch) -> (
+            let* odig = Odig.of_opam (opam, switch) in
+            match odig with
+            | Error e ->
+              let message =
+                match e with
+                | Odig.Odig_not_installed ->
+                  "\"OCaml: Generate Documentation\": the package \"odig\" \
+                   must be installed to generate documentation."
+              in
```
tatchi, line 109: Should we propose the user to install the package ?

src-bindings/vscode/vscode.mli
```diff
@@ -1086,6 +1086,16 @@ module ExtensionTerminalOptions : sig
...
 end
 
+module Extension : sig
+  include Js.T
+end
+
+module Extensions : sig
+  val getExtension : string -> Extension.t or_undefined
+
+  val is_extension_installed : string -> bool
```
rgrinberg, line 1096: I don't think this helper is useful enough to exist in the binding.

src/odig.ml
```diff
@@ -0,0 +1,59 @@
...
+  else
+    Error Odig_not_installed
+
+let cmd_ouput (opam, switch) args =
+  let cmd = Opam.exec opam switch ~args in
+  Cmd.output cmd
+
+let cache_dir t =
+  let opam, switch = t in
+  let default_cache_dir = Path.(Opam.path opam switch / "/var/cache/odig/") in
```
rgrinberg, line 21: Where did you get the `/var/cache/odig` path from?

src/odig.mli
```diff
@@ -0,0 +1,11 @@
+type t
+
+type of_opam_error = Odig_not_installed
+
+val of_opam : Opam.t * Opam.Switch.t -> (t, of_opam_error) result Promise.t
+
+val odoc_exec : t -> string -> (string, string) result Promise.t
```
rgrinberg, line 7: label the string argument please.

src/odig.mli
```diff
@@ -0,0 +1,11 @@
+type t
+
+type of_opam_error = Odig_not_installed
```
rgrinberg, line 3: I don't think we need this error type. We don't treat in any special way and just convert it to a message. Might as well do it as early as possible.

src/odig.mli
```diff
@@ -0,0 +1,11 @@
+type t
```
rgrinberg, line 1: Could you add a small summary for the purpose of this module at the top.

src/opam.ml
```diff
@@ -275,6 +275,11 @@ let upgrade t switch = spawn t [ "upgrade"; switch_arg switch; "-y" ]
 let remove t switch packages =
   spawn t ("remove" :: switch_arg switch :: "-y" :: packages)
 
+let path (opam : t) (switch : Switch.t) =
+  match switch with
+  | Local p -> Path.(p / "_opam")
+  | Named n -> Path.(opam.root / n)
```
rgrinberg, line 281: That doesn't seem right to me. Shouldn't we be running `opam var prefix` or something?

src/treeview_sandbox.ml
```diff
@@ -79,6 +79,134 @@ module Command = struct
...
+                              (Interop.Js.String))
+                    ~options ~task
+                in
+                match result with
+                | Error _ ->
+                  let (_ : 'a option Promise.t) =
+                    Window.showErrorMessage
+                      ~message:
+                        (Printf.sprintf
+                           "Error while generating documentation for %s" name)
```
rgrinberg, line 161: Can this error occur in practice? If so, how can the user find out more about the cause?

src/treeview_sandbox.ml
```diff
@@ -79,6 +79,134 @@ module Command = struct
...
+                    | Sandbox.Opam (_, Opam.Switch.Local _)
+                    | Sandbox.Esy (_, _)
+                    | Sandbox.Global
+                    | Sandbox.Custom _ ->
+                      ()
+                  in
+                  let htmlDocumentationPath = Path.(html_dir / name) in
+                  let (_ : Ojs.t option Promise.t) =
+                    Vscode.Commands.executeCommand
+                      ~command:"livePreview.start.preview.atFile"
```
rgrinberg, line 194: This is a builtin command, right? Then perhaps a typed binding for it should exist.

src/treeview_sandbox.ml
```diff
@@ -79,6 +79,134 @@ module Command = struct
...
+      let () =
+        let open Promise.Syntax in
+        let sandbox = Extension_instance.sandbox instance in
+        let (_ : unit Promise.t) =
+          match sandbox with
+          | Sandbox.Esy (_, _)
+          | Sandbox.Global
+          | Sandbox.Custom _ ->
+            let message =
+              "\"OCaml: Generate Documentation\" command only works with OPAM \
```
rgrinberg, line 93: `Opam`.

src/opam.mli
```diff
@@ -101,3 +103,6 @@ val exec : t -> Switch.t -> args:string list -> Cmd.t
 
 (** Check that two instances of [Opam] are equal. *)
 val equal : t -> t -> bool
+
+(** Check that a package is installed in the given switch. *)
+val has_package : t -> Switch.t -> string -> bool Promise.t
```
rgrinberg, line 108: naked string arugments should be labelled. We might also have a type for package names.

src/odig.ml
```diff
@@ -0,0 +1,59 @@
+open Import
+
+type t = Opam.t * Opam.Switch.t
+
+type of_opam_error = Odig_not_installed
+
+let of_opam (opam, switch) =
+  let open Promise.Syntax in
+  let+ is_installed = Opam.has_package opam switch "odig" in
```
rgrinberg, line 9: Why do we care if the `odig` package is installed? Isn't it sufficient for an `odig` binary to be present?

src/odig.ml
```diff
@@ -0,0 +1,62 @@
...
+let cache_dir t =
+  let opam, switch = t in
+  let default_cache_dir = Path.(Opam.path opam switch / "/var/cache/odig/") in
+  let open Promise.Syntax in
+  let+ cache_dir = conf t "cache-dir" in
+  Option.value cache_dir ~default:default_cache_dir
+
+let html_dir t =
+  let open Promise.Syntax in
+  let+ cache_dir = cache_dir t in
```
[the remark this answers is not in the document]
rgrinberg: I like getting all the config upfront. If only for the reason that it guarantees that we use a consistent set of values. However, to make it work we need to have a reasonable way to refresh this value and it has to be consistent with how we handle other binaries.

Option 2 is definitely a no-no. We don't want to push the complexity on users.

src/treeview_sandbox.ml
rgrinberg: After reading the other comments in this thread, I understand that this is not builtin. And neither is this thing very reliable. So never mind.

src/treeview_sandbox.ml
ulugbekna: Yes, I believe in another PR with proper infrastructure for package installation and proposing to install certain packages. I'd put a to-do to document that we should come back with package installation proposal.

ulugbekna:
I'll test this a little later, but for now: 

> For now the port is hardcoded to 4200 and if that one is already being used it won't work. We can rely on an external library to find a free port (might be easier to create the binding for), or just write that code ourselves. We could take inspiration from how the live preview extension is doing it.

The way live preview extension does this seems very reasonable to me. I'd do the same. 

> Currently the server is closed only when vscode itself is being closed (extension is disposed). The live preview extension closes the server once the user closes the webview. The problem here is that we don't control the webview, so there's no way to listen to the onClose (or whatever else) event. We could add some logic to inject a script into the HTML documentation pages that would ping our server at regular intervals. That way, we could know when the user closes the webview and stop the server. Does that sound reasonable/worth it?

If the resources taken by the server are significant enough, we could have a sort of debouncing? we could schedule server closure after `n` minutes, every time the server sends a response and thus proving it's needed, it could reschedule its closure again to be in `n` minutes (this is under assumption that on server closure the currently active webview is still valid)

src/treeview_sandbox.ml
tatchi: Isn't that already doable with [Opam.install](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/src/opam.mli#L49) ?

src/odig.ml
tatchi: From `odig conf --help`:

```
COMMON OPTIONS
       --cache-dir=PATH (absent ODIG_CACHE_DIR env)
           Cache directory. If unspecified, $PREFIX/var/cache/odig with
           $PREFIX the parent directory of odig's install directory.
```

My code might not be correct though if `odig` is not installed in the current opam switch 🤔 Should it be preferable to fail case we can't get the `cache_dir` from `odig`?

src/odig.mli
tatchi: Definitely better, thanks :)

ulugbekna:
When I open docs (using icon in the treeview) for one package, close the tab, and then try opening the docs for another package, the extension host terminates, which is very bad. This happens because of "port is already in use error", which means we're not reusing the same server (?)

image

tatchi:
> When I open docs (using icon in the treeview) for one package, close the tab, and then try opening the docs for another package, the extension host terminates, which is very bad. This happens because of "port is already in use error", which means we're not reusing the same server (?)

Sorry, I didn't notice the error (I probably didn't test with the console opened) 😅. There's indeed that error about the port already being used although everything still seems to work fine for me. The extension host doesn't terminate and the doc for the package shows up (probably why I didn't notice it).

Anyway, it should be fixed in https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/20260e5e19619f8264980efb18287e6f2abb0033. Thanks for testing 😁

src/opam.ml
tatchi: That function was copied over from the `Opam.Switch` module. I was struggling to expose it in the `.mli` file so I decided to copy it. I've fixed that now by moving the `type t` at the top of the file and by using a "type substitution" `type opam := t`. See https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/76b52651dc805ecc2a70ca006d6d09badc5475be

We can revert that change and not expose the `path` function if we decide not to provide a `default_cache_dir`. See my comment here: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#discussion_r760440462

src/treeview_sandbox.ml
tatchi: Yes, that can happen for instance with packages that don't have any documentation to generate. An example is `ocaml-lsp-server` which shows up in the list of packages installed in the current switch.

When you try to generate the documentation for that package `odig odoc ocaml-lsp-server`, the command actually succeed but no documentation is being written to disk:

```bash
~/Dev/vscode-ocaml-platform offline-doc ≡ 1.55s
❯ odig odoc ocaml-lsp-server
Updating documentation, this may take some time...

~/Dev/vscode-ocaml-platform offline-doc ≡ 1.16s
❯ echo $status
0
```

Hence why we [verify that the expected documentation folder for the package was created and return an error if not](https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/files#diff-dee540601fb09da08934f2e16441a5ef20ef7a1eaf925e55a200b9f369a02aaaR45).

The interesting thing is that while the `odoc` command doesn't produce any error, the `doc` command behave differently:

```bash
❯ odig doc ocaml-lsp-server
Updating documentation, this may take some time...
odig: [ERROR] No doc could be generated for:
              '/Users/corentin/Dev/vscode-ocaml-platform/_opam/var/cache/odig/html/ocaml-lsp-server/index.html'

~/Dev/vscode-ocaml-platform offline-doc ≡ 0.81s
❯ echo $status
3
```

I thought that was an issue on `odig` side but that's the expected behaviour. See relevant issue [here](https://github.com/b0-system/odig/issues/64)

src/odig.ml
tatchi: Definitely makes sense indeed. I'll make the changes, thanks 😊

rgrinberg:
> The vscode live preview internal server can't display files that are outside of the currently opened workspace.

What about symlinking the docs dir to the current workspace?

tatchi:
> > The vscode live preview internal server can't display files that are outside of the currently opened workspace.
> 
> What about symlinking the docs dir to the current workspace?

That should work indeed. We don't know when the user closes the webview though. So if we do that, I don't know when should we unlink the folder. We could eventually do it when closing the project ?

Btw, that's not an issue anymore with the latest changes that use our own server with the built-in `Simple Browser` extension. Would you still prefer to go back to using the "external" `Live preview` extension and try to symlink the doc folder?

tatchi:
I added the code to [find a free](https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/files#diff-c4345405ca0b702136608a49ddf57af1dbf74a039220552da8678e95d709591fR39) port and addressed some of @rgrinberg's comments. Please don't review the code just yet as I still would like to make some minor changes and adda few comments.

I would appreciate if you could test the PR and give feedback so we can decide on what option to pick. To recap, we can either:

* Go with the external "Live preview" extension. It would requires less code but has some issues:
  - Extension can freeze in some situations. Should hopefully be resolved soon: https://github.com/microsoft/vscode-livepreview/issues/174
  - Doesn't work well with global switches as the extension can't serve files outside of the workspace. We could `symlink` the documentation folder to make it work (question: when should we unlink it?)
  - It requires an external extension to be installed. We can try to create an [extension pack](https://code.visualstudio.com/api/references/extension-manifest#extension-packs) so that it get installed together with our extension?
  
* Go with our own server and the built-in "Simple Browser" extension. It requires more code from our side but didn't find any issues (for now). The only drawback is that, unlike the "Live Preview" extension, the built-in one doesn't have a [find feature](https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#issuecomment-968656066).

The current code implements the second option.

rgrinberg:
> Btw, that's not an issue anymore with the latest changes that use our own server with the built-in Simple Browser extension. Would you still prefer to go back to using the "external" Live preview extension and try to symlink the doc folder?

No, both approaches are fine. Keep whatever works.

src/treeview_sandbox.ml
ulugbekna: Sorry for a late reply. The code for opam installing is there, but I was thinking of a uniform and reusable piece of code that could be used to propose installing various dependencies that we rely on such as ocaml-lsp-server. It would have good error reporting and propose a nice way to propose installing a particular package depending on the current sandbox (opam or esy). Does this make sense? We can add opam install dialog here as well, I was just aimingfor a good piece of code reusable in various places

src-bindings/polka/polka.ml
```diff
@@ -0,0 +1,69 @@
...
+    val address : t -> Address.t [@@js.call]
+
+    val on : t -> string -> Ojs.t -> unit [@@js.call]]
+
+  let on t = function
+    | `Close f -> on t "close" @@ [%js.of: unit -> unit] f
+    | `Error f -> on t "error" @@ [%js.of: err:Node.JsError.t -> unit] f
+end
+
+type polka = Ojs.t
```
tatchi, line 25: Any idea how I could use `type t` instead? The reason why I had to change it to `polka` is to not conflict with the `Middleware.t`.

On the `mli` file I could do:

```ocaml
module Middleware : sig
  type polka := t
  type t
end
```

But here I don't know how to do that 😅

src/odig.ml
tatchi: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/457797fef7b602debc00c3948fae777d6ad01b93

src/opam.ml
tatchi: Removed that code

src/odig.ml
tatchi: I removed the `default_cache_dir`

src/treeview_sandbox.ml
tatchi: Ok I see thanks. I added a comment with a link to your reply 👆 https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/3a089980be0f485258b13b2604c7730e79dd02d3

src/odig.ml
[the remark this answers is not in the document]
tatchi: I went for option 3 in https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/0ca682d129a3b630703565fa8bd7c3b932c975b0. There's no way to refresh the `cache_dir` value but since a new `Odig.t` value is created every time we generate documentation,  it's not likely that it'll get stale.

tatchi:
@rgrinberg server is bound to port `0` now as suggested: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/091c5e60dd5bd035d50e31255953c9f22a8ba332 (can't find your comment anymore, probably because you did it on a commit and I've force pushed).


> If the resources taken by the server are significant enough, we could have a sort of debouncing? we could schedule server closure after n minutes, every time the server sends a response and thus proving it's needed, it could reschedule its closure again to be in n minutes (this is under assumption that on server closure the currently active webview is still valid)

I don't expect the server to take significant resources (I might be wrong ofc). Still, I added some code to close the server after 10 minutes without any requests received: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/abfe8cb66f4b865f6682d425547664bcfa1e3726

While I agree that it's probably an easy and reasonable solution for now, it's definitely not ideal. What if someone keeps the documentation's webview opens for some time and at some point later on clicks on a link? It might be that the server got closed by that time and the user will see a blank screen.

tatchi:
Alright, I should have addressed the remaining comments (just have one question: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#discussion_r770906535). Should be ready to get reviewed 😁

ulugbekna:
I’m also in the process of reviewing it. So far I have come up with the following feedback below. Not all of the points can technically be fixed however
- Bugs:
    - Minor: the tab name is `Simple Browser` . Can this be changed? — afaiu, no
    - Server can have a timeout-based termination, but it shouldn’t terminate if there’s a `Simple Browser` tab open. Only if the browser is closed, can the server terminate.
    - When a user has a tab with documentation open and try opening doc for another package, they get a mis-rendered tab
    - Clicking on links that lead to a website results in an empty doc tab (clicking on open in a new tab works - mouse wheel click for me)
- Possible improvements:
    - Add a setting `Generated documentation theme` with options in `odig odoc-theme list`
    - Make command more accessible:
        - Put it in the command palette
        - Put it in the Commands view

src-bindings/polka/polka.ml
ulugbekna: hmm, `type nonrec t = ...` should work? I'd say type `polka` is more readable when used in `Middleware.t`'s definition

src/documentation_server.ml
```diff
@@ -0,0 +1,52 @@
...
+
+  let serve =
+    polka
+    |> Polka.use
+         [ timeout_middleware; Polka.Sirv.serve (path |> Path.to_string) ]
+    |> Polka.listen port ~callback:(fun () ->
+           let server = Polka.server polka in
+           let address = Polka.Server.address server in
+           let port = Polka.Server.Address.port address in
+           resolve (Ok { server; port }))
```
ulugbekna, line 39: why do we keep the port number in type `t` if it's accessible from `server`?

src/treeview_sandbox.ml
```diff
@@ -79,6 +79,88 @@ module Command = struct
...
+              (module Interop.Js.Result (Interop.Js.String) (Interop.Js.String))
+              ~options ~task
+          in
+          match result with
+          | Error _ ->
+            let (_ : 'a option Promise.t) =
+              Window.showErrorMessage
+                ~message:
+                  (Printf.sprintf "Error while generating documentation for %s"
+                     package_name)
```
ulugbekna, line 116: There should be a util fn in `import.ml`: `show_message`, which can be useful to replace this piece of code

src-bindings/polka/polka.ml
tatchi: Let's stick with `polka` then :P

rgrinberg:
@mnxn do you have time to give this a review?

mnxn:
I won't have time to look at the code until Monday. I can review it then.

It's too early, but will we be integrating this with the [v3.ocaml.org package documentation](https://v3.ocaml.org/packages) in the future?
odig is still great for local/unpublished files, but the online documentation could be a smoother and faster experience. Also the last time I tried, odig had some problems on Windows (I'm not sure if that has changed).

ulugbekna:
@mnxn the nicety of this PR is that the docs are shown in vscode itself & have or will have hopefully lower latency

src/odig.ml
```diff
@@ -0,0 +1,58 @@
...
+
+type t =
+  { make_cmd : string list -> Cmd.t
+  ; cache_dir : Path.t
+  }
+
+(** TODO: propose to install odig. See
+    https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#discussion_r765297112 *)
+let of_sandbox (sandbox : Sandbox.t) =
+  let cmd = Sandbox.get_command sandbox "which" [ "odig" ] in
```
rgrinberg, line 11: We have `Cmd.which` exactly so that we don't rely on this binary existing on the user's system.

rgrinberg (approved):

src/odig.ml
```diff
@@ -0,0 +1,58 @@
...
+      let html_dir = html_dir t in
+      let package_html_dir = Path.to_string html_dir ^ package_name in
+      let open Promise.Syntax in
+      let+ dir_exists = Fs.exists package_html_dir in
+      if dir_exists then
+        ok
+      else
+        Error
+          (Printf.sprintf
+             "Directory %s should have been created but it doesn\'t exist."
```
rgrinberg, line 48: How is it possible to reach this error case?

src/odig.mli
```diff
@@ -0,0 +1,11 @@
+(** Provide an interface to Odig. *)
+
+type t
+
+val of_sandbox : Sandbox.t -> (t, string) result Promise.t
+
+(** Generates the odoc API documentation and manual of [package_name]. Interface
+    to the [odig odoc package_name] command *)
```
rgrinberg, line 8: The return value of this function isn't clear. Could you describe it in the documentation string?

src/odig.ml
```diff
@@ -0,0 +1,58 @@
...
+             package_html_dir)
+    | Error _ as e -> Promise.resolve e
+  in
+  result
+  |> Result.map_error ~f:(fun error ->
```
rgrinberg, line 53: `Result.map_error result` is more natural and one line shorter.

src/odig.ml
```diff
@@ -0,0 +1,58 @@
...
+      let open Promise.Syntax in
+      let+ dir_exists = Fs.exists package_html_dir in
+      if dir_exists then
+        ok
+      else
+        Error
+          (Printf.sprintf
+             "Directory %s should have been created but it doesn\'t exist."
+             package_html_dir)
+    | Error _ as e -> Promise.resolve e
```
rgrinberg, line 50: Since this clause is shorter, could you move it to the top?

src/odig.ml
```diff
@@ -0,0 +1,58 @@
...
+let html_dir t = Path.(t.cache_dir / "/html/")
+
+let odoc_exec t ~package_name =
+  let open Promise.Syntax in
+  let* ouput = cmd_output t ~args:[ "odoc"; package_name ] in
+  let+ result =
+    match ouput with
+    | Ok _ as ok ->
+      let html_dir = html_dir t in
+      let package_html_dir = Path.to_string html_dir ^ package_name in
```
rgrinberg, line 40: why don't you use path concatenation here?

src/odig.ml
```diff
@@ -0,0 +1,58 @@
+open Import
+
+type t =
+  { make_cmd : string list -> Cmd.t
```
rgrinberg, line 4: Why don't we just store the sandbox value instead of the function?

src/documentation_server.ml
```diff
@@ -0,0 +1,51 @@
...
+
+(** [debouncing_middleware ~timeout_in_ms f] on every request schedules [f] to
+    be run in [timeout_in_ms] milliseconds; on every new request the last
+    scheduled timeout is overridden *)
+let debouncing_middleware ~timeout_in_ms f =
+  let timeout = ref None in
+  fun ~request:_ ~response:_ ~next ->
+    (* on every request, we clear previous timeout and set a new one *)
+    Option.iter !timeout ~f:Node.clearTimeout;
+    timeout := Some (Node.setTimeout f timeout_in_ms);
```
rgrinberg, line 13: Since this function relies on staging, could you stage it explicitly with `Stage.t`?

rgrinberg (changes requested):
I noticed that the documentation server stores a reference to the sandbox (rather a function derived from the sandbox, but no matter). What happens when the sandbox is switched while the documentation is running and we trigger odoc?

src-bindings/polka/polka.ml
```diff
@@ -0,0 +1,69 @@
...
+
+let polka_of_js = Ojs.t_of_js
+
+let polka_to_js = Ojs.t_to_js
```
mnxn, line 29: These 3 lines can be replaced with just

```suggestion
type polka = Ojs.t [@@js]
```

src/odig.ml
```diff
@@ -0,0 +1,58 @@
...
+    match cache_dir with
+    | Ok cache_dir ->
+      let cache_dir = cache_dir |> String.strip |> Path.of_string in
+      Ok { make_cmd; cache_dir }
+    | Error _ -> Error "OCaml: Failed to retrieve odig cache_dir")
+  | Error _ ->
+    Promise.resolve
+      (Error
+         "\"OCaml: the \"odig\" binary must be available in the current \
+          sandbox to generate documentation.")
```
mnxn, line 27: The starting double-quote is unmatched. I don't think its necessary here.

mnxn:
I can confirm that odig works on Windows. However there is an issue that I can't seem to find to cause of:

Opening the documentation for some packages (like `gen_js_api` or `utop`) opens the browser tab but the page is blank and just says `Not found`. A 404 error is in the developer console. `odig doc gen_js_api` opens the HTML page in my default browser, so the documentation is definitely being generated.  

Other packages successfully open the documentation and I don't know what the difference is that makes these fail.
On macOS, I don't have this problem, maybe it has something to do with the server since the local files work fine.

<br>

Another thing I noticed is that the generate documentation button is shown for package which don't have documentation.
```bash
 $ odig doc ocamlformat
Updating documentation, this may take some time...
odig: [ERROR] No doc could be generated for:
              '/Users/max/.opam/default/var/cache/odig/html/ocamlformat/index.html'
```
Clicking on the documentation button for packages like these always causes an error
> Failed to generate documentation for package ocamlformat. Error: Directory /Users/max/.opam/default/var/cache/odig/html/ocamlformat should have been created but it doesn't exist.

Is there any way to hide the button if documentation can't be generated for a package?

tatchi:
Thanks @rgrinberg  and @mnxn for the valuable feedback 🙏  For whatever reason, I can't reply to the review comments so please find my answers below.

> How is it possible to reach this error case? https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#discussion_r806273155

See https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#discussion_r753828800

> Why don't we just store the sandbox value instead of the function? https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#discussion_r806274969

No specific reason. I changed it to store the `sandbox` value. https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/d7b91391cac74ad03f7a8e1c7014c8962e62c94a

> I noticed that the documentation server stores a reference to the sandbox (rather a function derived from the sandbox, but no matter). What happens when the sandbox is switched while the documentation is running and we trigger odoc?

We see a 404 not found page 😅  The reason is that the first time we generate the documentation for a package, we start the server and configure it to serve the current sandbox's documentation folder (i.e `_opam/var/cache/odig/html`). After that, if we want to see the documentation for another package, we reuse the current server and only change the `url` of the `simple browser extension` to point to the newly generated package's documentation.
The problem is that if in the meantime we've switched over to another sandbox, the server still serve the previous sandbox's documentation folder (hence why we see a 404 not found).
We could stop the documentation server when [we set the sandbox](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/src/extension_instance.ml#L165) or restart the documentation server each time we generate doc for a package. WDYT?

> Opening the documentation for some packages (like gen_js_api or utop) opens the browser tab but the page is blank and just says Not found. A 404 error is in the developer console. odig doc gen_js_api opens the HTML page in my default browser, so the documentation is definitely being generated.

That's because:

> The massive performance advantage over serve-static is explained by not relying on the file system for existence checks on every request. These are expensive interactions and must be avoided whenever possible! Instead, when not in "dev" mode, sirv performs all its file-system operations upfront and then relies on its cache for future operations.

I enabled the `dev` mode and it should work fine now :) 

> Another thing I noticed is that the generate documentation button is shown for package which don't have documentation.

>  $ odig doc ocamlformat
> Updating documentation, this may take some time...
> odig: [ERROR] No doc could be generated for:
>               '/Users/max/.opam/default/var/cache/odig/html/ocamlformat/index.html'
> Clicking on the documentation button for packages like these always causes an error
> 
> Failed to generate documentation for package ocamlformat. Error: Directory /Users/max/.opam/default/var/cache/odig/html/ocamlformat should have been created but it doesn't exist.
> 
> Is there any way to hide the button if documentation can't be generated for a package?

I quickly checked the `Odig` API and didn't find anything that could tell us if the documentation could be generated for a package. I could read the code and try to identify what's required to generate the documentation but not sure if that worth it?

mnxn:
> I enabled the dev mode and it should work fine now :)

Yep, I can confirm I no longer get `Not found` pages for those packages.

> I quickly checked the Odig API and didn't find anything that could tell us if the documentation could be generated for a package. I could read the code and try to identify what's required to generate the documentation but not sure if that worth it?

Ok, it's fine to leave it then. I think the VSCode message popup could be reworded a little though. I would prefer something like "Documentation could not be generated for" the package instead of the current message which emphasizes an error occurring.

If certain packages don't support generating documentation at all, leading with "error" suggests the behavior is unexpected and there is a way to fix it when there isn't (at least that's how I interpret the message).

rgrinberg:
> Ok, it's fine to leave it then. I think the VSCode message popup could be reworded a little though. I would prefer something like "Documentation could not be generated for" the package instead of the current message which emphasizes an error occurring.

Additionally, we should mention in the error messages that the package might not have any documentation. Later, we can try to detect this error case better.

rgrinberg:
> We could stop the documentation server when [we set the sandbox](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/src/extension_instance.ml#L165) or restart the documentation server each time we generate doc for a package. WDYT?

Restarting it when the sandbox is switched seems fine. It seems error prone to me to keep a reference to an out of date sandbox value.

src/documentation_server.ml
```diff
@@ -0,0 +1,52 @@
+open Import
+
+type t = Polka.Server.t
+
+(** [debouncing_middleware ~timeout_in_ms f] on every request schedules [f] to
+    be run in [timeout_in_ms] milliseconds; on every new request the last
+    scheduled timeout is overridden *)
+let debouncing_middleware ~timeout_in_ms f =
```
rgrinberg, line 8: Actually, I'm not sure why we need this debouncing middleware at all.

src/documentation_server.ml
```diff
@@ -16,6 +16,7 @@ let debouncing_middleware ~timeout_in_ms f =
 let start ~path ?(port = 0) () =
   Promise.make @@ fun ~resolve ~reject:_ ->
   let polka = Polka.create () in
+  (* Closes the server after some idle time *)
```
rgrinberg, line 19: What's the need to do this though? The fact that the server is turned off idling isn't at all obvious, and the resources saved by this don't see very impressive either. If it's so important to turn this off, let's just allow the user to do this manually with a status item element.

tatchi:
> Ok, it's fine to leave it then. I think the VSCode message popup could be reworded a little though. I would prefer something like "Documentation could not be generated for" the package instead of the current message which emphasizes an error occurring.

> Additionally, we should mention in the error messages that the package might not have any documentation. Later, we can try to detect this error case better.

Definitely makes sense. I changed the message in https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/57c3c13855e024ea55e840729ffd4225bafb44cb

> Restarting it when the sandbox is switched seems fine. 

https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/ba68598225d9ab7724d38e35156ab95472a87c1a

> It seems error prone to me to keep a reference to an out of date sandbox value.

A new instance of `Odig.t` is created every time we generate documentation for a package so in practice it shouldn't happen that we use an out of date `Sandbox.t`. Still, I agree with your point and removed the reference to the `Sandbox.t` in https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/337938fdd5b5910bc9984e532764ee3c8ca5fe05

tatchi:
I still can't reply to review comments, tried to close/reopen the PR but the problem persists 😩

> What's the need to do this though? The fact that the server is turned off idling isn't at all obvious, and the resources saved by this don't see very impressive either. If it's so important to turn this off, let's just allow the user to do this manually with a status item element.

That's what the live preview extension does so I thought that would be good to do something similar. See discussion [here](https://github.com/ocamllabs/vscode-ocaml-platform/pull/771#issuecomment-983875605)

I don't expect the server to take lots of resources though so that's probably fine to keep it alive. The status item with a way to manually close the server seems a good idea, that's also something offered by the live preview extension:

image

rgrinberg:
> I don't expect the server to take lots of resources though so that's probably fine to keep it alive. The status item with a way to manually close the server seems a good idea, that's also something offered by the live preview extension:

Yeah, that seems sufficient. It's good that we're trying to keep things more convenient for the user, but this seems like a small step past the limit of what we should be doing. If something requires a few extra clicks, it's a small price to pay for everything being as predictable a possible.

tatchi:
I removed the code that closes the server after some idle time and added a status bar item that shows the server port (when running). We can close the server by clicking on that status bar item as shown in the video below:

https://user-images.githubusercontent.com/5595092/155941229-7ded8923-f273-4432-a9cf-e5154af5e834.mov

It's heavily inspired by how the live preview implements it. I even used the same icon and text (open to any better suggestion 😄).

src/documentation_server.ml
```diff
@@ -0,0 +1,40 @@
+open Import
+
+type t = Polka.Server.t
+
+let start ~path ?(port = 0) () =
```
rgrinberg, line 5: Since we are never passing the `?port` parameter, I would just get rid of it and hardcode 0. No use in making things general here.

src/documentation_server.ml
```diff
@@ -0,0 +1,40 @@
...
+
+let port t =
+  match Polka.Server.address t with
+  | None ->
+    (* if it's [None], server must not be listening, but we aim to have server
+       only if it's listening *)
+    assert false
+  | Some a -> Polka.Server.Address.port a
+
+let on_close t ~f = Polka.Server.on t (`Close f)
```
rgrinberg, line 28: I don't ese where this function is used.

src/extension_instance.ml
```diff
@@ -153,18 +182,41 @@ end
...
+      let open Promise.Syntax in
+      let+ (_ : (unit, Promise.error) result) =
+        Documentation_server.stop server
+      in
+      t.documentation_server <- None;
+      let (_ : Ojs.t option Promise.t) =
+        Vscode.Commands.executeCommand ~command:"setContext"
+          ~args:[ Ojs.string_to_js document_server_on; Ojs.bool_to_js false ]
+      in
+      StatusBarItem.hide t.documentation_server_info)
```
rgrinberg, line 211: Why don't we dispose of the status bar item? If we don't, I'm not sure that reloading the plugin is going to be reliable.

src/treeview_sandbox.ml
```diff
@@ -78,6 +78,83 @@ module Command = struct
...
+            match port with
+            | Error e ->
+              show_message `Error "%s" e;
+              ()
+            | Ok port ->
+              let (_ : Ojs.t option Promise.t) =
+                Vscode.Commands.executeCommand ~command:"simpleBrowser.show"
+                  ~args:
+                    [ Ojs.string_to_js
+                        (Printf.sprintf "http://localhost:%i/%s/index.html" port
```
rgrinberg, line 147: It doesn't matter so much, but I think it's more correct to obtain the host from the server rather than hard code localhost. Is that possible?

tatchi:
Thanks for the review @rgrinberg. Since I still can't directly reply to your comments, let me write my reply down here:

> Does that mean this command is useless?

No, that command is still triggered when we click on the icon to generate the documentation for a package. That line of code hides the command from the `command palette`. It looks like by default commands are available in the command palette so we need to explicitly tell VScode to not show it.

> Why don't we dispose of the status bar item? If we don't, I'm not sure that reloading the plugin is going to be reliable.

It should be correctly disposed now: https://github.com/ocamllabs/vscode-ocaml-platform/pull/771/commits/817c53a2ac3eb731dbcab548fa49663344258bbb

> It doesn't matter so much, but I think it's more correct to obtain the host from the server rather than hard code localhost. Is that possible?

I wasn't able to find an easy way to obtain the host. I tried using [server.address().address](https://nodejs.org/api/net.html#serveraddress) but it doesn't work.

rgrinberg:
@tatchi i made a few more refactorings to bring things close to the style of the rest of the code base. Please take a look.

tatchi:
> @tatchi i made a few more refactorings to bring things close to the style of the rest of the code base. Please take a look.

Thanks @rgrinberg and sorry to give you extra work 😅  I dropped a few comments but nothing serious :) 

I also pushed 2 new commits (see my comments inline)

rgrinberg:
Merged. Thanks tatchi.

PR #774, opened 2021-11-22
ci: fix lint-fmt job

tatchi:
It currently fails to parse the `ocamlformat` version due to the comment added in https://github.com/ocamllabs/vscode-ocaml-platform/pull/746

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -161,6 +161,6 @@ jobs:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
           dune-cache: true
 
-      - run: opam depext ocamlformat=$(cat .ocamlformat | grep version | cut -d '=' -f 2 | cut -d ' ' -f 1) --install
+      - run: opam depext --install ocamlformat=$(awk -F = '/version=/{print $2}' .ocamlformat)
 
       - run: opam exec -- dune build @fmt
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -161,6 +161,6 @@ jobs:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
           dune-cache: true
 
-      - run: opam depext ocamlformat=$(cat .ocamlformat | grep version | cut -d '=' -f 2) --install
+      - run: opam depext ocamlformat=$(cat .ocamlformat | grep version | cut -d '=' -f 2 | cut -d ' ' -f 1) --install
```
smorimoto, line 164:
```suggestion
      - run: opam depext --install ocamlformat=$(awk -F = '/version/{print $2}' .ocamlformat)
```

smorimoto: How about simply moving the comment to a line above and replacing the step with this?

smorimoto: There is not much difference, but I'm avoiding UUOC these days.
https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat

tatchi: I'm not familiar with Unix commands in general so I fully trust you on that 😄 Thanks for the suggestion :)

smorimoto (approved):

smorimoto:
Oh no please do not merge this.

smorimoto:
Okay, I think this is now ready for merging.

tatchi:
Thanks for the review @smorimoto 🤗

PR #777, opened 2021-11-23
minify release bundle even more

tatchi:
The `--minify` option does [3 things](https://esbuild.github.io/api/#minify): `--minify-whitespace`, `--minify-syntax` and `--minify-identifiers`. When `esbuild` was introduced in https://github.com/ocamllabs/vscode-ocaml-platform/pull/591, the `--minify-identifiers` was not used because [it broke the extension](https://github.com/ocamllabs/vscode-ocaml-platform/pull/591#discussion_r619703197).

I tried to re-introduce that flag it seems to work fine now.

Before: ` Packaged: ocaml-platform.vsix (76 files, 1.24MB)`
After: `Packaged: ocaml-platform.vsix (76 files, 1.06MB)`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,7 @@ build-release:
 		--outdir=dist \
 		--platform=node \
 		--target=es6 \
-		--minify-whitespace \
-		--minify-syntax \
+		--minify \
 		--sourcemap \
 		--sources-content=false
 


smorimoto:
I think it was disabled to make debugging easier. In fact, unlike a website, doing it for 0.2 MB is not a huge payoff. CC: @mnxn

tatchi:
> I think it was disabled to make debugging easier. In fact, unlike a website, doing it for 0.2 MB is not a huge payoff. CC: @mnxn

Good point, that's probably not worth the 0.2MB saving.

mnxn:
> I think it was disabled to make debugging easier.

Yes, I believe it was useful for debugging some runtime errors in published versions of the extension.

I don't know if it's still necessary to keep the identifiers (or if the sourcemap is enough), but for only a 0.2 MB difference we should leave it alone.

PR #780, opened 2021-11-26
update latest ocaml-lsp version (for ocaml 4.13) to 1.9.1

ulugbekna:


tmattio:
Doesn't 1.9.1 support formatting of dune files from LSP? Maybe we should also remove the "manual" formatting we do at the moment as part of this PR?

ulugbekna:
I don’t remember. I’ll update  the PR after checking if necessary

rgrinberg:
It's not supported because it relies on dune rpc which isn't released yet.

ulugbekna:
> It's not supported because it relies on dune rpc which isn't released yet.

uh, indeed, merging then

PR #783, opened 2021-11-26
Clean up ast explorer code a little

ulugbekna:


mnxn (approved):

src/ast_editor_state.ml
```diff
@@ -1,53 +1,49 @@
...
-      ~f:(fun data -> String.equal (Uri.toString uri ()) data)
-      t.origin_to_pp_doc_map
-    |> Map.keys
-  in
-  match r with
-  | k :: _ -> Some k
-  | _ -> None
+  let uri_s = Uri.toString uri () in
+  Map.filter ~f:(fun data -> String.equal uri_s data) t.origin_to_pp_doc_map
+  |> Map.keys |> List.hd
```
rgrinberg, line 41: Why aren't we using `min_binding` or even `find_first` here?

src/ast_editor_state.mli
```diff
@@ -5,6 +5,11 @@ open Import
 
 type t
 
+type ast_mode =
+  [ `Original_ast
+  | `Preprocessed_ast
+  ]
```
rgrinberg, line 11: normal variant please

src/ast_editor_state.ml
ulugbekna: my first question as well -- because `Map` assumes that you find by a key (`val find : ('k, 'v, 'cmp) t -> 'k -> 'v option`), while here we search by a value

we could perhaps do as below to not filter the whole map but I'm not 100% sure it's much better

```ocaml
  t.origin_to_pp_doc_map |> Map.to_sequence
  |> Sequence.find ~f:(fun (_, data) -> String.equal uri_s data)
```

rgrinberg: I see. Then I suppose we should just iter or fold and raise once we find something. What we'd really like to avoid is just generating all these intermediate data structures

rgrinberg: You might want to double check if converting a map to a sequence is at all helpful. I think it also allocates the entire map in a list like data structure before it traverses with sequence

ulugbekna: Updated the implementation to short-circuit with a local exception. 

re Sequence
I thought the idea of a sequence is to lazily access elements

rgrinberg: The idea is there, but the implementation of to_seq in map leaves much to be desired:

```
    type 'a enumeration = End | More of key * 'a * 'a t * 'a enumeration

    let rec cons_enum m e =
      match m with
        Empty -> e
      | Node {l; v; d; r} -> cons_enum l (More(v, d, r, e))

    let rec seq_of_enum_ c () = match c with
      | End -> Seq.Nil
      | More (k,v,t,rest) -> Seq.Cons ((k,v), seq_of_enum_ (cons_enum t rest))

    let to_seq m =
      seq_of_enum_ (cons_enum m End)
```

As you can see, not much is gained by converting it to a Seq.

PR #784, opened 2021-11-27
Fix jump to hole on destruct/construct to jump only within new generated text

ulugbekna:
comes in pair with https://github.com/ocaml/ocaml-lsp/pull/559

Some thoughts:

1. I think this was one of the last things I wanted to do before cutting a release for the extension. 
1.1. I think some testing would be nice, but few people use the master version, so we can release and then release again if there're problems reported
1.2. Could someone shed some light on how to cut a release? 

2. It would be nice to have a ppx to generate `of_/to_json` code. Doing that manually is wrong in many ways. It would be nice to reuse yojson ppxes but then we need to switch to using yojson ?

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -191,10 +191,38 @@ end = struct
     TextEditor.revealRange text_editor ~range:hole
       ~revealType:TextEditorRevealType.InCenterIfOutsideViewport ()
 
+  let jump_to_hole jump (instance : Extension_instance.t) ~args =
+    (* this command is available (in the command palette) only when a file is
+       open *)
+    match Window.activeTextEditor () with
+    | None ->
+      Extension_consts.Command_errors.text_editor_must_be_active
+        "Jump to Previous/Next Typed Hole"
+        ~expl:"The command looks for holes in an open file."
+      |> show_message `Error "%s"
+    | Some text_editor -> (
+      let open Promise.Syntax in
+      match Extension_instance.lsp_client instance with
+      | None -> show_message `Warn "ocamllsp is not running."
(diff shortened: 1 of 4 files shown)


src/extension_commands.ml
```diff
@@ -153,109 +153,49 @@ module Holes_commands : sig
...
-      failwith
-        "Jump to Previous/Next Hole: incorrect params passed to the command"
+  let hole_not_found_msg = "No typed hole was found in this file"
+
+  (** Shows appropriate message for when the [ocaml-lsp] in use by the extension
+      doesn't support jumping to holes. *)
+  let ocaml_lsp_doesn't_support_holes instance ocaml_lsp =
+    let suggestion =
+      match
+        Ocaml_lsp.is_version_up_to_date ocaml_lsp
```
rgrinberg, line 163: Why are we going to version numbers to query for features? We're using capabilities elsewhere.

ulugbekna: We do use capabilities to track whether ocaml-lsp can handle the command (see code below). This piece of code is run when ocaml-lsp doesn't have that capability to determine whether we need to ask the user to upgrade their ocaml-lsp version or they are bounded by their ocaml version. 

```ocaml
| Some (_, ocaml_lsp)
        when not (Ocaml_lsp.can_handle_typed_holes ocaml_lsp) ->
        ocaml_lsp_doesn't_support_holes instance ocaml_lsp
```

rgrinberg (approved):
Looks fine. We might need to modify the client to signal that it has a cursor jumping capability though. We don't want the server to offer vscode commands to all editors.

PR #789, opened 2021-11-30
Unify and clean up how we send custom requests

ulugbekna:


src/custom_requests.mli
```diff
@@ -0,0 +1,22 @@
+open Import
```
rgrinberg, line 1: You can use shorter names at no loss to readability:

* `Custom_request` for the module
* `Custom_request.t` for the type
* `send` instead of `send_request`.

rgrinberg (approved):

PR #795, opened 2021-12-07
Use stable lsp package in test fixture for esy

smorimoto:


rgrinberg:
Why are we switching from yarn to npm? Somebody made a PR to switch to npm recently so this is going contrary to that.

smorimoto:
It was done in https://github.com/ocamllabs/vscode-ocaml-platform/pull/726 by @tmattio. But npm isn't really that good, and it's rather unstable. In actual practice, yarn is much better. Also, even if you prefer npm, you can still install consistent packages using npm because the latest release already supports parsing of the yarn lock file. So switching to npm wasn't exactly an improvement.

smorimoto:
This will not only make use of the stable version of lsp, but also restore the esy cache. Now that GitHub Actions has increased the maximum cache limit, everything is fine.

smorimoto:
I think it will be simpler to use workspaces in the future, but that's not the scope of this PR, so I would like to do it in a different PR.
https://classic.yarnpkg.com/lang/en/docs/workspaces

ulugbekna:
In current master, we use both yarn and npm in Makefile. I think it's nice that we're settling with one of the them

smorimoto:
Now this PR is ready to merge!

tatchi:
Shouldn't we delete the `package-lock.json` file now that we've switched over to `yarn`?

smorimoto:
Oops... Indeed yes.

Issue #505, opened 2021-01-21
Support view *.cmi, *.cmt, *.cmo, *.bc and *.pp.ml files


hackwaly:
VS Code supports define custom editor. It would be useful that we can use it to show dump info of these files.

Files `*.cmi`, `*.cmt`, `*.cmo`, `*.bc`... can dump by `ocamlobjdump`.
And files `*.pp.ml` produced by dune can also dump by `ocamlc -dsource`.

Here's demo custom editor:
![image](https://user-images.githubusercontent.com/540298/105319418-ced84480-5bff-11eb-9170-0f598d38b004.png)

rgrinberg:
> Files *.cmi, *.cmt, *.cmo, *.bc... can dump by ocamlobjdump.

Sounds like a good idea to me.

> And files *.pp.ml produced by dune can also dump by ocamlc -dsource.

Does this require `-I` and other flags as well? If so, this is more complex and will require integration from dune.

In any case, this is a useful feature, so if someone wants to work on it, a pr would be welcome.

tatchi:
Hey @hackwaly,

> Files *.cmi, *.cmt, *.cmo, *.bc... can dump by ocamlobjdump

Did you mean `ocamlobjinfo` by any chance?

tatchi:
I created a little POC. Is that what you had in mind?

https://user-images.githubusercontent.com/5595092/143494208-285fda4d-f44d-42bf-b2b2-8ca30c7f4755.mov

rgrinberg:
@tatchi looks solid to me.

mnxn:
In the ocaml source code there is both [objinfo.ml](https://github.com/ocaml/ocaml/blob/4.13/tools/objinfo.ml) and [dumpobj.ml](https://github.com/ocaml/ocaml/blob/4.13/tools/dumpobj.ml) in the tools directory. 

`objinfo.ml` seem to be built into `ocamlobjinfo` and included by default, but `dumpobj` is not.

It might be useful to use `dumpobj` because it shows the bytecode instructions (according to it's source code). `ocamlobjinfo` seems pretty limited in the information it shows.

mnxn:
I had to build `dumpobj` from source with `make dumpobj` in `tools`. 
Here's a sample of what it can output:

<details>
  <summary> Excerpt </summary>
  
```
$ ./dumpobj.opt ./dumpobj.cmo
## start of ocaml dump of "./dumpobj.cmo"
[...]

PR #798, opened 2021-12-13
support cm* files

tatchi:
This PR adds a new custom editor that shows info about `.cm(t, i, ti, o)` and `.bc` files. It uses the `ocamlobjinfo` behind the scene.

I didn't test it with `.bc` files because I don't know how to generate these files 😅 

Question: it looks like `ocamlobjinfo` also supports `.cmx`, `.cmxa`, `.cma`, `.cmxs`. Should we also add support for them?

Partially implements #505

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@
   Currently supported artifacts include `.cmi`, `.cmt(i)`, `.cmo`, `.cma`,
   `.cmx(a/s)`, and `.bc` files.
 
-  To learn more about these files: https://ocaml.org/manual/comp.html
+  To learn more about these files, see: https://ocaml.org/manual/comp.html
 
 - Warn if the extension sees not the latest OCaml-LSP version compatible with
   the OCaml distribution installed in the current sandbox.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,8 +4,13 @@
 
 - Fix the detection of opam's root directory when no switch is detected (#831)
 
-- Support for opening `.cmi`, `.cmt`, `cmti`, `.cmo`, `.cma`, `.cmx`, `.cmxa`,
-  `.cmxs` and `.bc` files. (#798)
+- Add support for opening compilation artifacts in human-readable form in the
+  editor (#798)
+
+  Currently supported artifacts include `.cmi`, `.cmt(i)`, `.cmo`, `.cma`,
+  `.cmx(a/s)`, and `.bc` files.
+
+  To learn more about these files: https://ocaml.org/manual/comp.html
 
 - Warn if the extension sees not the latest OCaml-LSP version compatible with
   the OCaml distribution installed in the current sandbox.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 - Fix the detection of opam's root directory when no switch is detected (#831)
 
+- Support for opening `.cmi`, `.cmt`, `cmti`, `.cmo`, `.cma`, `.cmx`, `.cmxa`,
+  `.cmxs` and `.bc` files. (#798)
+
 - Warn if the extension sees not the latest OCaml-LSP version compatible with
   the OCaml distribution installed in the current sandbox.
 
--- a/package.json
+++ b/package.json
@@ -965,6 +965,18 @@
             "filenamePattern": "*.cmo"
           },
           {
+            "filenamePattern": "*.cma"
+          },
+          {
+            "filenamePattern": "*.cmx"
+          },
+          {
+            "filenamePattern": "*.cmxa"
+          },
+          {
+            "filenamePattern": "*.cmxs"
+          },
+          {
             "filenamePattern": "*.bc"
           }
         ]
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "onCommand:ocaml.show-preprocessed-document",
     "onCommand:ocaml.open-pp-editor-and-ast-explorer",
     "onCustomEditor:ast-editor",
+    "onCustomEditor:cm-files-editor",
     "workspaceContains:**/dune-workspace",
     "workspaceContains:**/dune",
     "workspaceContains:**/dune-project",
@@ -944,6 +945,27 @@
           },
           {
             "filenamePattern": "*.mli"
+          }
+        ]
+      },
+      {
+        "viewType": "cm-files-editor",
+        "displayName": "Cm files preview",
+        "selector": [
+          {
+            "filenamePattern": "*.cmi"
+          },
+          {
+            "filenamePattern": "*.cmt"
+          },
+          {
+            "filenamePattern": "*.cmti"
+          },
+          {
+            "filenamePattern": "*.cmo"
+          },
+          {
+            "filenamePattern": "*.bc"
           }
         ]
       }
--- a/package.json
+++ b/package.json
@@ -950,7 +950,7 @@
       },
       {
         "viewType": "cm-files-editor",
-        "displayName": "Cm files preview",
+        "displayName": "OCaml Compilation Artifact Viewer",
         "selector": [
           {
             "filenamePattern": "*.cmi"
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2902,7 +2902,7 @@ end
 module CustomReadonlyEditorProvider = struct
   include Interface.Generic (Ojs) ()
 
-  module Make (T : module type of CustomDocument) = struct
+  module Make (T : CustomDocument.T) = struct
     type t = T.t generic [@@js]
 
     include
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2334,8 +2334,6 @@ module CustomDocument = struct
     val uri : t -> Uri.t
 
     val dispose : t -> unit
-
-    val create : uri:Uri.t -> dispose:(unit -> unit) -> t
   end
 
   include Interface.Make ()
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2328,6 +2328,20 @@ module TreeItemCollapsibleState = struct
 end
 
 module CustomDocument = struct
+  module type T = sig
+    type t
+
+    val t_of_js : Ojs.t -> t
+
+    val t_to_js : t -> Ojs.t
+
+    val uri : t -> Uri.t
+
+    val dispose : t -> unit
+
+    val create : uri:Uri.t -> dispose:(unit -> unit) -> t
+  end
+
   include Interface.Make ()
 
   include
@@ -2888,7 +2902,7 @@ end
 module CustomReadonlyEditorProvider = struct
   include Interface.Generic (Ojs) ()
 
-  module Make (T : Js.T) = struct
+  module Make (T : module type of CustomDocument) = struct
     type t = T.t generic [@@js]
 
     include
@@ -3152,7 +3166,7 @@ module Window = struct
     [%js.to: TreeView.t] (createTreeView ~viewId ~options)
 
   let registerCustomReadonlyEditorProvider (type a)
-      (module T : Js.T with type t = a) ~(viewType : string)
+      (module T : CustomDocument.T with type t = a) ~(viewType : string)
       ~(provider : a CustomReadonlyEditorProvider.t)
       ?(options : RegisterCustomEditorProviderOptions.t or_undefined) () :
       Disposable.t =
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2218,9 +2218,9 @@ end
 module FileSystemWatcher = struct
   include Interface.Make ()
 
-  include
-    [%js:
-    val onDidChange : t -> callback:(Uri.t -> unit) -> Disposable.t [@@js.call]]
+  module OnDidChange = Event.Make (Uri)
+
+  include [%js: val onDidChange : t -> OnDidChange.t [@@js.get]]
 end
 
 module Workspace = struct
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2336,7 +2336,7 @@ module CustomDocument = struct
 
     val dispose : t -> unit [@@js.call]
 
-    val create : uri:Uri.t -> dispose:(unit -> unit) -> unit -> t [@@js.builder]]
+    val create : uri:Uri.t -> dispose:(unit -> unit) -> t [@@js.builder]]
 end
 
 module TreeItemLabel = struct
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2330,27 +2330,13 @@ end
 module CustomDocument = struct
   include Interface.Make ()
 
-  module OnDidChange = Event.Make (Js.Unit)
-  module OnDidDispose = Event.Make (Js.Unit)
-
   include
     [%js:
     val uri : t -> Uri.t [@@js.get]
 
     val dispose : t -> unit [@@js.call]
 
-    val onDidChange : t -> OnDidChange.t [@@js.get]
-
-    val onDidDispose : t -> OnDidDispose.t [@@js.get]
-
-    val create :
-         uri:Uri.t
-      -> ?onDidChange:OnDidChange.t
-      -> ?onDidDispose:OnDidDispose.t
-      -> dispose:(unit -> unit)
-      -> unit
-      -> t
-      [@@js.builder]]
+    val create : uri:Uri.t -> dispose:(unit -> unit) -> unit -> t [@@js.builder]]
 end
 
 module TreeItemLabel = struct
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -2329,11 +2329,7 @@ end
 
 module CustomDocument = struct
   module type T = sig
-    type t
-
-    val t_of_js : Ojs.t -> t
-
-    val t_to_js : t -> Ojs.t
+    include Js.T
 
     val uri : t -> Uri.t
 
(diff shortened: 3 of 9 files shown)


src/cm_editor.ml
```diff
@@ -0,0 +1,100 @@
...
+
+let openCustomDocument _instance ~(uri : Uri.t) ~openContext:_ ~token:_ :
+    Cm_document.t Promise.t =
+  let document = Cm_document.create ~uri in
+  Promise.resolve document
+
+let register (extension : ExtensionContext.t) (instance : Extension_instance.t)
+    =
+  let module CustomReadonlyEditorProvider =
+    CustomReadonlyEditorProvider.Make (Cm_document) in
```
tatchi, line 85: According to the TS types, `CustomReadonlyEditorProvider` works only with [objects that implement the `CustomDocument` interface](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/vscode/index.d.ts#L8188).

The functor here accepts any modules, not only those that implement a specific interface. Not sure if there's a way to enforce that in OCaml?

mnxn: Would this work?

```ocaml
module Make (T : module type of CustomDocument) = ...
```

src-bindings/vscode/vscode.ml
```diff
@@ -2215,6 +2215,14 @@ module TextDocumentContentProvider = struct
       [@@js.builder]]
 end
 
+module FileSystemWatcher = struct
+  include Interface.Make ()
+
+  include
+    [%js:
+    val onDidChange : t -> callback:(Uri.t -> unit) -> Disposable.t [@@js.call]]
```
mnxn, line 2223: You should use the `Event` functor here.

src-bindings/vscode/vscode.ml
```diff
@@ -2310,6 +2327,32 @@ module TreeItemCollapsibleState = struct
...
+
+  include
+    [%js:
+    val uri : t -> Uri.t [@@js.get]
+
+    val dispose : t -> unit [@@js.call]
+
+    val onDidChange : t -> OnDidChange.t [@@js.get]
+
+    val onDidDispose : t -> OnDidDispose.t [@@js.get]
```
mnxn, line 2344: I don't think `CustomDocument` has these properties? It's not listed in the api documentation: https://code.visualstudio.com/api/references/vscode-api#CustomDocument

ulugbekna:
`.bc` are ocaml bytecode files, so you get those easily:

```
$ cat > test.ml
print_endline "Hi @tatchi!"
$ ocamlc test.ml -o test.bc
$ ocamlrun test.bc
Hi @tatchi!
```

Having a dune file such as 

```
(executable
 (name hello_world)
 (modes byte exe))
```

and building your project should also produce `hello_world.bc` in `_build`. 

A little more info: https://discuss.ocaml.org/t/ocamlc-bytecode-file-vs-dune-bytecode-bc-file/7395

src/cm_editor.ml
tatchi: When I do that I've another error elsewhere that says:

```bash
File "src-bindings/vscode/vscode.ml", line 3174, characters 41-42:
3174 |       CustomReadonlyEditorProvider.Make (T) in
                                                ^
Error: Signature mismatch:
       ...
       The value `create' is required but not provided
       File "src-bindings/vscode/vscode.ml", line 2353, characters 4-80:
         Expected declaration
       The value `dispose' is required but not provided
       File "src-bindings/vscode/vscode.ml", line 2351, characters 4-39:
         Expected declaration
       The value `uri' is required but not provided
       File "src-bindings/vscode/vscode.ml", line 2349, characters 4-35:
         Expected declaration
```

I think that makes sense given how `registerCustomReadonlyEditorProvider` is defined:

```ocaml
let registerCustomReadonlyEditorProvider (type a)
      (module T : Js.T with type t = a) ~(viewType : string)
      ~(provider : a CustomReadonlyEditorProvider.t)
      ?(options : RegisterCustomEditorProviderOptions.t or_undefined) () :
      Disposable.t =
    let module CustomReadonlyEditorProvider =
      CustomReadonlyEditorProvider.Make (T) in
    let provider = [%js.of: CustomReadonlyEditorProvider.t] provider in
    registerCustomReadonlyEditorProvider ~viewType ~provider ?options ()
```

I tried to create a `module type T = sig ...` inside of the `CustomDocument` module so that I could change `(module T : Js.T with type t = a)` by `(module T : CustomDocument.T with type t = a)` but then I've another error that I have no clue how to fix 😅

src-bindings/vscode/vscode.ml
tatchi: Thanks, that makes sense indeed. I've addressed that in https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/0de141cff215a4f5170679e535a5fcadcc6ffe48

src-bindings/vscode/vscode.ml
tatchi: Yes, that's correct, and actually we don't even use the `onDidDispose`. I initially didn't include them to closely mirror the Vscode API. My plan was to add them in the `Cm_document` module. It would become:

```diff
diff --git a/src/cm_editor.ml b/src/cm_editor.ml
index 6ba2ced..a9243a3 100644
--- a/src/cm_editor.ml
+++ b/src/cm_editor.ml
@@ -7,8 +7,23 @@ module Cm_document : sig
   val content : t -> (string, string) result Promise.t
 
   val create : uri:Uri.t -> t
+
+  val onDidChange : t -> Js.Unit.t Event.t
 end = struct
   include CustomDocument
+  module OnDidChange = Event.Make (Js.Unit)
+
+  include
+    [%js:
+    val onDidChange : t -> OnDidChange.t [@@js.get]
+
+    val create :
+         uri:Uri.t
+      -> ?onDidChange:OnDidChange.t
+      -> dispose:(unit -> unit)
+      -> unit
+      -> t
+      [@@js.builder]]
 
   let content t =
     let uri = uri t in
@@ -20,8 +35,6 @@ end = struct
     let module EventEmitter = EventEmitter.Make (Js.Unit) in
     let onDidChange_event_emitter = EventEmitter.make () in
     let onDidChange_event = EventEmitter.event onDidChange_event_emitter in
-    let onDidDispose_event_emitter = EventEmitter.make () in
-    let onDidDispose_event = EventEmitter.event onDidDispose_event_emitter in
     let watcher =
       Workspace.createFileSystemWatcher
         (`String (Uri.fsPath uri))
@@ -32,10 +45,8 @@ end = struct
         ~listener:(fun _ -> EventEmitter.fire onDidChange_event_emitter ())
         ()
     in
-    create ~uri ~onDidChange:onDidChange_event ~onDidDispose:onDidDispose_event
-      ~dispose:(fun () ->
-        EventEmitter.fire onDidDispose_event_emitter ();
-        Disposable.dispose disposable)
+    create ~uri ~onDidChange:onDidChange_event
+      ~dispose:(fun () -> Disposable.dispose disposable)
       ()
 end
```

But it means that we would have to preprocess the file with the `gen_js_api` ppx, i.e adding the following to the `src/dune` file:

```bash
 (preprocess
  (pps gen_js_api.ppx))
```

Not sure if that's something we want to avoid or if it's totally ok?

tatchi:
> `.bc` are ocaml bytecode files, so you get those easily:
> 
> ```
> $ cat > test.ml
> print_endline "Hi @tatchi!"
> $ ocamlc test.ml -o test.bc
> $ ocamlrun test.bc
> Hi @tatchi!
> ```
> 
> Having a dune file such as
> 
> ```
> (executable
>  (name hello_world)
>  (modes byte exe))
> ```
> 
> and building your project should also produce `hello_world.bc` in `_build`.
> 
> A little more info: https://discuss.ocaml.org/t/ocamlc-bytecode-file-vs-dune-bytecode-bc-file/7395

Oh cool, thanks for the great explanations 🤗 I just tested it and it seems to work well with these files 👌

What about the other extensions: .cmx, .cmxa, .cma, .cmxs ? Should we support them too?

src/cm_editor.ml
mnxn: I see. I think adding the `T` module type is the correct solution. What is the error that you get if you include it?

src-bindings/vscode/vscode.ml
mnxn: I think it's fine to add the `preprocess` field. It's more correct to have `onDidChange` in the `Cm_document` module.

mnxn:
> What about the other extensions: .cmx, .cmxa, .cma, .cmxs ? Should we support them too?

If `ocamlobjinfo` works for those file types, we should definitely support them.

src-bindings/vscode/vscode.ml
tatchi: Thanks, done in https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/40a34fbec022ee31e1c7aeeee3bf457dd42c1961

src/cm_editor.ml
tatchi: There was two errors. I fixed one by adding the `dispose` argument to the `Cm_document.create` function to match the `CustomDocument.create` function (which I think makes sense).

The second error is:

```ocaml
File "src-bindings/vscode/vscode.ml", line 3174, characters 41-42:
3174 |       CustomReadonlyEditorProvider.Make (T) in
                                                ^
Error: Signature mismatch:
       ...
       The module type `T' is required but not provided
       File "src-bindings/vscode/vscode.ml", lines 2331-2343, characters 2-5:
         Expected declaration
make: *** [all] Error 1    
```

I don't know how to fix that one... I pushed a commit https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/2c60512e0d9f72fd5886996ed49517978e865aeb with the changes if you want to have a look.

mnxn: You need to change `module type of CustomDocument` to `CustomDocument.T` here:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/2c60512e0d9f72fd5886996ed49517978e865aeb/src-bindings/vscode/vscode.ml#L2905

tatchi: Oops 🤦‍♂️ I also had to apply the following change and now it works! 🥳 Thanks for your help! 🤗❤️

```diff
@@ -2373,7 +2373,7 @@ module Window : sig
     -> Disposable.t
 
   val registerCustomReadonlyEditorProvider :
-       'a Js.t
+       (module CustomDocument.T with type t = 'a)
     -> viewType:string
     -> provider:'a CustomReadonlyEditorProvider.t
     -> ?options:RegisterCustomEditorProviderOptions.t
```

src-bindings/vscode/vscode.ml
```diff
@@ -2310,6 +2327,28 @@ module TreeItemCollapsibleState = struct
...
 
+module CustomDocument = struct
+  module type T = sig
+    include Js.T
+
+    val uri : t -> Uri.t
+
+    val dispose : t -> unit
+
+    val create : uri:Uri.t -> dispose:(unit -> unit) -> t
```
mnxn, line 2338: I think the `create` function shouldn't be part of module type since modules that `include` this one might have more fields.

tatchi: Ah yes, that definitely makes sense. Thanks 👍

tatchi: Addressed in https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/446401d6471aabad92a3ff580e234ce41bc60def

src-bindings/vscode/vscode.mli
```diff
@@ -782,6 +782,24 @@ module CancellationToken : sig
...
+    val uri : t -> Uri.t
+
+    val dispose : t -> unit
+  end
+
+  include Js.T
+
+  val uri : t -> Uri.t
+
+  val dispose : t -> unit
```
mnxn, line 798: I believe these lines can be replaced with just `include T`.

src/cm_editor.ml
```diff
@@ -0,0 +1,107 @@
+open Import
+open Interop
+
+module Cm_document : sig
+  include module type of CustomDocument
```
mnxn, line 5: Should be `include CustomDocument.T` now that we had a dedicated module type.

tatchi:
Thanks for the review @mnxn 🤗 I should have addressed all your comments :)

src-bindings/vscode/vscode.mli
```diff
@@ -782,6 +782,22 @@ module CancellationToken : sig
...
+module CustomDocument : sig
+  module type T = sig
+    include Js.T
+
+    val uri : t -> Uri.t
+
+    val dispose : t -> unit
+  end
+
+  include Js.T
```
mnxn, line 794: This line is unnecessary since `T` already includes `Js.T`.

tatchi: Good catch, thanks :)

mnxn (approved):

src/cm_editor.ml
```diff
@@ -0,0 +1,107 @@
...
+    val onDidChange : t -> OnDidChange.t [@@js.get]
+
+    val create :
+      uri:Uri.t -> onDidChange:OnDidChange.t -> dispose:(unit -> unit) -> t
+      [@@js.builder]]
+
+  let content t =
+    let uri = uri t in
+    let file_path = Uri.fsPath uri in
+    let bin = Path.of_string "ocamlobjinfo" in
```
ulugbekna, line 27: Shouldn’t this binary be invoked within a sandbox environment? 

(I’ll have time to have a full review this Friday if the PR can wait)

tatchi: Ah yes, I think you're right. I'll make the changes 👍

> (I’ll have time to have a full review this Friday if the PR can wait)

No problem, it can definitely wait 😁

src/cm_editor.ml
```diff
@@ -0,0 +1,107 @@
...
+  Promise.resolve document
+
+let register (extension : ExtensionContext.t) (instance : Extension_instance.t)
+    =
+  let module CustomReadonlyEditorProvider =
+    CustomReadonlyEditorProvider.Make (Cm_document) in
+  let editorProvider =
+    CustomReadonlyEditorProvider.create
+      ~resolveCustomEditor:(resolveCustomEditor instance)
+      ~openCustomDocument:(openCustomDocument instance)
```
ulugbekna, line 96: am I missing something or `openCustomDocument` definition doesn't use `instance` so we could simply remove that arg?

ulugbekna (changes requested):
I looked into `TextDocumentContentProvider` as a lighter alternative to using custom documents because custom documents seem quite heavy. According to its doc: 

> A text document content provider allows to add readonly documents to the editor, such as source from a dll or generated html from md.
> 
> Content providers are registered for a uri-scheme. When a uri with that scheme is to be loaded the content provider is asked.

It seems we could use that and avoid using a custom document, which seems a (heavy-weight) overkill for this feature. I also suppose that then we would get file-change watching for free. What do you think? 

Here's an example of how to use that API and it seems the right fit for this task: https://github.com/microsoft/vscode-extension-samples/tree/main/contentprovider-sample

tatchi:
> I looked into `TextDocumentContentProvider` as a lighter alternative to using custom documents because custom documents seem quite heavy. According to its doc:
> 
> > A text document content provider allows to add readonly documents to the editor, such as source from a dll or generated html from md.
> > Content providers are registered for a uri-scheme. When a uri with that scheme is to be loaded the content provider is asked.
> 
> It seems we could use that and avoid using a custom document, which seems a (heavy-weight) overkill for this feature. I also suppose that then we would get file-change watching for free. What do you think?
> 
> Here's an example of how to use that API and it seems the right fit for this task: https://github.com/microsoft/vscode-extension-samples/tree/main/contentprovider-sample

Thanks for your review @ulugbekna 🤗 `TextDocumentContentProvider` is what I wanted to use in the first place for the reasons you mentioned above. According to my understanding, `TextDocumentContentProvider` can't be used when opening a file. It only works when we programmatically fire an action, like in your example when a user clicks on `Show all references`.

See this comment and the reply below: https://github.com/Microsoft/vscode/issues/53121#issuecomment-405241450

CHANGES.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Support for opening `.cmi`, `.cmt`, `cmti`, `.cmo`, `.cma`, `.cmx`, `.cmxa`,
+  `.cmxs` and `.bc` files. (#798)
```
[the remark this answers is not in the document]
tatchi: Definitely better 😁 Thanks!

package.json
```diff
@@ -946,6 +947,39 @@
             "filenamePattern": "*.mli"
           }
         ]
+      },
+      {
+        "viewType": "cm-files-editor",
+        "displayName": "Cm files preview",
```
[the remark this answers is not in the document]
tatchi: Yes, I agree. I was definitely not sure what to use as `displayName` here. What you suggest is definitely better 👍

src/cm_editor.ml
tatchi: Oh yes, you're totally right. `resolveCustomEditor ` doesn't use it either. Removed in https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/6bbbdc8c156543a549b3162da0656d16a2bf62b0

src/cm_editor.ml
tatchi: I made the change in https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/e881317633ebabbf9638f6c19f056b078c8c1c39

src/cm_editor.ml
```diff
@@ -0,0 +1,111 @@
...
+    Cm_document.onDidChange document
+      ~listener:(fun () ->
+        let (_ : unit Promise.t) = update_content document in
+        ())
+      ()
+  in
+  let (_ : Disposable.t) =
+    WebviewPanel.onDidDispose webviewPanel
+      ~listener:(fun () -> Disposable.dispose disposable)
+      ()
```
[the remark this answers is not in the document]
tatchi: Good catch! I fixed it in https://github.com/ocamllabs/vscode-ocaml-platform/pull/798/commits/20c9896304cece78799246fe5eb7766aa2d160f1

I went for using a `Stack` since it's a mutable structure. I could also just use a regular `List` and mutate it if you prefer.

PR #831, opened 2022-01-25
fix detection of opam root when no switch is selected

### LINKED ISSUE #701: Local switches not detected on opam 2.1
I just upgraded to opam 2.1, created a new local switch at the root of the project, installed `ocaml-lsp-server` and realised that I can't select the local switch for use with the extension. Non-local switches seem to work as usual.

I first tried to work around it using a custom sandbox `eval $(opam env) && $prog $args`. This allowed me to create terminals on the local switch, and `which ocamllsp` shows the correct path to the LSP server, but the extension still fails to start it, notifying me that "ocaml-lsp-server is not installed".

For reference, I'm using `ocamllabs.ocaml-platform` v1.8.4 on VS Code 1.60.0 and `ocaml-lsp-server` 1.7.0.

Khady:
fix #701

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix the detection of opam's root directory when no switch is detected (#831)
+
 - Warn if the extension sees not the latest OCaml-LSP version compatible with
   the OCaml distribution installed in the current sandbox.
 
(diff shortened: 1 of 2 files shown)


rgrinberg:
Thanks

PR #837, opened 2022-01-31
Update instructions surrounding ReasonML/OCaml + Rescript/Bucklescript Syntax Support.

chrischen:
There is a section in the readme referencing using ReasonML/Ocaml syntax in ReScript. I was told that specific use case is no longer supported and it could be helpful for the (few) users still stuck doing that trying to migrate off reason-language-server. I also added a workaround using Melange which allows usage of ocaml-lsp-server while still compiling to JS. The wording around it was carefully chosen to try to avoid any confusion/conflict between Rescript and Melange.


rgrinberg:
Thanks

PR #852, opened 2022-02-10
update to ocaml-lsp 1.10.0

ulugbekna:


rgrinberg:
It's alright, it's for this reason I prefer an uncoupled release process. I need to think a little more on how to make this check a little less harmful first.

ulugbekna:
We could only show the message based on the latest version available on opam (without using opam update) if there's internet and not show message otherwise. 

We could add a code action that warns that a vscode release needs to happen after an ocaml-lsp release. 

We could make this opt-in but then it's unlikely anyone will discover that option really and it interferes with plans for automatic ocamllsp upgrading flow we could have.

PR #859, opened 2022-02-12
fix: remove intrusive version checking

rgrinberg:
Only notify users when all of the following conditions hold:

* The running version of ocamllsp is "official". We don't offer
  suggestions for development versions.

* We know something about the user's version ocamllsp. If the version
  isn't recognized by us, we do not second guess the user.

* We are certain the user is able to upgrade to a newer version. If the
  user is forced to an older version of lsp, we allow him to use the
  plugin without notifications.

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -158,25 +158,15 @@ end = struct
   (** Shows appropriate message for when the [ocaml-lsp] in use by the extension
       doesn't support jumping to holes. *)
   let ocaml_lsp_doesn't_support_holes instance ocaml_lsp =
-    let suggestion =
-      match
-        Ocaml_lsp.is_version_up_to_date ocaml_lsp
-          (Extension_instance.ocaml_version_exn instance)
-      with
-      | Ok is_up_to_date ->
-        if is_up_to_date then
-          (* ocamllsp is "up-to-date", so user needs newer ocaml to get newer
-             ocamllsp, which supports typed holes *)
-          "The extension requires a newer version of `ocamllsp`, which needs a \
-           new version of OCaml. Please, consider upgrading the OCaml version \
-           used in this sandbox."
-        else "Consider upgrading the package `ocaml-lsp-server`."
-      | Error (`Msg m) ->
-        sprintf "There is something wrong with your `ocamllsp`. Error: %s" m
-    in
-    show_message `Warn
-      "The installed version of `ocamllsp` does not support typed holes. %s"
-      suggestion
+    match
+      Ocaml_lsp.is_version_up_to_date ocaml_lsp
+        (Extension_instance.ocaml_version_exn instance)
+    with
+    | Ok () -> ()
+    | Error (`Msg msg) ->
+      show_message `Warn
+        "The installed version of `ocamllsp` does not support typed holes. %s"
+        msg
 
   let current_cursor_pos text_editor =
     let selection = TextEditor.selection text_editor in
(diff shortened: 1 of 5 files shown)


src/extension_commands.ml
```diff
@@ -158,25 +158,15 @@ end = struct
...
-      "The installed version of `ocamllsp` does not support typed holes. %s"
-      suggestion
+    match
+      Ocaml_lsp.is_version_up_to_date ocaml_lsp
+        (Extension_instance.ocaml_version_exn instance)
+    with
+    | Ok () -> ()
+    | Error (`Msg msg) ->
+      show_message `Warn
+        "The installed version of `ocamllsp` does not support typed holes.\n%s"
```
ulugbekna, line 168: IIRC, messages shown in vscode do not support formatting, eg new lines. The message should be plain text

src/ocaml_lsp.ml
```diff
@@ -53,64 +53,76 @@ type t =
...
   in
   Result.map_error res ~f:(fun err ->
       let msg =
         match err with
-        | `Ocaml_version_not_supported v ->
+        | `Newer_available (old, new_) ->
+          let upgrade =
+            match old with
+            | None -> sprintf "to %s" new_
+            | Some old -> sprintf "%s -> %s" old new_
```
ulugbekna, line 120:
```suggestion
            | Some old -> sprintf "%s to %s" old new_
```

src/ocaml_lsp.ml
```diff
@@ -53,64 +53,76 @@ type t =
...
-        | `Unexpected `Missing_serverInfo
-        | `Unexpected `ServerInfo_version_missing ->
-          "The extension expected the server version to be sent by `ocamllsp`. \
-           It is missing. Please, consider upgrading the package \
-           `ocaml-lsp-server`."
-        | `Unexpected `Unable_to_parse_version ->
-          "The extension was unable to parse `ocamllsp` version. That's \
-           strange. Consider filing an issue on the project GitHub with the \
-           version of your `ocamllsp`."
+            "The current version of ocamllsp is out of date. Consider \
```
ulugbekna, line 123:
```suggestion
            "There is a newer version of `ocamllsp` available. Consider \
```

"out of date" means it may not be supported already?

ulugbekna (commented):
LGTM except for some nitpicks

src/ocaml_lsp.ml
```diff
@@ -53,64 +53,76 @@ type t =
...
+         ; "1.6.0"
+         ; "1.6.1"
+         ; "1.7.0"
+         ; "1.8.0"
+         ; "1.8.1"
+         ; "1.8.2"
+         ; "1.8.3"
+         ; "1.9.0"
+        |] )
+    ; ((4, 13), [| "1.9.1"; "1.10.0"; "1.10.1" |])
```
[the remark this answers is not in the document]
ulugbekna: Btw, ocamlc version library should have values corresponding to ocamlc versions, ie we could use `Ocaml_version.v4_12` instead of `(4, 12)`

src/ocaml_lsp.ml
rgrinberg: Can you confirm that we can use markdown in the messages? I believe you said it was plaintext only earlier.

src/ocaml_lsp.ml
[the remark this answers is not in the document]
rgrinberg: >  ie we could use Ocaml_version.v4_12 instead of (4, 12)

This could work, but is it really clarifying anything? We are mapping version OCaml versions prefixes to ocamllsp versions.

I'll add some comments

rgrinberg:
> It’s not formatted as markdown, but marking the binary name in the message
Why don't we just use normal quotes then? Backticks are weird in plaintext.

ulugbekna:
Normal quotes are fine with me

Ulugbek

Issue #864, opened 2022-02-14
Dune's `rule` stanza in `subdir` is not highlighted properly


tmattio:
The following dune code doesn't seem to be highlighted properly with VSCode 1.9.0:

```
(subdir
 asset/
 (rule
  (target main.css)
  (deps
   (:data ../tailwind.config.js ../postcss.config.js)
   (source_tree ../node_modules)
   (source_tree ../src/ocamlorg_frontend))
  (action
   (chdir
    %{workspace_root}
    (run
     node
     node_modules/postcss-cli/index.js
     src/ocamlorg_frontend/css/styles.css
     -o
     %{target})))))
```

![image](https://user-images.githubusercontent.com/6162008/153873320-9fc77a02-c4d3-4f23-8c0b-e1a2f2a6f949.png)

cc @SaySayo

SaySayo:
I'll look into it now

SaySayo:
<img width="719" alt="Screenshot 2022-02-14 at 20 06 28" src="https://user-images.githubusercontent.com/75541866/153929667-6675aebb-0e29-4db6-9772-98b8a660bbd1.png">
@tmattio I just made these changes. Are these good now?

PR #870, opened 2022-02-14
Fix dune subdir stanza syntax highlighting

SaySayo:
Fixes (#864)


tmattio:
Thanks @SaySayo 🙂 

The `subdir` stanza works by evaluating the dune stanza in a subdirectory. So it should support *every* dune stanza, not only the rule one. To do this I think we can extract all the patterns in a scope, and include this in the `subdir` patterns (assuming the textmate grammar support recursivity).

SaySayo:
That's a great idea! I'll do that now

syntaxes/dune.json
```diff
@@ -1235,9 +1235,214 @@
...
+                      "include": "#general"
+                    }
+                  ]
+                },
                 {
                   "include": "#action"
                 },
+                {
+                  "include": "#variables"
+                },
```
mnxn, line 1445: What's the point of including `#variables` here?

The rules in "variables" are just for highlighting the name in between `%{` and `}`.
The actual syntax for the full variable `%{...}` is already included in `#general`.

mnxn:
> The `subdir` stanza works by evaluating the dune stanza in a subdirectory. So it should support _every_ dune stanza, not only the rule one. To do this I think we can extract all the patterns in a scope, and include this in the `subdir` patterns (assuming the textmate grammar support recursivity).

I think extracting all of the patterns into a new scope is actually unnecessary. You can make the patterns of `subdir` include `$self` and that should highlight everything

I believe this should work:
```json
"comment": "subdir",
"begin": "\\([[:space:]]*(subdir)\\b",
"end": "\\)",
"beginCaptures": {
  "1": {
    "name": "keyword.language.dune"
  }
},
"patterns": [{ "include": "$self" }]
```

syntaxes/dune.json
```diff
@@ -1225,6 +1222,9 @@
             }
           },
           "patterns": [
+            { 
+              "include": "$self" 
+            },
             {
```
mnxn, line 1228: I think the rest of this is unnecessary now since "$self" already recognizes a `rule`.

SaySayo: Oh right, Thank you

mnxn (approved):

mnxn:
Thanks!

PR #878, opened 2022-02-17
Add 4.14 preview lsp version

yannl35133:
Didn't test this in any way


yannl35133:
I guess ocamlformat doesn't like the comment in the middle of everything

PR #884, opened 2022-02-21
fix: restore compat with opam 2.0.x

rgrinberg:
check --version of opam before using --global

cc @Khady

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.9.4
+
+- Restore compatibility with opam 2.1.x (#884)
+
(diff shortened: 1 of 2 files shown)


mnxn (approved):
Works great, thanks for the fix.
This was quite annoying on Windows since it uses an older opam version.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.9.4
+
+- Restore compatibility with opam 2.1.x (#884)
```
Khady, line 5:
```
- Restore compatibility with opam 2.0.x (#884)
```

PR #902, opened 2022-03-15
Upgrade dune and ocamlformat

smorimoto:


ulugbekna:
I think this has been already done. Thanks Sora! :-)

PR #911, opened 2022-04-04
Add a setting to Enable/Disable Utop usage

OscarLahaie:
I have modified package settings and repl script to let the user decides if he wants to use utop or not.

### DIFF
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -20,7 +20,10 @@ let ocaml_utop_setting =
   Settings.create_setting ~scope:Global ~key:"ocaml.useUtop"
     ~of_json:Jsonoo.Decode.bool ~to_json:Jsonoo.Encode.bool
 
-let use_utop () = Settings.get ocaml_utop_setting
+let use_utop () = 
+  match (Settings.get ocaml_utop_setting) with
+  | Some x -> x
+  | None -> true
 
 module Repl_args = struct
   type t = string list option
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -16,6 +16,12 @@ module Repl_path = struct
   let t = Settings.create_setting ~scope:Global ~key ~of_json ~to_json
 end
 
+let ocaml_utop_setting =
+  Settings.create_setting ~scope:Global ~key:"ocaml.useUtop"
+    ~of_json:Jsonoo.Decode.bool ~to_json:Jsonoo.Encode.bool
+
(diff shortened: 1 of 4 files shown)


OscarLahaie:
This should now work without the errors

mnxn:
We already have a few settings that configure the REPL: `ocaml.repl.path` and `ocaml.repl.args`.
Since this option is only for the REPL, it should have the same prefix. It should be `ocaml.repl.useUtop`.

You should also mention that this is for the REPL in the `description` field.

OscarLahaie:
It should be good now 😉

src/repl.ml
```diff
@@ -20,7 +20,7 @@ let ocaml_utop_setting =
   Settings.create_setting ~scope:Global ~key:"ocaml.useUtop"
     ~of_json:Jsonoo.Decode.bool ~to_json:Jsonoo.Encode.bool
 
-let use_utop () = Settings.get ocaml_utop_setting
+let use_utop = Settings.get ocaml_utop_setting
```
mnxn, line 23: You should keep the `()` argument to make it a function. Without the argument, the setting is only read once and the user would have to restart the extension after changing the utop setting.

OscarLahaie: Done

OscarLahaie:
Why are some tests failing ?

mnxn:
There was a trailing comma in `package.json` that caused an error. I fixed it.
If the new checks don't fail, I will merge the PR.

OscarLahaie:
It seems that i still have a type error

mnxn:
> It seems that i still have a type error

Indeed. `Settings.get` returns a `bool option` in this case. You should handle the `None` possibility in the body of `use_utop` so it returns just a `bool`.

OscarLahaie:
> > It seems that i still have a type error
> 
> Indeed. `Settings.get` returns a `bool option` in this case. You should handle the `None` possibility in the body of `use_utop` so it returns just a `bool`.

Should i return a `false` if i get a `None` ?

mnxn:
I think it should be `true` since that's the default in `package.json`.

OscarLahaie:
> I think it should be `true` since that's the default in `package.json`.

Ok i modify this

OscarLahaie:
Now it is not a problem from my additions normally

mnxn:
Thanks!

PR #928, opened 2022-04-12
Add indentation rules for comments

### LINKED ISSUE #182: Indentation of comments can be improved
Considering entering `RET` at the cursor:

```
(* foo| *)
```

The result is:

```
(* foo
|*)
```

The expected result is:

```
(* foo
   |*)
```

OscarLahaie:
I have modified the ocaml.json file to resolve #182 issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Fix external syntax highlighting for embedded ocaml source in comments (#906)
 - Add `ocaml.repl.useUtop` setting to control whether to use Utop for the REPL
   (#911)
-- Fix the indentation values for simple comments and documentation comments (#182)
+- Fix the indentation values for simple comments and documentation comments
+  (#182)
 
 ## 1.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 - Fix external syntax highlighting for embedded ocaml source in comments (#906)
 - Add `ocaml.repl.useUtop` setting to control whether to use Utop for the REPL
   (#911)
-- Fix the indentation values for simple comments and documentation comments
-  (#182)
+- Add indentation rules for comments
+  (#928)
 
 ## 1.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fix external syntax highlighting for embedded ocaml source in comments (#906)
 - Add `ocaml.repl.useUtop` setting to control whether to use Utop for the REPL
   (#911)
+- Fix the indentation values for simple comments and documentation comments (#182)
 
 ## 1.10.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@
 - Fix external syntax highlighting for embedded ocaml source in comments (#906)
 - Add `ocaml.repl.useUtop` setting to control whether to use Utop for the REPL
   (#911)
-- Add indentation rules for comments
-  (#928)
+- Add indentation rules for comments (#928)
 
 ## 1.10.0
 
--- a/languages/ocaml.json
+++ b/languages/ocaml.json
@@ -29,10 +29,18 @@
   },
   "onEnterRules": [
     {
+      "beforeText": "\\(\\*\\*",
+      "action": { 
+        "indent": "none",
+        "appendText": "    "
+      }
+    },
+    {
       "beforeText": "\\(\\*",
       "action": { 
         "indent": "none",
-        "appendText": "   "}
+        "appendText": "   "
+      }
     }
   ]
 }
--- a/languages/ocaml.json
+++ b/languages/ocaml.json
@@ -41,6 +41,20 @@
         "indent": "none",
         "appendText": "   "
       }
+    },
+    {
+      "beforeText": "\\*\\)",
+      "action": {
+        "indent": "outdent",
+        "removeText": 3
+      }
+    },
+    {
+      "beforeText": "\\*\\*\\)",
+      "action": {
+        "indent": "outdent",
+        "removeText": 4
+      }
     }
   ]
 }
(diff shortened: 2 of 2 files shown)


OscarLahaie:
You are right, there was a way through `OnEnterRules`. It remains to be seen if this is what you expect.

mnxn:
Oh cool! I didn't know there was a way to do this in the language configuration file already. That's exactly the behavior I expect.

You should also add a rule for documentation comments that start with `(**`. Those need 4 spaces:
```ocaml
(** foo
    |*)
```
Make sure to add it before the normal comment rule so it gets matched first.

OscarLahaie:
This should now work for both types of comments

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@
 - Fix external syntax highlighting for embedded ocaml source in comments (#906)
 - Add `ocaml.repl.useUtop` setting to control whether to use Utop for the REPL
   (#911)
-- Fix the indentation values for simple comments and documentation comments (#182)
+- Fix the indentation values for simple comments and documentation comments
+  (#182)
```
mnxn, line 9: This should be the PR number, 928.

mnxn: Also, it's not really a "fix" since the issue was a lack of a feature and not a bug. I think a more accurate description would start with
> Add indentation rules for ...

mnxn:
Sorry, one more thing I think we need to handle is the cursor position after you close a comment.
With this PR it is
```
(* foo
   *)
   |
```

I think the preferable behavior would be to reset it to the position before the comment started:
```
(* foo
   *)
|
```

I haven't tested this thoroughly, but I think this rule should work:
```json
{
  "beforeText": "\\*\\)",
  "action": {
    "indent": "outdent",
    "removeText": 2
  }
}
```

`"removeText": 2` removes 2 of the indentation characters and then `"indent": "outdent"` removes 1 or 2 characters so it works for both normal and documentation comments.

OscarLahaie:
The outindentation is not working for now. I will look at it later

languages/ocaml.json
```diff
@@ -26,5 +26,35 @@
...
+        "removeText": 3
+      }
+    },
+    {
+      "beforeText": "\\*\\*\\)",
+      "action": {
+        "indent": "outdent",
+        "removeText": 4
+      }
+    }
```
mnxn, line 58: Documentation comments are typically still closed with `*)`, so there should only be one rule for the comment closing.

In other words, the `*)` has to outdent both 3 and 4 spaces.

languages/ocaml.json
```diff
@@ -26,5 +26,35 @@
...
+      "action": {
+        "indent": "none",
+        "appendText": "   "
+      }
+    },
+    {
+      "beforeText": "\\*\\)",
+      "action": {
+        "indent": "outdent",
+        "removeText": 3
```
mnxn, line 49: Did the snippet I provide not work?

This currently removes too many spaces. `outdent` removes up to one indentation level so 1 or 2 spaces. And then `removeText` removes an _additional_ 3 spaces for a total of 4 or 5 spaces removed.

We need to handle 3 or 4 spaces, so the `removeText` needs to be smaller.

```suggestion
        "indent": "outdent",
        "removeText": 2
```

OscarLahaie: I didn't have much time so I tried to do it in one go but it didn't work out.

languages/ocaml.json
mnxn: Seems to work fine for me inside a function with `"removeText": 2`:

recording

OscarLahaie: Oh yes, it does work. Perfect !
Thank you !

OscarLahaie:
Good to be merge, no ?

mnxn:
Sorry for the delay.

I found an issue with the indentation after a single-line comment:
recording

I pushed a changed with lookahead/lookbehind that should fix it:
recording

Can you please check it out and make sure everything works as expected?
If there are no more problems, I will merge the PR.

OscarLahaie:
I found an error and i don't know how to fix it.
When I put a selection in comment the first indentation is not taken into account so when I press enter the indentation is not good.

https://user-images.githubusercontent.com/25438086/163702317-22687b16-2430-4193-ae93-e7c601e8abd9.mp4

mnxn:
Good catch.
I now think that outdents are not possible without introducing lots of issues like these.

I believe a good compromise is just to handle the first line indent of 3 or 4 spaces and leave the indentation alone after the comment closing. It's a little annoying to decrease the indentation manually afterwards, but that was actually the behavior before this PR.

Also, a simpler way to leave the indentation alone for single-line comments is to use `"afterText": "\\*\\)"`.

Even without outdents, I think it is a good improvement over the current behavior.
I pushed the commit, please check it out again.

Before:
before

After:
after

OscarLahaie:
> Good catch. I now think that outdents are not possible without introducing lots of issues like these.
> 
> I believe a good compromise is just to handle the first line indent of 3 or 4 spaces and leave the indentation alone after the comment closing. It's a little annoying to decrease the indentation manually afterwards, but that was actually the behavior before this PR.
> 
> Also, a simpler way to leave the indentation alone for single-line comments is to use `"afterText": "\\*\\)"`.
> 
> Even without outdents, I think it is a good improvement over the current release behavior. I pushed the commit, please check it out again.
> 
> Before PR: before
> 
> After PR: after

I checked it with your patch and it seems good to me. However, I wonder if the brackets' behavior would be appropriate for comments. And if it is, maybe we can use it for this issue. 

https://user-images.githubusercontent.com/25438086/163817399-caf29b48-51fb-4ddc-a12d-a1cf933451cf.mp4

OscarLahaie:
Ok, I tried to add `(* *)` and `(** *)` to the brackets' list, but it is not viable. This behavior is not valid for comments.
However, after checking, I think that you are right this is the best compromise for now.

mnxn:
Thanks!

Issue #791, opened 2021-12-05
Handle empty switches in "opam switches" tree view


ulugbekna:
If one has an empty switch (`opam switch create empty --empty`), the extension:

1. Shows an error message with file not found error 

![image](https://user-images.githubusercontent.com/16353531/144764085-697fbb25-e311-4ea8-81c2-6e68d412c744.png)

2. The error prevents "opam switches" tree view from showing other switches apparently

![image](https://user-images.githubusercontent.com/16353531/144764157-c7a654e8-353d-4fb4-80b7-358c95508510.png)

PR #940, opened 2022-04-21
Fix errors for empty switches

OscarLahaie:
I fixed the error caused by the empty switches. However, it remains to be seen how to handle them and consider them.
This concerns the issue #791


OscarLahaie:
It is working but I am not 100% about my code. However when the script searches packages in the empty switch you now get this error. I can make a special error message but is it useful?
image

src/opam.ml
```diff
@@ -288,8 +299,11 @@ let remove t switch packages =
 let switch_compiler t switch =
   let open Promise.Syntax in
   let+ switch_state = Switch_state.of_switch t switch in
-  let compilers = Switch_state.compilers switch_state in
-  Option.bind compilers ~f:List.hd
+  match switch_state with
+  | Some switch_state ->
+    let compilers = Switch_state.compilers switch_state in
+    Option.bind compilers ~f:List.hd
+  | None -> None
```
ulugbekna, line 306: This could look like 

```ocaml
let switch_compiler t switch =
  let open Promise.Syntax in
  let+ switch_state = Switch_state.of_switch t switch in
  let open Option.O in
  let* switch_state (* this is called "let-punning" *)
  in
  let* compilers = Switch_state.compilers switch_state in
  List.hd compilers
```

ulugbekna (commented):
I think this PR is a first step in solving the problem, but we need to think about when to show what error. I would say that a user shouldn't see any errors despite having an empty switch.

src/opam.ml
```diff
@@ -115,32 +115,43 @@ module Switch_state : sig
...
-    in
-    file_contents
+    let* file_exists = Fs.exists switch_state_filepath in
+    if file_exists then
+      let+ file_content = Fs.readFile switch_state_filepath in
+      let { OpamParserTypes.file_contents; _ } =
+        OpamParser.FullPos.string file_content switch_state_filepath
+        |> OpamParser.FullPos.to_opamfile
+      in
+      Some file_contents
```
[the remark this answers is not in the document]
OscarLahaie: Should I adopt your solution or should I keep mine ?

src/opam.ml
OscarLahaie: > (* this is called "let-punning" *)
Thank you for the info. I wasn't sure to understand correctly what it was. I have to take a deeper look at the doc of Ocaml.

src/opam.ml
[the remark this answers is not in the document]
ulugbekna: Yours is fine :-)

OscarLahaie:
image
Should we keep the arrow for empty switches ?

src/opam.ml
```diff
@@ -301,17 +301,20 @@ let switch_compiler t switch =
...
+      Promise.return (Error "Could not get the packages from the switch state")
+    | Some l ->
+      let path = Switch.path t switch in
+      Promise.List.filter_map
+        (fun name ->
+          let package_path = Opam_path.package_dir path name in
+          Package.of_path package_path)
+        l
+      |> Promise.map Result.return)
+  | None -> Promise.return [] |> Promise.map Result.return
```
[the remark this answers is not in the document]
rgrinberg: Yeah, `Promise.return (Ok [])`.

src/opam.ml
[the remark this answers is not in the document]
rgrinberg: The current solution suffers from races between the existence check and the actual read. While the proposed solution swallows parsing exceptions without logging. I would prefer a solution that suffers from neither of these problems if it's possible.

OscarLahaie:
> Should we keep the arrow for empty switches ?

(If we can remove it)

rgrinberg:
> Should we keep the arrow for empty switches ?

If you can remove it, then please do.

OscarLahaie:
image

How do I retrieve the package list from the promise?

OscarLahaie:
It seems to me that I have made good progress. However, I don't see what's left to do. So it seems to me that it should be good.

ulugbekna:
Hi, I was looking into this. See Rudi’s in-line  comment about readfile. Could you fix that?

OscarLahaie:
> Hi, I was looking into this. See Rudi’s in-line comment about readfile. Could you fix that?

Ok I will take a look 👍🏻

src/opam.ml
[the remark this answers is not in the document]
OscarLahaie: > The current solution suffers from races between the existence check and the actual read. While the proposed solution swallows parsing exceptions without logging. I would prefer a solution that suffers from neither of these problems if it's possible.

I tried something. I don't know if we need to handle `Fs.readFile switch_state_filepath` errors or if it good as it is.

src/opam.ml
```diff
@@ -131,15 +131,18 @@ end = struct
...
+      ~fulfilled:(fun file_exists ->
+        if file_exists then
+          let+ file_content = Fs.readFile switch_state_filepath in
+          let { OpamParserTypes.file_contents; _ } =
+            OpamParser.FullPos.string file_content switch_state_filepath
+            |> OpamParser.FullPos.to_opamfile
+          in
+          Some file_contents
+        else Promise.return None)
+      ~rejected:(fun x -> Promise.reject x)
```
ulugbekna, line 145: This code is effectively the same as previous and susceptible to a race condition. When you first check if the file exists and _then_ try reading it, the file that you checked to exist may already have been removed. 

I'll push a fix :-)

OscarLahaie: > This code is effectively the same as previous and susceptible to a race condition. When you first check if the file exists and _then_ try reading it, the file that you checked to exist may already have been removed.
> 
> I'll push a fix :-)

Oh right, I misunderstood. But I'm curious to see the solution.

src/treeview_switches.ml
```diff
@@ -84,7 +84,7 @@ module Dependency = struct
     let item = Vscode.TreeItem.make_label ~label ~collapsibleState () in
     Vscode.TreeItem.set_iconPath item icon;
     TreeItem.set_contextValue item (context_value dependency);
-    let+ _ =
+    let+ (_ : unit) =
```
rgrinberg, line 87: `let+ () =`

src/opam.ml
```diff
@@ -127,14 +129,30 @@ end = struct
...
-    let+ file_content = Fs.readFile switch_state_filepath in
-    let { OpamParserTypes.file_contents; _ } =
-      OpamParser.FullPos.string file_content switch_state_filepath
-      |> OpamParser.FullPos.to_opamfile
+    let switch_state_filepath =
+      Switch.path opam switch |> Opam_path.switch_state |> Path.to_string
     in
-    file_contents
+    let+ file_res =
+      Promise.Result.from_catch (Fs.readFile switch_state_filepath)
```
rgrinberg, line 136: I'm surprised we don't need a thunk to wrap the read operation here. Can't `Fs.readFile` raise immediately as well?

src/opam.ml
```diff
@@ -127,14 +129,30 @@ end = struct
...
+    in
+    match file_res with
+    | Error e ->
+      log "Error reading switch-state file at %s. Error: %s"
+        switch_state_filepath
+        (Promise.error_to_js e |> Ojs.string_of_js);
```
mnxn, line 142: Does `Fs.readFile` throw strings as exceptions on the JS side?
I think it is more appropriate to read the message with `JsError.message`.

ulugbekna: That's a good point. But then 

1. Did you mean `Js_of_ocaml.Js_error.message`? 
2. How does one go from `Promise.error` to `Js_error.t`?

ulugbekna: In case you're interested: based on this typescript [intf](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7f7fef74f8c7116816392e6ba1d873203babb7cc/types/mz/fs.d.ts#L1132-L1136), the exception thrown should be [NodeJs.ErrornoException](http://definitelytyped.org/docs/pg--pg/interfaces/nodejs.errnoexception.html) 

I wonder if having a safe wrapper over such functions would prove useful, efficient, and maintainable, eg `readFile: path:string -> (string, NodeJs.ErrornoException.t) result Promise.t`

src/opam.ml
ulugbekna: To my best knowledge, raising an exception would violate nodejs-way of working with async code. I think the user of this function can safely expect the error in `err` argument in the callback: 

```
export function readFile(
    path: PathLike | number,
    options: { encoding: string; flag?: string | undefined } | string,
    callback: (err: NodeJS.ErrnoException | null, data: string) => void
): void;
```

When it comes to `Promise.Result.from_catch` and its behavior, I think it does what it should since `(<function call returning promise>).catch(...)` doesn't catch the exception if the function call throws one outside of promise, e.g.,

```js
function delayedPrint() { 
  throw new Error("this is thrown and not caught by `.catch`")
  return new Promise(resolve => 
         setTimeout(() => { console.log("not printed"); resolve() }, 100));
}

delayedPrint().catch(e => console.log("caught e: " + e))
```

src/opam.ml
mnxn: > 1. Did you mean `Js_of_ocaml.Js_error.message`?
> 
> 2. How does one go from `Promise.error` to `Js_error.t`?

No, I mean `Node.JsError.message` from our own bindings. `JsError.t` is `Promise.error`:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/0fc1e0d59c7d406c410d283146950c9a543bf4bb/src-bindings/node/node.mli#L41-L45

We already use this function to get the message in other places:
https://github.com/ocamllabs/vscode-ocaml-platform/blob/0fc1e0d59c7d406c410d283146950c9a543bf4bb/src/extension_instance.ml#L234-L236

> I wonder if having a safe wrapper over such functions would prove useful, efficient, and maintainable

It could be useful in the future, but I don't think it's necessary here since we only need the message for logging purposes.

ulugbekna: Thanks!

PR #943, opened 2022-04-26
Better repl evaluation system

OscarLahaie:
This pr is intended to be a part of the work of @ulugbekna in #677.
It would be interesting to have an evaluation system similar to Tuareg in Emacs. For this it is not only necessary to evaluate the user's selection but also when there is no selection the expression corresponding to the cursor position.
-
The objectives :
- [x]  Evaluate a file 
- [x] Preserve focus on editor after evaluations
- [x]  Evaluate expressions
~~Jump to the next expression after evaluation~~


OscarLahaie:
New objective : 

- [ ] Jump to the next expression after evaluation

OscarLahaie:
Hi, @ulugbekna I have two questions :
- How can I properly integrate `;;` as a part of the expression ? Because if I want to reproduce the Tuareg behavior i need to access to the expression from its end so after `;;`.
- How can I access to the next expression if there is one ?

ulugbekna:
Hi! 

> How can I properly integrate ;; as a part of the expression ? Because if I want to reproduce the Tuareg behavior i need to access to the expression from its end so after ;;.

I don't understand the question. Could you provide more context? The current REPL integration simply appends `;;` to the expression if it already doesn't have it at the end, i.e., if the last two chars of the string (expression) aren't `;;`

> How can I access to the next expression if there is one ?

I don't understand the question again, I'm afraid. Do you want to split a given string by `;;` sub-strings?

OscarLahaie:
> I don't understand the question. Could you provide more context? The current REPL integration simply appends `;;` to the expression if it already doesn't have it at the end, i.e., if the last two chars of the string (expression) aren't `;;`

Sorry i was not clear at all. When you request to ocaml-lsp in which expression you are it don't take in count the `;;` so when you evaluate there is a problem.

## Ex :
pr behavior when i press <kbd>Shift</kbd> + <kbd>Enter</kbd>:
image
The Ast node section is taking a weird code part

wanted behavior :
image
with the cursor at the end

> I don't understand the question again, I'm afraid. Do you want to split a given string by `;;` sub-strings?

In tuareg when you evaluate an expression the cursor jump to the next expression and I want to do the same. So I need to access to the end of the next expression and put the cursor at this position

ulugbekna:
> The Ast node section is taking a weird code part

The extension should be sending a custom request to ocaml-lsp to get a range in the document corresponding to the current definition. You can view all requests and responses happening between ocaml-lsp and the extension in the output pane -- invoke a command "Show OCaml Language Server Output" 

> In tuareg when you evaluate an expression the cursor jump to the next expression and I want to do the same. So I need to access to the end of the next expression and put the cursor at this position

Oh, I see. Honestly, I'm not sure about whether we would like this for two reasons: 

1. As far as I understand, if you want to jump to the next definition without some string manipulation and other workarounds, ie you would like to find the next definition from the AST, you need to create a new custom command in ocaml-lsp. I think that this is too much work for small win because one can simply "search" the next let and jump to it, or if you use vim, you can jump to a certain line or `let` easily. 
2. There are a lot of things to work on in the extension, so this small feature may not be worth adding code both here and in ocaml-lsp. 

I can give pointers if you still want to work on this, but it seems like a small win.

ulugbekna:
Also, I'd suggest to format source files that you edit before committing them. You can do that by either enabling "Format on save" in vscode settings, or running `make fmt` in console, or invoking vscode command "Format document" from command palette.

OscarLahaie:
> Also, I'd suggest to format source files that you edit before committing them. You can do that by either enabling "Format on save" in vscode settings, or running `make fmt` in console, or invoking vscode command "Format document" from command palette.

Yeah, I often forget to apply formatting 😕

OscarLahaie:
> I can give pointers if you still want to work on this, but it seems like a small win.

I think you are right. It will be enough for now I think !

ulugbekna (changes requested):
could you leave only relevant commits in this PR? it's hard to review, otherwise

also, it's better to rebase on master rather than merge

in future, I recommend not basing one PR on another unrelated PR, which is the case currently, I think with indentation changes

OscarLahaie:
I removed part from the indentation branch but it is still not clear. 
Should I start a new branch with a new pr for a cleaner commits ?

OscarLahaie:
> Hm, I think you could just drop undesired commits from this branch.
> 
> If it’s easier for you to create a new PR, feel free to do so

How can I drop them ?

ulugbekna:
I usually use interactive rebase for that. (git rebase <commit until which
you want to look through> -i)

I hope it works with merge commits as well.

You could also use git cherry-pick to pick specific commits you’d like

PR #947, opened 2022-04-27
Require ocaml >= 413 to compile

mlasson:
# Description

Since commit 93e88b7799830dc4711cf6b61957715460eac1fa support for let*/let+ punning introduced in ocaml 4.13 is required to compile the repository. 

Here I propose to bump ocaml version lower bound but an alternative fix could be to reformat for 4.11. 
Feel free to merge or close this PR.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -30,7 +30,7 @@
    (= :version))
   (ocaml
    (and
-    (>= 4.11)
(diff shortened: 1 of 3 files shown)


mlasson:
Thanks !

PR #951, opened 2022-05-11
More functionnalities for repl

OscarLahaie:
This pr is intended to be a part of the work of @ulugbekna in #677.
It would be interesting to have an evaluation system similar to Tuareg in Emacs. For this it is not only necessary to evaluate the user's selection but also when there is no selection the expression corresponding to the cursor position.
-
The objectives :
- [x]  Evaluate a file 
- [x] Preserve focus on editor after evaluations
- [x]  Evaluate expressions
- [x] Option to jump to the next expression


OscarLahaie:
This is the cleaner version of pull request #943.

src/repl.ml
```diff
@@ -129,23 +129,27 @@ let create_terminal instance sandbox =
...
+  text_editor |> Vscode.TextEditor.document |> Vscode.TextDocument.uri
+  |> Vscode.Uri.path
+
+let preformat_code code =
+  let is_repl_ready s = String.is_suffix s ~suffix:";;" in
+  let trimmed_code = String.strip code in
+  if is_repl_ready code then
+    (* newline is for nicer look in REPL *)
+    "\n" ^ trimmed_code
+  else Printf.sprintf "\n%s\n;;" trimmed_code
```
mnxn, line 152: This formatting sometimes looks worse like in this example with a single expression:

```
# 
  x        
  ;;       
- : int = 1
```

Maybe it's worth checking if the code is a single line first? If it is, then there's no need to add any extra formatting.

I would prefer the following in simple cases:
```
# x ;;       
- : int = 1
```

mnxn (changes requested):
A few changes need to be made for Windows and the official OCaml REPL.

I haven't had a chance to try macOS or utop yet.

src/repl.ml
```diff
@@ -174,15 +226,14 @@ module Command = struct
         match term with
         | Error err -> show_message `Error "Could not start the REPL: %s" err
         | Ok term ->
-          let code = get_code textEditor in
-          if String.length code > 0 then
-            let code = prepare_code code in
-            Terminal_sandbox.send term code
+          let uri = get_uri textEditor in
+          let use = "#use \"" ^ uri ^ "\";;" in
```
mnxn, line 230: `get_uri` will return paths with backslashes like `c:\a\b\c` when it uses `Uri.fsPath` on Windows. [Quoted string literals](https://v2.ocaml.org/manual/lex.html#sss:stringliterals) are more appropriate in this situation.

```suggestion
          let use = "#use {|" ^ uri ^ "|};;" in
```

src/repl.ml
```diff
@@ -129,23 +129,27 @@ let create_terminal instance sandbox =
...
+    let selected_text =
+      TextDocument.getText document ~range:(selection :> Range.t) ()
+    in
+    Some selected_text
 
-let prepare_code code =
-  if String.is_suffix code ~suffix:";;" then code else code ^ ";;"
+let get_uri text_editor =
+  text_editor |> Vscode.TextEditor.document |> Vscode.TextDocument.uri
+  |> Vscode.Uri.path
```
mnxn, line 144: `Uri.path` returns paths like `/c:/a/b/c` on Windows, which are invalid for `#use` in the REPL. This should be `Uri.fsPath` instead.
```suggestion
  |> Vscode.Uri.fsPath
```

mnxn (changes requested):

src/repl.ml
OscarLahaie: Done!
I have done some tests and it seems to work well with the new version of the function.
I am not sure about the `List.mem` efficiency, what do you think about this solution?

src/repl.ml
OscarLahaie: I have not tried it on MacOs and I can't. However, with your suggestions and the new patch, it works on Windows and Linux on `Utop` and `Ocaml`.

src/repl.ml
mnxn: The new behavior is great.

`String.mem` is fine but I'm not sure what the difference is from `String.contains` that we use in other parts of the extension.

OscarLahaie:
I made several tests to see if the jump is working but I am sure some cases unknown to me.
So if you find any bugs tell me so I can fix them.

mnxn (approved):

rgrinberg:
@mnxn is this good to merge?

ulugbekna:
I will have a look at this today

mnxn:
> is this good to merge?

I looked good to me when I tried it.

Ready to merge unless @ulugbekna finds something.

src/repl.ml
```diff
@@ -165,7 +183,188 @@ module Command = struct
     in
     Extension_commands.register ~id:Extension_consts.Commands.open_repl handler
 
-  let _evaluate_selection =
+  let _evaluate_expression =
```
ulugbekna, line 186: I'm a little skeptical about having a 180-line function

ulugbekna (changes requested):
This needs a changelog entry. 

Didn't look too closely at the 180-line function, but functionality added by the PR seems to work well

ulugbekna:
I think this can me squashed and merged, but jumping to next expression implementation needs to be made on the ocamllsp side as a new custom request so that we avoid hacks that we have now.

OscarLahaie:
> I think this can me squashed and merged, but jumping to next expression implementation needs to be made on the ocamllsp side as a new custom request so that we avoid hacks that we have now.
> 
> Also, one should be careful using `registerTextEditorCommand` because it's supposed to be run synchronously, meaning starting off a promise as we do it now is unsafe. See [microsoft/vscode#16814](https://github.com/microsoft/vscode/issues/16814)

I don't have much time this week. Nevertheless I will look at and improve the code in the coming weeks if necessary. Because there are some behaviours to be fixed at the ocaml-lsp level.

src/repl.ml
```diff
@@ -129,23 +129,40 @@ let create_terminal instance sandbox =
...
-    Vscode.Selection.start selection |> Vscode.Position.character
-  in
-  let end_char = Vscode.Selection.end_ selection |> Vscode.Position.character in
-  let document = Vscode.TextEditor.document text_editor in
-  if start_line = end_line && start_char = end_char then
-    let line = Vscode.TextDocument.lineAt document ~line:start_line in
-    Vscode.TextLine.text line
+let get_selected_code text_editor =
+  let selection = TextEditor.selection text_editor in
+  let document = TextEditor.document text_editor in
```
rgrinberg, line 134: Why do you obtain `document` right away but only use it in one branch of the conditional?

src/repl.ml
```diff
@@ -129,23 +129,40 @@ let create_terminal instance sandbox =
...
+  text_editor |> Vscode.TextEditor.document |> Vscode.TextDocument.uri
+  |> Vscode.Uri.fsPath
 
-let prepare_code code =
-  if String.is_suffix code ~suffix:";;" then code else code ^ ";;"
+let preformat_code code =
+  let is_repl_ready s = String.is_suffix s ~suffix:";;" in
+  let trimmed_code = String.strip code in
+  if is_repl_ready trimmed_code then
+    if String.mem trimmed_code '\n' then
```
rgrinberg, line 150: As far as I can tell `String.mem trimmed_code '\n'` is always evaluated. Therefore it should be assigned a binding and moved before the conditionals.

README.md
```diff
@@ -259,7 +260,12 @@ prefix `OCaml:`:
 | `ocaml.current-dune-file`    | Open Dune File (located in the same folder) |                    |
 | `ocaml.switch-impl-intf`     | Switch implementation/interface             | `Alt+O`            |
 | `ocaml.open-repl`            | Open REPL                                   |                    |
-| `ocaml.evaluate-selection`   | Evaluate Selection                          | `Shift+Enter`      |
+| `ocaml.evaluate-expression`  | Evaluate Expression in REPL                 | `Shift+Enter`      |
+| `ocaml.evaluate-file`        | Evaluate Selection                          | `Shift+Alt+Enter`  |
+
+Note: `ocaml.evaluate-expression` was previously named
+`ocaml.evaluate-selection`. You can also disabe the new cursor jumps after this
```
rgrinberg, line 267: Why don't we preserve `ocaml.evaluate-selection`? I don't see anything wrong with having both commands.

rgrinberg (changes requested):

rgrinberg:
Before merging this, I think we need to get on the same page of what good REPL functionality looks like. I have no interest in copying tuareg and its limitations. We can do a lot better if we think about things a bit more.

README.md
rgrinberg: I understood. I just don't see why we should remove a command with a clear name and semantics and replace it with a command that has neither of those things. Seems like we're just irritating users who are used to the old command this way.

ulugbekna:
> Before merging this, I think we need to get on the same page of what good REPL functionality looks like. I have no interest in copying tuareg and its limitations. We can do a lot better if we think about things a bit more.

Do you have precise limitations you'd like gone before merging? 

Things that are not addressed in this PR but can be done in other, IMO: 

- configuration for how module is evaluated (#use or #mod_use) as a dialog 
- configurable REPL instance -- the best decision is to have two commands for picking a terminal to send commands to:
  - one command to pick an integrated terminal that already exists from a list (this allows users to jump into existing repl session and send commands to it, for example) 
  - one command to launch a terminal from settings-defined repl binary and args OR default utop - this already exists but it uses a timeout before sending commands to a new REPL -- this needs to be fixed

README.md
ulugbekna: Good point, but then I would show a deprecation message saying to use the new command because 

1. The new command has almost the same semantics - select and evaluate (no selection resulted in evaluating the whole line but I'm not sure that's very useful and the new command does the same thing if it's a single-line let binding) 
2. There's then contention for the shortcut: keep the old and miss out on a better command or replace and still irritate

rgrinberg:
First of all, as a general comment, there's just too many knobs, settings, commands that all make the experience way too complicated. I've already went through this stuff with tuareg mode. In the end, their toplevel integration is hardly usable.

99% of the time when I use the toplevel, I do the following:

1. Open a project
2. Navigate to a module I want to experiment with
3. Figure out how to load into the toplevel.

We should focus on this core feature first before we worry about directives, selecting repl instances. Finally, we should absolutely avoid any code that tries to parse OCaml. I see this PR is adding quite a bit of that and it's really wasted effort.

Here's a simple idea for how to make the toplevel more useful.

* We add a command `$ dune ocaml top-module path/to/module.ml --source=/tmp/selection/file`. This command will either load the toplevel binary or print the directives required to load `module.ml` as if its source was `--source`.

* We add a request to lsp that given a position in the document, will return a completed & pretty printed parse tree. Luckily for us, merlin already knows how to do this.

* On the vscode side, we add a command to launch a toplevel where the cursor is. It will obtain the source using lsp, then run the dune command mentioned above. There's no need to select toplevel instances. We associate every module with a toplevel instance. If there's already a toplevel for a module, then we kill it and launch it again.

Let's forget about all the other stuff for now.

src/repl.ml
```diff
@@ -165,7 +182,194 @@ module Command = struct
...
+                         ~end_:selection_start)
+                    ~revealType:TextEditorRevealType.AtTop ());
+                Promise.return ())
+              else
+                let+ range =
+                  Custom_requests.Wrapping_ast_node.send_request client ~doc:uri
+                    ~position:correct_position
+                in
+                match range with
+                | None -> ()
```
rgrinberg, line 308: When can this request return None?

smorimoto:
Close this for now, but feel free to re-open.

PR #957, opened 2022-05-26
Fix `useUtop`

OscarLahaie:
I think that this wrong name makes the option unavailable.


rgrinberg:
@mnxn would you mind amending the CHANGES?

PR #958, opened 2022-05-28
Fix #183 idea

### LINKED ISSUE #183: OCaml keywords are not suggested
I suggest OCaml keywords be included in suggestions.
This is mostly an issue when you type `in` and hit Return, as most of the time another value beginning by `in` will be inserted instead, and this is annoying

OscarLahaie:
Why don't we add `in` and `do` to snippets to facilitate their insertion ? With this, words like `do_at_exit` and `in_channel_length` would not be given priority. I think vscode's behaviour is better with these snippets.

### DIFF
--- a/snippets/ocaml.json
+++ b/snippets/ocaml.json
@@ -50,5 +50,13 @@
   "type declaration": {
     "prefix": "type",
     "body": ["type ${1:name} = $0"]
+  },
+  "in": {
+    "prefix": "in",
+    "body": ["in"]
+  },
+  "do": {
+    "prefix": "do",
+    "body": ["do"]
   }
 }


rgrinberg:
What happens when I write:
```
let inner = 123 in
10 + in|
```

Where the cursor is at `|`. Does vscode suggest to insert a snippet here?

OscarLahaie:
> Where the cursor is at |. Does vscode suggest to insert a snippet here?

Yes, it does.
image
Is that an issue?

rgrinberg:
Yes, it's essentially the same problem under different circumstances.

OscarLahaie:
> Yes, it's essentially the same problem under different circumstances.

Yes, but I think this is preferable as I think this case is not annoying. Where the non-suggestion of the `in` is a real lack of comfort.

rgrinberg:
If you can convince the other maintainers it's a good idea then I won't block this.

ulugbekna:
I'm in favor

mnxn:
I don't think this is the correct solution.

The lack of keyword suggestions affects all editors that use ocaml-lsp, so it shouldn't only be fixed in this editor. 
This also only handles two keywords, but there are other keywords that are not suggested. Users can have the same issues with other keywords if they `open` any additional modules that expose similar identifiers.

I don't mind if keywords are suggested when an identifier is expected, but I would prefer a solution in ocaml-lsp or merlin so we don't need to maintain a list of keywords just for VS Code.

OscarLahaie:
> I don't think this is the correct solution.
> 
> The lack of keyword suggestions affects all editors that use ocaml-lsp, so it shouldn't only be fixed in this editor. This also only handles two keywords, but there are other keywords that are not suggested. Users can have the same issues with other keywords if they `open` any additional modules that expose similar identifiers.
> 
> I don't mind if keywords are suggested when an identifier is expected, but I would prefer a solution in ocaml-lsp or merlin so we don't need to maintain a list of keywords just for VS Code.

I totally agree with you but I don't think it will be supported on `ocaml-lsp` again. Also maintaining a word list won't be that complicated as the implementation is quite minimalist. Finally for the case of imported keywords it is clear that it should be supported by `ocaml-lsp` but in the meantime these few keywords are quite handy to be supported in this way.

3Rafal:
This solution was already discussed in #1088 . IMO, it would be the best to solve keyword completion in merlin/ocaml-lsp. 
The snippet workaround could be added to readme though, so that users can decide for themselves.

bzy-debug:
ocaml-lsp has already supported the completion of `in` keyword, see https://github.com/ocaml/ocaml-lsp/pull/1217

Issue #733, opened 2021-10-08
Make the extension compatible with GitHub Codespace


tmattio:
See: https://code.visualstudio.com/docs/remote/codespaces#_why-is-an-extension-not-installable-in-the-browser

This would require making the features that depend on the system (lsp, filesystem, etc.) optional, to essentially provide only the syntax.

cc @TheLortex

PR #962, opened 2022-06-09
Minimal support for web

TheLortex:
This is the ground work for https://github.com/ocamllabs/vscode-ocaml-platform/issues/733

Basically the only requirement to have the extension listed as web-compatible is to add a `"browser": ...` field in `package.json`. 
Obviously the script in `"main": "./dist/vscode_ocaml_platform.bc.js"` is not web-compatible as it uses various nodejs apis. 

This PR adds the plumbing to generate a `vscode_ocaml_platform_web.bc.js` script that would act as the web version of the extension. Currently it's empty as no commands are implemented. Instead they are all disabled in the manifest file using `"enablement": "!virtualWorkspace"`.

So what remains working is syntax highlighting, which is a nice thing to have when inspecting code from the web (using github.dev for example).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ node_modules/
 # VSCode build artifacts
 dist/
 .vscode-test/
+.vscode-test-web/
 *.vsix
 
 # DS_Store
--- a/package.json
+++ b/package.json
@@ -14,10 +14,10 @@
   },
   "homepage": "https://github.com/ocamllabs/vscode-ocaml-platform",
   "capabilities": {
-      "virtualWorkspaces": {
-          "supported": "limited",
-          "description": "Only syntax highlighting is provided."
-      }
+    "virtualWorkspaces": {
+      "supported": "limited",
+      "description": "Only syntax highlighting is provided."
+    }
   },
   "main": "./dist/vscode_ocaml_platform.bc.js",
   "browser": "./dist/vscode_ocaml_platform_web.bc.js",
@@ -1068,7 +1068,7 @@
     "deploy:ovsx": "ovsx publish --yarn",
     "fmt:check": "prettier . --check",
     "fmt": "prettier . --write",
-    "web": "npx @vscode/test-web --extensionDevelopmentPath=. ."
+    "run:web": "npx @vscode/test-web --extensionDevelopmentPath=. ."
   },
   "dependencies": {
     "polka": "^1.0.0-next.22",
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -1025,7 +1067,8 @@
     "deploy:vsce": "vsce publish --packagePath ocaml-platform.vsix --yarn",
     "deploy:ovsx": "ovsx publish --yarn",
     "fmt:check": "prettier . --check",
-    "fmt": "prettier . --write"
+    "fmt": "prettier . --write",
+    "web": "npx @vscode/test-web --extensionDevelopmentPath=. ."
```
tmattio, line 1071: Should this be named `test:web` if it's running tests?

src/dune
```diff
@@ -1,5 +1,6 @@
 (executable
  (name vscode_ocaml_platform)
+ (modules :standard \ vscode_ocaml_platform_web)
```
tmattio, line 3: I'm expecting that when we want to add features to the web extension, this won't reuse anything in the existing code base. I'd maybe suggest we put this in a separate directory (e.g. `src-browser`) to avoid any confusion.

.gitignore
```diff
@@ -23,6 +23,7 @@ node_modules/
 # VSCode build artifacts
 dist/
 .vscode-test/
+.vscode-test-web/
```
tmattio, line 26: Is this generated when running `npx @vscode/test-web`?

tmattio (approved):
This is great, thank you for doing this @TheLortex!

The changes shouldn't impact anything on the "native" extension. I'm leaving a few minor suggestions, but overall I'm happy to merge.

tmattio:
Also CI complaining about the formatting, can you run prettier?

package.json
TheLortex: It's not running tests, just launching vscode in a web browser with the web extension. But yes I used this to test the extension so it could be prefixed by "test:"

.gitignore
TheLortex: Yes!

src/dune
TheLortex: I agree

package.json
TheLortex: Thinking about it, it is not a good fit because `"test": "npm-run-all -s test:*",` would then run it even if it's not an automated test target. 

I have added a commit to change the name to `run:web` but feel free to suggest something else.

tmattio:
Thanks @TheLortex!

PR #964, opened 2022-06-11
fix #961: reveal ast node can now open ast explorer

### LINKED ISSUE #961: "Reveal AST Node" action should open AST explorer if it doesn't exist
## Expected behavior: 

"Reveal AST Node" action should open AST explorer for the file and reveal the AST node. 

## Current behavior: 

"Reveal AST Node" action fails saying that the AST explorer is not open.

ulugbekna:
reveal ast node command can now open an ast explorer and reveal the node in the newly opened explorer

fixes #961

### DIFF
--- a/src/ast_editor.ml
+++ b/src/ast_editor.ml
@@ -419,12 +419,13 @@ let open_both_ppx_ast instance ~document =
   | Error e -> raise (User_error e)
 
 module Command = struct
+  let open_ast_explorer_handler textEditor =
+    let uri = TextEditor.document textEditor |> TextDocument.uri in
+    open_ast_explorer ~uri
+
   let _open_ast_explorer_to_the_side =
     let handler _ ~textEditor ~edit:_ ~args:_ =
-      let (_ : unit Promise.t) =
-        let uri = TextEditor.document textEditor |> TextDocument.uri in
-        open_ast_explorer ~uri
-      in
+      let (_ : unit Promise.t) = open_ast_explorer_handler textEditor in
       ()
     in
     let handler e ~textEditor ~edit ~args =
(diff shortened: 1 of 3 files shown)


src/ast_editor.ml
```diff
@@ -436,24 +436,42 @@ module Command = struct
...
+          match
+            Ast_editor_state.find_webview_by_doc ast_editor_state
+              (TextDocument.uri document)
+          with
+          | Some _ as r -> Promise.return r
+          | None ->
+            let* (_ : Ojs.t option) =
+              Commands.executeCommand
+                ~command:Extension_consts.Commands.open_ast_explorer_to_the_side
+                ~args:[]
```
rgrinberg, line 452: Instead of relying on the command, we should call the underlying function.

src/ast_editor.ml
```diff
@@ -436,24 +436,42 @@ module Command = struct
...
+            in
+            (* without this small timeout, sometimes the command only opens an
+               AST explorer but doesn't reveal the node. *)
+            let+ () = Node.setTimeout' ~wait_ms:200 in
```
rgrinberg, line 456: While it's understandable why you've done it this way, I don't think we can use this. What's going to guarantee that the node will be revealed in 200ms for big documents, slower laptops, etc?

We need to think of a different of a different approach.  I'm looking at this `send_msg` function:

```
let send_msg t value ~(webview : WebView.t) =
  let msg = Ojs.empty_obj () in
  Ojs.set_prop_ascii msg "type" (Ojs.string_to_js t);
  Ojs.set_prop_ascii msg "value" value;
  let (_ : bool Promise.t) = WebView.postMessage webview msg in
  ()
```

perhaps we wait for the promise it returns?

ulugbekna: If we want to be on the safe side, I think we need to refactor `vscode.commands.registerCommand`'s `callback` to return `Ojs.t` rather than `unit` as it is the case currently. Then we register "open ast explorer" command not as a text editor command but an ordinary command, then we'll be able to wait for the command to complete before revealing a node.

ulugbekna: references: 
- https://code.visualstudio.com/api/references/vscode-api#commands
- https://github.com/microsoft/vscode/issues/16814

rgrinberg: Sounds good, but why do we need to go through the command api at all? It's untyped, error handling is not clear, and is meant to be user facing. Let's do things directly when possible.

ulugbekna: You're absolutely right. Pushed a fixed version. Seems to work. Tested on `lsp/types.ml` a couple of times

src/ast_editor.ml
```diff
@@ -436,24 +437,34 @@ module Command = struct
...
+            Ast_editor_state.find_webview_by_doc ast_editor_state
+              (TextDocument.uri document)
+        in
+        let offset =
+          let selection = Vscode.TextEditor.selection textEditor in
+          let position = Vscode.Selection.start selection in
+          TextDocument.offsetAt document ~position
+        in
+        match webview with
+        | Some webview -> send_msg "focus" (Ojs.int_to_js offset) ~webview
```
rgrinberg, line 460: In a subsequent PR, consider wrapping all the `"focus"` calls into a function. It's best to wrap these junky untyped apis as early as possible and stick to our typed wrappers in the rest of the codebase.

rgrinberg (approved):
Can you add a CHANGES entry?

src/ast_editor.ml
ulugbekna: Yes, I have a branch actually with some ast explorer cleanups. Will send a PR at some point

PR #980, opened 2022-07-02
chore: use standard github actions

rgrinberg:
ps-id: 86ef1580-bf2a-4b7e-9aab-574998f46397


smorimoto:
This will be taken care of in #1116

PR #983, opened 2022-07-05
switchImplIntf: don't skip uri encoding

tatchi:
I noticed while looking at that issue https://github.com/ocaml/ocaml-lsp/issues/184#issuecomment-708241481 that the URI is not encoded when we send the `switchImplIntf ` request.

Example with the `switchImplIntf` request:

```
[Trace - 20:36:18] Sending request 'ocamllsp/switchImplIntf - (12)'.
Params: [
    "file:///c:/OCaml64/home/Corentin/ocaml-lsp/lsp/src/coco-test.mli"
]
```

Same file but with another request:

```
[Trace - 20:36:18] Sending request 'textDocument/codeAction - (13)'.
Params: {
    "textDocument": {
        "uri": "file:///c%3A/OCaml64/home/Corentin/ocaml-lsp/lsp/src/coco-test.ml"
    },
```

I believe that URIs should always be encoded. I'm not sure why the encoding was explicitly skipped in that case.


rgrinberg:
It looks right, but doesn't this break when used with old servers?

rgrinberg:
ping @tatchi

tatchi:
I still think that the fixes is correct, but I have no idea if it can break old servers 🤔

rgrinberg:
Looking at git blame, it seems like @ulugbekna introduced the `~skipEncoding:true`. Perhaps he could clarify and/or review this PR?

ulugbekna:
I just didn't know the URIs need to be encoded at the time, sorry :-(

tatchi:
> Okay, then this change is fine then. @tatchi do you mind adding a CHANGES entry?

Done 👍

PR #991, opened 2022-07-27
Fix syntax highlighting for let bindings

dlesbre:
Currently `let x : t` colors correctly but the spaceless version `let x:t` does not color `x`.
This PR fixes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+- Fix syntax highlighting for let bindings with type annotations (#991)
+
 ## 1.10.8
 
 - Add more known versions of ocamllsp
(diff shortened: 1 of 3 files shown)


mnxn:
Thanks for your contribution! I pushed a small change but the behavior should be the same.

Before we merge this, I noticed that `and x:t` still has the old highlighting. Would you mind taking a look at it? It's defined in `ocaml.interface.json`. 

The regex rule is a little more complicated but it should be the same type of modification.
https://github.com/ocamllabs/vscode-ocaml-platform/blob/bd4156237b7cba2bc5fcf21aa7b7326702de642b/syntaxes/ocaml.interface.json#L41-L42

dlesbre:
I've added the same fix to and. It seems to work just as well with the testing I did.

mnxn:
Thanks!

PR #1037, opened 2022-10-26
cleanup: remove unused `Import.Ocaml_version.Releases` because ocaml v4.14.0 is already supported by `Ocaml_version.Releases`

ulugbekna:


rgrinberg:
I think this needs a lower bound on the version library?

ulugbekna:
Indeed! Fixed.

PR #1038, opened 2022-10-27
remove `lint-opam` CI check

ulugbekna:
This check always fails because we can't install `opam-dune-lint` because it conflicts with our deps.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,44 +113,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
 
-  lint-opam:
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


rgrinberg:
Do we even need to install our dependencies to run lint?

ulugbekna:
Oh good question. 

AFAIU, this tool makes sure that all dune deps have corresponding opam packages. I don't think we add/remove too many deps in this project, so we don't necessarily need it? Wouldn't CI that simply compiles the project detect it given a dependency would be missing anyway? 

It also doesn't support dune 3: https://github.com/ocurrent/opam-dune-lint/issues/45

So all in all, I don't think we need this

rgrinberg (approved):

PR #1042, opened 2022-11-05
add setting to pass extra args to ocamllsp

### LINKED ISSUE #1036: Setting to pass extra arguments to ocamllsp
Something similar to `ocaml.server.extraEnv` to pass CLI arguments to ocamllsp. Useful at least for passing `--fallback-read-dot-merlin`.

tatchi:
Close #1036

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 - Fix AST Explorer not displaying anything with error "Blocked script execution"
   (#1050)
 
+- Add `ocaml.server.args` configuration option to pass extra arguments to the
+  language server, i.e. ocamllsp (#1042)
+
 ## 1.11.0
 
 - Fix syntax highlighting for let bindings with type annotations (#991)
--- a/package.json
+++ b/package.json
@@ -466,6 +466,12 @@
           "default": null,
           "markdownDescription": "Extra environment variables that will be passed to OCaml LSP executable. Useful for debugging purposes mostly."
         },
+        "ocaml.server.args": {
+          "type": "array",
+          "items": "string",
+          "default": [],
+          "markdownDescription": "Extra arguments to pass to ocamllsp."
+        },
         "ocaml.dune.autoDetect": {
           "type": "boolean",
           "default": true,
--- a/src/extension_instance.ml
+++ b/src/extension_instance.ml
@@ -53,7 +53,8 @@ end = struct
       ()
 
   let server_options sandbox =
-    let command = Sandbox.get_command sandbox "ocamllsp" [] in
+    let args = Settings.(get server_args_setting) |> Option.value ~default:[] in
+    let command = Sandbox.get_command sandbox "ocamllsp" args in
     Cmd.log command;
     let env =
       let extra_env_vars =
--- a/src/settings.ml
+++ b/src/settings.ml
@@ -106,3 +106,10 @@ module ExtraEnv = struct
 end
 
 let server_extraEnv = ExtraEnv.get
+
+let server_args_setting =
+  create_setting
+    ~scope:ConfigurationTarget.Workspace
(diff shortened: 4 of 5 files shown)


ulugbekna:
Hi! I’m inclined to believe that the setting should be a string list rather than a string, what do you think?

tatchi:
> Hi! I’m inclined to believe that the setting should be a string list rather than a string, what do you think?

I think it makes sense yes. I'll make the changes :)

ulugbekna (approved):
Seems correct. 

Have you tested passing something to olsp by any chance?

Also, the spec in `package.json` says the setting is `string[] | null` but we parse as `string[]`. This is fine because `Settings.get` catches decoding errors but we could make the spec have `[]` as the default and avoid `| null`

tatchi:
> Also, the spec in `package.json` says the setting is `string[] | null` but we parse as `string[]`. This is fine because `Settings.get` catches decoding errors but we could make the spec have an empty array as the default and avoid `| null`

Done

> Have you tested passing something to olsp by any chance?
 
Yep :) But I noticed that no notifications are shown (or just flashing) when passing a wrong argument. That's not ideal.



https://user-images.githubusercontent.com/5595092/203837006-8b8428c3-073d-4347-93fa-e980fcbb8b27.mov

I was able to reproduce the issue using a [smaller example](https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-sample) and found out that the issue doesn't happen with the previous major version of `vscode-languageclient`, which is version 7 (we use version 8)

tatchi:
It seems to work better if we slightly override the default `errorHandler.close` not to restart the server when the client's state is `starting`:

```diff
errorHandler: {
	closed() {
+		if (client.state === State.Starting)
+			return { action: CloseAction.DoNotRestart };
		restarts.push(Date.now());
		if (restarts.length <= maxRestartCount) {
			return { action: CloseAction.Restart };
		} else {
			const diff = restarts[restarts.length - 1] - restarts[0];
			if (diff <= 3 * 60 * 1000) {
				return {
					action: CloseAction.DoNotRestart,
					message: `The example server crashed ${
						maxRestartCount + 1
					} times in the last 3 minutes. The server will not be restarted. See the output for more information.`,
				};
			} else {
				restarts.shift();
				return { action: CloseAction.Restart };
			}
		}
	},
},
```

In that case we should get a `Server initialization failed` error.

https://user-images.githubusercontent.com/5595092/203848223-55641819-b07d-4503-b68a-7ceffa1ed40b.mov

ulugbekna:
Thanks for good investigation and patience (for my delayed replies)! :-)

I believe it's best if we don't fail server initialization if the passed arguments are not correct. I think it would be better if we launch without user's arguments if ocamllsp failed to launch because of unknown options, and show a message saying that "ocamllsp doesn't recognize argument X" (not necessarily this wording)

rgrinberg:
On the other hand, I think it's better to fail. If the user does not notice his flags aren't being respected, the server is unlikely to provide the correct behavior. So let's fail early instead of possibly creating confusion down the line.

ulugbekna:
Fine with me

tatchi:
Fine with me too. But it means that if we want to get the same behaviour as in the second video (i.e get an initialization failed error), we'll have to replicate the VScode's [default error handler](https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts#L378) with theses 2 extra lines.

Not ideal IMO but I guess we have no other choice.

ulugbekna:
I would leave it as is.
1. There’s only one effective flag supported by olsp, and it’s intended for advanced users mostly
2. One can still infer there’s an error from the server not working afaiu
3. There’re higher priority things 
4. This should rather be fixed upstream in the vscode lib

ulugbekna (approved):

PR #1069, opened 2022-12-18
chore: add lsp version 1.15.0~5.0preview1

ulugbekna:
I guess we can make a release after merging this?


rgrinberg:
Yes, we should make a release.

PR #1074, opened 2023-01-16
chore: remove env var that's no longer used in ocaml-lsp

ulugbekna:


ulugbekna:
olsp doesn't have env var `OCAML_LSP_SERVER_LOG` anymore right? @rgrinberg

3Rafal (approved):

PR #1083, opened 2023-03-01
Reorder sandbox buttons

### LINKED ISSUE #1054: sandbox treeview UX: exchange places of "generate documentation" and "open documentation"
because the former is shown for all and the latter is only available for some. It would make more sense to have "generate documentation" button to not jump from left to right, ie "gen doc" is 2nd rightmost icon

image
image

3Rafal:
Fixes #1054 
Screenshot 2023-03-01 at 16 58 30

Ordering notation:
https://code.visualstudio.com/api/references/contribution-points#Sorting-inside-groups

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.12.3
+
+- Reorder sandbox buttons (#1083)
+
 ## 1.12.2
 
 - Add more known versions of ocamllsp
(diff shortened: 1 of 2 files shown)


tmattio (approved):
Looks good, thanks for the PR and the link to the documentation for the ordering notation, I didn't know about it!

tmattio:
Before merging, could you add an entry to the changelog?

PR #1100, opened 2023-05-17
Let prepare_code check trimmed code

### LINKED ISSUE #1099: Evaluate Selection does not strip `;;` if whitespace follows
'Evaluate Selection' strips `;;` at the end of a selected text to avoid duplicated `;;`s (e.g., to avoid running `let x = 3;; ;;`).
But, if whitespace follows `;;`, the `;;` is not stripped, causing the error to happen again.

For example, selecting `let x = 3;;` works, but `let x = 3;; ` does not work.
Specifically, this error causes the command to fail when a line break (`\n`) follows `;;`.

aqjune:
This patch resolves #1099 by using the trimmed code when checking whether the code has suffix `;;`.

### DIFF
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -148,7 +148,7 @@ let get_code text_editor =
     Vscode.TextDocument.getText document ~range:(selection :> Vscode.Range.t) ()
 
 let prepare_code code =
-  if String.is_suffix code ~suffix:";;" then code else code ^ ";;"
+  if String.is_suffix (String.trim code) ~suffix:";;" then code else code ^ ";;"
 
 module Command = struct
   let _open_repl =


rgrinberg:
Could you review this one @3Rafal?

src/repl.ml
```diff
@@ -148,7 +148,7 @@ let get_code text_editor =
     Vscode.TextDocument.getText document ~range:(selection :> Vscode.Range.t) ()
 
 let prepare_code code =
-  if String.is_suffix code ~suffix:";;" then code else code ^ ";;"
+  if String.is_suffix (String.trim code) ~suffix:";;" then code else code ^ ";;"
```
3Rafal, line 151: We use `base` as a replacement for standard lib, so the following should be used instead:
```suggestion
  if String.is_suffix (String.strip code) ~suffix:";;" then code else code ^ ";;"
```

3Rafal (changes requested):
Please update [changelog](https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/CHANGELOG.md)

3Rafal (approved):

aqjune:
Thanks!

PR #1102, opened 2023-05-25
Bump node for Nix

3Rafal:
Currently, Nix build uses Node in version 14.x. It fail with following message:
```
error: Package ‘nodejs-14.21.3’ in /nix/store/wdi2jyp0d556jhyriaksrmfx4a798nhr-source/pkgs/development/web/nodejs/v14.nix:11 is marked as insecure, refusing to evaluate.


       Known issues:
        - This NodeJS release has reached its end of life. See [https://nodejs.org/en/about/releases/.](https://nodejs.org/en/about/releases/)
```

I tried to bump it to 16.x, but it fails as well. The lowest version that works is 18.x, which is current Node LTS.
This PR is the smallest amount of work that makes Nix build go green.
In separate PR we should consider bumping Node for other workflows as well.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -2,7 +2,7 @@
 let
   strings = pkgs.lib.strings;
   args = {
-    inherit (pkgs.ocaml-ng.ocamlPackages_4_13) ocaml;
+    inherit (pkgs.ocaml-ng.ocamlPackages_4_14) ocaml;
     selection = ./opam-selection.nix;
     src =
       let
--- a/nix/opam-selection.nix
+++ b/nix/opam-selection.nix
@@ -14,8 +14,8 @@ let
         {
           owner = "ocaml";
           repo = "opam-repository";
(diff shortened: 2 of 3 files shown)


3Rafal:
@rgrinberg 
my fix resolves node version issue, but workflow fails later with:
```
dune build @runtest @all
File "dune-project", line 1, characters 11-14:
1 | (lang dune 3.0)
               ^^^
Error: Version 3.0 of the dune language is not supported.
Supported versions of this extension in version 3.0 of the dune language:
- 1.0 to 1.12
- 2.0 to 2.9
make: *** [Makefile:85: nix-tests] Error 1
Error: Process completed with exit code 2.
```

I tried to fix it, but my nix knowledge isn't deep enough (yet?). Do you have an idea about what should be done to fix this?

rgrinberg:
I think you can remove `nix/opam-selection.nix` and regenerate it with make: `make nix/opam-selection.nix`.

3Rafal:
It worked, thanks! :)

PR #1105, opened 2023-05-30
Update README.md

mseri:
ocamlformat-rpc is no longer needed since ocamlformat 0.22.4 (See https://github.com/ocaml/opam-repository/pull/23824)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,7 @@ full without all of its dependencies.
      project root directory.
 
    - When you hover the cursor over OCaml code, the extension shows you the type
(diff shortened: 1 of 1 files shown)


avsm (approved):

avsm:
The Makefile also still installs the ocamlformat-rpc package.

mseri:
Well spotted. I have updated it, there are no remaining instances of ocamlformat-rpc from what I can see. The esy lock file still requires `ocamlformat-rpc-lib` but that seems correct

PR #1116, opened 2023-06-09
update gha workflows

smorimoto:


.github/workflows/main.yml
```diff
@@ -66,6 +77,8 @@ jobs:
 
       - name: Build extension
         run: opam exec -- make build-release
+        env:
+          NODE_OPTIONS: --openssl-legacy-provider
```
smorimoto, line 81: we have to upgrade webpack related package versions to fix this...

PR #1142, opened 2023-06-20
Fix astexplorer in `make build`

sim642:
Previously `make build`, as suggested by CONTRIBUTING, fails with:
```
dune build src/vscode_ocaml_platform.bc.js
yarn --cwd astexplorer start           
yarn run v1.22.19
warning package.json: No license field
error Command "start" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [Makefile:24: build] Error 1
```


smorimoto (approved):
Good catch!

PR #650, opened 2021-06-28
OCaml earlybird debugger integration

hackwaly:

PR #1148, opened 2023-06-23
OCaml earlybird debugger

### LINKED ISSUE #534: Integrate ocamlearlybird
https://github.com/ocaml/opam-repository/pull/18198

As ocamlearlybird beta2 will soon be released. I think it's the right time to start considering integrating ocamlearlybird with ocaml-platform.

If we decide to integrate it. We have two choices:

1. Make a vscode extension pack that simply includes hackwaly/ocamlearlybird extension.
2. Add debugger contribution in ocaml-platform extension that directly using cli version of ocamlearlybird.

sim642:
Closes #534.

This is more-or-less a port of earlybird's own small VSCode extension to OCaml: https://github.com/hackwaly/ocamlearlybird/tree/271bdd877ae285c06df1d4e90bc1d11dbdf83fb5/integrations/vscode. Parts of this port are borrowed from the earlier attempt #650.

Unlike that earlier attempt, this contains no dune executable detection or anything of that sort. It simply allows the earlybird DAP server to be launched if it is installed in the opam switch. The only additional integration is that `ocamlearlybird` is launched from the switch chosen in this extension.

By the way, I have picked up maintaining and updating earlybird itself such that it works on newer OCaml versions, but that is not yet released (only available directly from https://github.com/hackwaly/ocamlearlybird). It would be nice if this PR was part of the revival of earlybird, such that its own extension could be deprecated.


tmattio:
I'm thrilled to see this PR, thanks a lot for working on this @sim642!

Coincidentally, we've had a lot of discussions related to debugging with the OCaml Platform team recently. One thing that was blocking the adoption of earlybird was its license. I see that you've moved the code (I assume with the permission of the author) to be licensed under MIT. That's great, this is going to enable the use of earlybird in the Platform!

Before we consider using earlybird as a DAP in VSCode, it'd be great to incubate it in the Platform: the vscode extension is a driver for the tools of the OCaml Platform. Would you be ok to work together on an RFC for the incubation of early bird as part of your work to have earlybird used in VSCode?

tmattio:
We discussed this further with @rgrinberg.

We agree that promoting an incubation too early doesn't make sense. At the same time, an integration in VSCode would help earlybird to get the user feedback and user testing it needs to demonstrate that it's ready for incubation in the Platform.

We'll still want to wait for it to be incubated to fully commit to maintaining an integration, but in the immediate term, we could merge a first integration behind an experimental flag.

tmattio:
> I'm not sure it's possible to put the debugger declaration behind a configuration option in package.json.

We could maintain a staging branch for pre-release versions (https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions)

rgrinberg:
I would personally lean towards just marking this feature as "experimental" in big letters and making it available without much hassle. At this point, the main advantage of integrating this feature would be to gather some user experience reports from using it. It would be counter productive to make people go through another step before they're able to use this.

sim642:
> I would personally lean towards just marking this feature as "experimental" in big letters and making it available without much hassle.

Indeed as-is this isn't too visible: it just shows up in the list of debuggers when trying to add a a debug configuration and in the context menu for `.bc` files. Right now there aren't any messages suggesting to `opam install earlybird` either.

I'll append "experimental" to the registered debugger name, etc, which should be visible to anyone trying to activate it.

tmattio:
Alright, as long as (1) we're comfortable we'll be able to remove support (without alienating users) in case we see that earlybird isn't ready for wide adoption just yet, or that the maintenance status becomes unclear in the future and (2) we have in mind that we should aim for incubation to commit to long term support - I'm happy to go with the integration you see fit.

rgrinberg:
It should be possible to replace or remove this component without much hassle. All debuggers have the same interface in vscode so swapping in our own will not be an issue if it becomes necessary.

@3Rafal @mnxn can you give this some testing and review?

mnxn (changes requested):
I left some comments after reading the code but I haven't had a chance to try the debugger yet.

Let me know if you have any questions or need help with something.

sim642:
I appreciate all the feedback on the bindings. I'm not familiar with the whole mechanism and just cherry-picked parts of #650. I'll try to make the suggested changes soon.

mnxn (changes requested):
Nice work with the binding changes!
I added a comment about one more issue in the code that I overlooked.

I attempted to use the debugger yesterday on macOS, but wasn't able to get it to work. I'll try again soon, but I think that either my setup is broken or there might be a lack of feedback on invalid configurations.

mnxn (approved):
I got the debugging feature working on macOS. The only awkward behavior I noticed was that the debugger doesn't terminate if the `a.out` file (from the default launch configuration) doesn't exist. The integrated terminal will show a missing file error, but the VS Code debugging UI still stays active. I'm not sure there's a way to present an VS Code error in this case, but it might be worth looking into in the future.

I believe this PR is ready to merge in a experimental state.

sim642:
> The only awkward behavior I noticed was that the debugger doesn't terminate if the `a.out` file (from the default launch configuration) doesn't exist. The integrated terminal will show a missing file error, but the VS Code debugging UI still stays active. I'm not sure there's a way to present an VS Code error in this case, but it might be worth looking into in the future.

Thanks for the feedback! As a quick improvement I changed that `a.out` path to something more suggestive.
In some further future it'd be nice if the extension could get a list of executables from dune and the debugging configurations would be suggested dynamically from them.

src-bindings/vscode/vscode.ml
```diff
@@ -2978,15 +2978,14 @@ module Window = struct
...
-           ([ `TextDocument of TextDocument.t | `Uri of Uri.t ][@js.union])
+         document:TextDocument.t
       -> ?column:ViewColumn.t
       -> ?preserveFocus:bool
       -> unit
       -> TextEditor.t Promise.t
       [@@js.global "vscode.window.showTextDocument"]
 
-    val showTextDocument2 :
+    val showTextDocument' :
```
rgrinberg, line 2988: Could you describe the issue this commit is fixing? I think also putting it in a separate PR would be valuable.

src/earlybird.ml
```diff
@@ -0,0 +1,51 @@
+open Import
+
+module VariableGetClosureCodeLocation = struct
```
rgrinberg, line 3: Could you add an mli to this module?

src-bindings/vscode/vscode.ml
sim642: In VSCode extension API `showTextDocument` has a couple of overloads. Specifically the cases _without_ `TextDocumentShowOptions` only take `TextDocument` arguments, not `Uri`: https://code.visualstudio.com/api/references/vscode-api#window.showTextDocument.

I think there wasn't a problem before because the `Uri` overload is not used by OCaml Platform.

mnxn:
The debugging feedback is much better now when the file doesn't exist.
I'm still concerned that the initial configuration is invalid by default.

The VS Code debugger extension example shows a way to ask for a program name: https://code.visualstudio.com/api/extension-guides/debugger-extension#anatomy-of-the-package.json-of-a-debugger-extension

Maybe it would be better to ask for a path to the program go debug that way?

sim642:
I think I can give it a try although typing in a long path into `_build` into a pop-up dialog isn't super convenient either. The most convenient way is to use the "Start debugging" command from the file explorer context menu on a `.bc` file, although it's completely undiscoverable right now.
I should probably also add something to README about the debugging support, both for OCaml Platform and earlybird itself (if OCaml Platform becomes the new way to use it).

mnxn:
I agree writing a long path isn't ideal, but I think having this behavior is the best option for now. In the future (depending on how the experiment progresses), the command could be changed to list executables built by dune like you mention.

In the meantime, you can describe the right-click workflow in the README so it's easier to find.

sim642:
> I agree writing a long path isn't ideal, but I think having this behavior is the best option for now. In the future (depending on how the experiment progresses), the command could be changed to list executables built by dune like you mention.

I found a nice compromise using `vscode.window.showOpenDialog` and implemented it for the ask command.
 
> In the meantime, you can describe the right-click workflow in the README so it's easier to find.

That and other information about earlybird is now in the README.

src/earlybird.ml
```diff
@@ -0,0 +1,85 @@
...
+    type t = { location : range option } [@@js]
+  end
+end
+
+let debugType = Extension_consts.Debuggers.earlybird
+
+let createDebugAdapterDescriptor ~instance ~session:_ ~executable:_ =
+  let sandbox = Extension_instance.sandbox instance in
+  let command = Sandbox.get_command sandbox "ocamlearlybird" [ "debug" ] in
+  let { Cmd.bin; args } = Cmd.to_spawn command in
```
rgrinberg, line 37: Can we check if the binary is available before trying to run something? In particular, it would be nice to give the user a hint as to which package should be installed.

src/earlybird.ml
```diff
@@ -0,0 +1,85 @@
...
+  in
+  ExtensionContext.subscribe extension ~disposable;
+
+  let callback ~args:_ =
+    let open Promise.Syntax in
+    let defaultUri =
+      Option.map (Workspace.rootPath ()) ~f:(fun path -> Uri.parse path ())
+    in
+    let filters =
+      Interop.Dict.singleton "OCaml Bytecode Executable" [ "bc"; "d.byte" ]
```
rgrinberg, line 59: Is d.byte for ocamlbuild? If so, I propose we don't include it. I doubt anyone will find it useful.

sim642: Possibly. Earlybird's own VS Code extension handled that alternative so I ported it over but I'll gladly remove it here (and for the explorer context menu command).

src/earlybird.ml
sim642: #650 had something to that extent but I removed it when cleaning things up. Earlybird's own VS Code extension didn't even have such check.

I'll give it a shot.

sim642: The check and error message are now added.

PR #1149, opened 2023-06-23
chore: Update list of ocaml lsp versions

3Rafal:


src/ocaml_lsp.ml
```diff
@@ -112,12 +112,16 @@ let lsp_versions =
          ; "1.14.2"
          ; "1.15.0-4.14"
          ; "1.15.1-4.14"
+         ; "1.16.0"
```
rgrinberg, line 115: Should be 1.16.0-4.14 actually

src/ocaml_lsp.ml
```diff
@@ -112,12 +112,16 @@ let lsp_versions =
...
+         ; "1.16.1"
         |] )
     ; ( (5, 0)
       , [| "1.13.2~5.0preview"
          ; "1.15.0~5.0preview1"
          ; "1.15.0-5.0"
          ; "1.15.1-5.0"
+         ; "1.16.0"
```
rgrinberg, line 123: There's no 1.16.0 for 5.0

rgrinberg:
How come you've removed 1.16.1 and 1.16.0-4.14? It doesn't really matter because not many people are using them, but there's no harm in keeping the list of versions exhaustive.

3Rafal:
I brought them back immediately, because I figured out that they may be useful

3Rafal:
@rgrinberg , ocaml-lsp 1.16.2 was merged to opam-repo: https://github.com/ocaml/opam-repository/pull/23985
This means that we can merge this PR too, right?

rgrinberg:
Yup. Note that we could merge this PR even earlier. We just need to wait until opam-repository is updated before releasing.

smorimoto (approved):

3Rafal:
I was thinking of some edge cases, where opam-repository PR would be rejected and then we would need to revert, but the risk is low and even if it happened it probably wouldn't be a big deal. :)

voodoos:
Sorry, I am a little late to the party, but `1.16.1-2` releases also support OCaml 5.1 (alpha2 and very probably the future final release), are we waiting for the release to add it to the list ?

rgrinberg:
Don't think we released yet so you can send a PR for 5.1

PR #1157, opened 2023-06-30
Support lsp server settings

Khady:
ocamllsp has settings for codelens and extendedHover. They can be changed dynamically. And they are now critical since by default the codelens are disabled without an easy way to restore them. This is an effort to support those settings in vscode.

https://github.com/ocaml/ocaml-lsp/blob/ca325e770100adca845a563cad7eaffa75480fbc/ocaml-lsp-server/docs/ocamllsp/config.md?plain=1#L1-L25
https://github.com/ocaml/ocaml-lsp/pull/1134

This work is paid for by Ahrefs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -472,6 +472,16 @@
           "default": [],
           "markdownDescription": "Extra arguments to pass to ocamllsp."
         },
+        "ocaml.server.codelens": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Enable/Disable codelens"
+        },
+        "ocaml.server.extendedHover": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Enable/Disable extended hover"
+        },
         "ocaml.dune.autoDetect": {
           "type": "boolean",
           "default": true,
--- a/package.json
+++ b/package.json
@@ -474,7 +474,7 @@
         },
         "ocaml.server.codelens": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "markdownDescription": "Enable/Disable codelens"
         },
         "ocaml.server.extendedHover": {
--- a/src-bindings/vscode_languageclient/vscode_languageclient.ml
+++ b/src-bindings/vscode_languageclient/vscode_languageclient.ml
@@ -78,7 +78,7 @@ module DocumentSelector = struct
     `Filter (DocumentFilter.createLanguage ~language:l ~scheme ?pattern ())
 end
 
-module OcamllspSetting = struct
+module OcamllspSettingEnable = struct
   include Interface.Make ()
 
   include
@@ -88,32 +88,18 @@ module OcamllspSetting = struct
     val create : enable:bool -> unit -> t [@@js.builder]]
 end
 
-module OcamllspConfig = struct
+module OcamllspSettings = struct
   include Interface.Make ()
 
   include
     [%js:
-    val codelens : t -> OcamllspSetting.t or_undefined [@@js.get]
+    val codelens : t -> OcamllspSettingEnable.t or_undefined [@@js.get]
 
-    val extendedHover : t -> OcamllspSetting.t or_undefined [@@js.get]
-
-    val createCodelens :
-         codelens:OcamllspSetting.t
-      -> ?extendedHover:OcamllspSetting.t
-      -> unit
-      -> t
-      [@@js.builder]
-
-    val createExtendedHover :
-         ?codelens:OcamllspSetting.t
-      -> extendedHover:OcamllspSetting.t
-      -> unit
-      -> t
-      [@@js.builder]
+    val extendedHover : t -> OcamllspSettingEnable.t or_undefined [@@js.get]
 
     val create :
-         ?codelens:OcamllspSetting.t
-      -> ?extendedHover:OcamllspSetting.t
+         ?codelens:OcamllspSettingEnable.t
+      -> ?extendedHover:OcamllspSettingEnable.t
       -> unit
       -> t
       [@@js.builder]]
@@ -130,13 +116,13 @@ module ClientOptions = struct
 
     val revealOutputChannelOn : t -> RevealOutputChannelOn.t [@@js.get]
 
-    val initializationOptions : t -> OcamllspConfig.t [@@js.get]
+    val initializationOptions : t -> Jsonoo.t [@@js.get]
 
     val create :
          ?documentSelector:DocumentSelector.t
       -> ?outputChannel:Vscode.OutputChannel.t
       -> ?revealOutputChannelOn:RevealOutputChannelOn.t
-      -> ?initializationOptions:OcamllspConfig.t
+      -> ?initializationOptions:Jsonoo.t
       -> unit
       -> t
       [@@js.builder]]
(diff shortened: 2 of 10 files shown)


Khady:
sadly ocamllsp doesn't appear to have support for those settings through initializationOptions yet, but I think it would be good to have so I'm leaving this code here anyway

src/vscode_ocaml_platform.ml
```diff
@@ -37,6 +37,37 @@ let activate (extension : ExtensionContext.t) =
...
+               LanguageClient.OcamllspConfig.create ?codelens ?extendedHover ()
+             in
+             let payload =
+               LanguageClient.DidChangeConfiguration.create ~settings:options ()
+             in
+             LanguageClient.sendNotification
+               client
+               "workspace/didChangeConfiguration"
+               payload)
+         ());
```
Khady, line 70: I'm not sure where this code should live if not here

src-bindings/vscode_languageclient/vscode_languageclient.ml
```diff
@@ -207,5 +259,8 @@ module LanguageClient = struct
       -> Jsonoo.t Promise.t
       [@@js.call]
 
-    val registerFeature : t -> feature:StaticFeature.t -> unit [@@js.call]]
+    val registerFeature : t -> feature:StaticFeature.t -> unit [@@js.call]
+
+    val sendNotification : t -> string -> DidChangeConfiguration.t -> unit
```
Khady, line 264: here this is partially incorrect, as the payload will not be of type `DidChangeConfiguration.t` if the notification is not for `workspace/didChangeConfiguration`, but I don't know how to encode it

package.json
```diff
@@ -472,6 +472,16 @@
           "default": [],
           "markdownDescription": "Extra arguments to pass to ocamllsp."
         },
+        "ocaml.server.codelens": {
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Enable/Disable codelens"
+        },
```
Khady, line 479: I'm not a fan of having this option in addition to `editor.codeLens`, but I'm not sure if it's ok to rely only on `editor.codeLens`

aspeddro: When the `editor.codelens` option is disabled the client (vscode) continues making requests to the server, but does not render.

In the past I have observed this behavior in the server trace log.

package.json
```diff
@@ -472,6 +472,16 @@
           "default": [],
           "markdownDescription": "Extra arguments to pass to ocamllsp."
         },
+        "ocaml.server.codelens": {
```
aspeddro, line 475: Can't we put all the settings inside the `ocaml.server.settings` object?

Khady: As for as I could see it's not possible to define an object with a specific set of fields. So if we want those options to be nicely available in the GUI of vscode I think they have to be independent of each other and not in an object. But I'm happy to be corrected if this isn't true.

src/extension_instance.ml
```diff
@@ -46,10 +46,24 @@ end = struct
...
+        Option.map
+          Settings.(get server_codelens_setting)
+          ~f:(fun enable -> LanguageClient.OcamllspSetting.create ~enable ())
+      in
+      let extendedHover =
+        Option.map
+          Settings.(get server_extendedHover_setting)
+          ~f:(fun enable -> LanguageClient.OcamllspSetting.create ~enable ())
+      in
+      LanguageClient.OcamllspConfig.create ?codelens ?extendedHover ()
```
Khady, line 60: I think that this format is correct for the initialization options, but I'm not sure how to confirm

rgrinberg: initializationOptions and configuration are not the same thing. See this discussion https://github.com/Microsoft/language-server-protocol/issues/567

src-bindings/vscode_languageclient/vscode_languageclient.ml
rgrinberg: What about just making it a generic json payload and then encoding it at the call site?

src/extension_instance.ml
Khady: apparently using `workspace/didChangeConfiguration` with a payload is also bad, so what is recommended in ocamllsp is not good practice.

https://github.com/microsoft/language-server-protocol/issues/567#issuecomment-448538082

Looks like they never actually updated the spec with something clearer :(

rgrinberg: Indeed. Feel free to update the server to account for this.

src/settings.ml
```diff
@@ -113,3 +113,17 @@ let server_args_setting =
...
+    ~key:"ocaml.server.codelens"
+    ~of_json:Jsonoo.Decode.(bool)
+    ~to_json:Jsonoo.Encode.(bool)
+
+let server_extendedHover_setting =
+  create_setting
+    ~scope:ConfigurationTarget.Workspace
+    ~key:"ocaml.server.extendedHover"
+    ~of_json:Jsonoo.Decode.(bool)
+    ~to_json:Jsonoo.Encode.(bool)
```
mnxn, line 129: The parentheses around `bool` aren't necessary.

src/vscode_ocaml_platform.ml
mnxn: I don't think there's enough here to require it's own module, so I would extract this code into a function in this file.

Also, could the `let sandbox_opt = ...` be moved after this code? It doesn't look like the new additions rely on its value.

src/settings.ml
Khady: right, fixed

src-bindings/vscode_languageclient/vscode_languageclient.ml
Khady: I went the generic way with a `Ojs.t` payload

src/vscode_ocaml_platform.ml
Khady: ok I moved the code to its own function. And indeed it doesn't have to be after `sandbox_opt`, that was a mistake

mnxn (approved):

src/extension_instance.ml
Khady: I don't have too much time to dedicate to this issue currently. So for now I dropped the `initializationOptions` part and made sure the `workspace/didChangeConfiguration` way works, which should give good support of ocamllsp 1.16.{1,2}.

I created an issue and I'll comment there if I have time to work on the switch to `workspace/configuration`
https://github.com/ocaml/ocaml-lsp/issues/1160

src-bindings/vscode_languageclient/vscode_languageclient.ml
```diff
@@ -78,6 +78,47 @@ module DocumentSelector = struct
...
+      -> unit
+      -> t
+      [@@js.builder]
+
+    val createExtendedHover :
+         ?codelens:OcamllspSetting.t
+      -> extendedHover:OcamllspSetting.t
+      -> unit
+      -> t
+      [@@js.builder]
```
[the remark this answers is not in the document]
Khady: I dropped them

Khady:
Thanks for the review. I made the changes accordingly. The CI is red for the changelog. I'm not sure what's the expected fix.

mnxn:
> The CI is red for the changelog. I'm not sure what's the expected fix.

You can ignore that failure. It looks like the changelog check fails because it's hardcoded to look for an edit in `CHANGES.md`, but VS Code requires the changes file to be called `CHANGELOG.md`.

Khady:
the mac CI had a timeout but otherwise it's ready to be merged I believe.

just-max:
Any chance of getting this in a release very soon? From what I understand, there is no way of re-enabling CodeLens in the current release, leaving users with effectively a broken extension.

If it is currently possible, could someone point me in the right direction (and possibly also document how, perhaps also on the Marketplace extension's page?).

PR #1164, opened 2023-07-07
Update Action to check the right file

Leonidas-from-XIV:
The action has been updated so it accepts a filename to be checked.

Pointed out by @smorimoto.


smorimoto:
Thank you so much! @Leonidas-from-XIV

PR #1165, opened 2023-07-07
Add more patterns to `opam-metadata.source`

smorimoto:
https://dune.readthedocs.io/en/stable/dune-files.html#stanza-source

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - Maintain focus on editor when sending code to REPL (#1161)
 
+- Add more patterns to `opam-metadata.source` (#1165)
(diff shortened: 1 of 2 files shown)


smorimoto:
@mnxn We can also manage it as a single "hosting services" pattern, but I wasn't sure what the policy was.

mnxn:
Since the structure of the repo name is the same for the 4 hosting services, I think it would be better to combine the 4 rules into one pattern.

smorimoto:
Agreed 👍

smorimoto:
OK, I updated.

mnxn (approved):

PR #1192, opened 2023-08-01
Add instructions for debugging with dune 3.0+

vthemelis:
See https://github.com/hackwaly/ocamlearlybird/issues/20#issuecomment-1658203855

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,17 @@ Two steps to set up debugging:
    For newer OCaml version support, opam pin the development version from
    <https://github.com/hackwaly/ocamlearlybird>.
 
-2. Build _bytecode_ version of your OCaml program executable.
+2. If you use `dune` language version 3.0+ to build your project, switch to 3.7+
+   and make sure that you add:
+
+   ```
+   (map_workspace_root false)
+   ```
+
+   to your `dune-project` file. More info on this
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -277,7 +277,15 @@ Two steps to set up debugging:
    For newer OCaml version support, opam pin the development version from
    <https://github.com/hackwaly/ocamlearlybird>.
 
-2. Build _bytecode_ version of your OCaml program executable.
+2. If you use `dune` 3.0+ to build your project, switch to 3.7+ and make sure that you add:
```
sim642, line 280: It would be more precise to refer to the `(lang dune ...)` used by the project, rather than the version of dune itself. For example, `(lang dune 2.8)` still works fine even if dune 3.7 is installed.

vthemelis: @sim642 thanks, changed!

sim642 (approved):

vthemelis:
@sim642, thanks for the review! I think that you also need to enable the workflows before I can merge this PR.

mnxn:
There was a failing check because there's no changelog entry.

I don't think a README edit requires an entry, so I just added the `no changelog` label to this PR.

vthemelis:
Thanks @mnxn! There was also an issue with formatting. I've now ran prettier.

mnxn (approved):
Thanks!

Issue #1186, opened 2023-07-21
restart every startup


axplus:
Type: <b>Bug</b>

Extension host terminated unexpectedly 3 times within the last 5 minutes.

Extension version: 1.13.0
VS Code version: Code 1.80.1 (74f6148eb9ea00507ec113ec51c489d6ffb4b771, 2023-07-12T17:20:58.115Z)
OS version: Darwin arm64 22.5.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 (8 x 24)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: disabled_off<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>metal: disabled_off<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|2, 2, 3|
|Memory (System)|16.00GB (0.24GB free)|
|Process Argv|--crash-reporter-id 924721b6-5f82-411c-ac63-5e226384253e --inspect-extensions=64896 --crash-reporter-id 924721b6-5f82-411c-ac63-5e226384253e|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vsreu685:30147344
python383cf:30185419
vspor879:30202332
vspor708:30202333
vspor363:30204092
vslsvsres303:30308271
vserr242cf:30382550
pythontb:30283811
vsjup518:30340749
pythonptprofiler:30281270
vshan820:30294714
vstes263cf:30335440
vscorecescf:30445987
vscod805cf:30301675
binariesv615:30325510
bridge0708:30335490
bridge0723:30353136
vsaa593cf:30376535
pythonvs932:30410667
py29gd2263cf:30792227
vsclangdf:30486550
c4g48928:30535728
dsvsc012:30540252
pynewext54:30695312
azure-dev_surveyone:30548225
vsccc:30610678
3biah626:30602489
89544117:30613380
2i9eh265:30646982
showlangstatbar:30737416
vsctsb:30748421
03d35959:30757346
pythonfmttext:30731395
[...]

Issue #1077, opened 2023-01-29
Extension crashes when using HTTP proxy


QDelta:
Environment: macOS 13.2 on M1 Pro, OCaml Platform v1.12.2 on VSCode 1.74.3

I noticed random crashes with message:

```
[Error: read ECONNRESET	at TLSWrap.onStreamRead (node:internal/stream_base_commons:217:20)Emitted 'error' event on ClientRequest instance at:	at TLSSocket.socketErrorListener (node:_http_client:442:9)	at TLSSocket.emit (node:events:526:28)	at emitErrorNT (node:internal/streams/destroy:157:8)	at emitErrorCloseNT (node:internal/streams/destroy:122:3)%c	at process.processTicksAndRejections (node:internal/process/task_queues:83:21)] {%c  errno: -54,%c  code: 'ECONNRESET',%c  syscall: 'read'%c}
```

Workaround:  set `http.proxySupport = "off"`

I'm not sure whether this is a bug of the extension or the proxy server.

bzy-debug:
disable proxy works for me. Thanks, my friend.

rgrinberg:
Please re-open if you can reproduce in 1.13.1

PR #1212, opened 2023-09-05
Fix unintended extension host termination

bzy-debug:
There are several issues about this extension causing extension host terminated unexpectedly: #1186 #1077, etc.

After investigating the compiled javascript code. I think those problems are related with jsoo's handling with uncaught exception.

In the compiled javascript code (line 7804 of _build/default/src/vscode_ocaml_platform.bc.js):

``` js
   function caml_setup_uncaught_exception_handler(){
    var process = globalThis.process;
    if(process && process.on)
     process.on
      ("uncaughtException",
       function(err, origin){
        caml_fatal_uncaught_exception(err);
        process.exit(2);
       });
    else if(globalThis.addEventListener)
     globalThis.addEventListener
      ("error",
       function(event){
        if(event.error) caml_fatal_uncaught_exception(event.error);
       });
   }
   caml_setup_uncaught_exception_handler();
```

This code register a callback that can terminate current process , which is the extension host process, when `uncaughtException` happens.

Since all vscode extensions run on the same extension host process. This callback will be invoked when other extensions trigger uncaught exceptions.

This PR solves this problem by simply comment the invocation of the setup function, so that the dangerous callback function won't be registered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # Unreleased
 
 - Fix highlighting of flags stanza in dune files (#1182)
+- Fix extension crashing when overriding global exception handler (#1212)
 
 ## 1.13.0
 
--- a/src-bindings/node/node.ml
+++ b/src-bindings/node/node.ml
@@ -35,13 +35,7 @@ module Process = struct
 
     val platform : string [@@js.global "process.platform"]
 
-    val arch : string [@@js.global "process.arch"]
-
-    val listeners : string -> (unit -> unit) array
-    [@@js.global "process.listeners"]
-
-    val removeListener : string -> (unit -> unit) -> unit
-    [@@js.global "process.removeListener"]]
+    val arch : string [@@js.global "process.arch"]]
 
   module Env = struct
     include [%js: val env : Ojs.t [@@js.global "process.env"]]
--- a/src-bindings/node/node.ml
+++ b/src-bindings/node/node.ml
@@ -35,7 +35,13 @@ module Process = struct
 
     val platform : string [@@js.global "process.platform"]
 
-    val arch : string [@@js.global "process.arch"]]
+    val arch : string [@@js.global "process.arch"]
+
+    val listeners : string -> (unit -> unit) array
+    [@@js.global "process.listeners"]
+
+    val removeListener : string -> (unit -> unit) -> unit
+    [@@js.global "process.removeListener"]]
 
   module Env = struct
     include [%js: val env : Ojs.t [@@js.global "process.env"]]
--- a/src-bindings/node/node.mli
+++ b/src-bindings/node/node.mli
@@ -29,10 +29,6 @@ module Process : sig
 
   val arch : string
 
-  val listeners : string -> (unit -> unit) array
-
-  val removeListener : string -> (unit -> unit) -> unit
-
   module Env : sig
     val get : string -> string option
 
--- a/src-bindings/node/node.mli
+++ b/src-bindings/node/node.mli
@@ -29,6 +29,10 @@ module Process : sig
 
   val arch : string
 
+  val listeners : string -> (unit -> unit) array
+
+  val removeListener : string -> (unit -> unit) -> unit
+
   module Env : sig
     val get : string -> string option
 
--- a/src/import.ml
+++ b/src/import.ml
@@ -27,11 +27,11 @@ include Base
 
 (* This is used to remove the listener jsoo adding, which could cause vscode
    extension host terminated unexpectedly *)
-let _ =
-  let listeners = Process.listeners "uncaughtException" in
-  Process.removeListener
-    "uncaughtException"
-    (Array.get listeners (Array.length listeners - 1))
+let () =
+  Js_of_ocaml.Js.Unsafe.eval_string
+    "process.removeListener('uncaughtException', \
+     process.listeners('uncaughtException')[process.listeners('uncaughtException').length \
+     - 1])"
 
 module List = struct
   include List
--- a/src/import.ml
+++ b/src/import.ml
@@ -25,6 +25,14 @@ let property_exists json property =
 
 include Base
 
+(* This is used to remove the listener jsoo adding, which could cause vscode
+   extension host terminated unexpectedly *)
+let _ =
+  let listeners = Process.listeners "uncaughtException" in
+  Process.removeListener
+    "uncaughtException"
+    (Array.get listeners (Array.length listeners - 1))
+
 module List = struct
   include List
 


rgrinberg:
cc @hhugo

bzy-debug:
The failed build on mac is because its `sed` is not compatible with gnu's `sed`.

It's not hard to write a compatible script to comment out the invoke of function.

Problem is, I don't think it's a good idea to directly change the compiled js code. It would be great if jsoo can be configured to not to produce this code.

rgrinberg:
@hhugo why does jsoo call `process.exit(2)` in the uncaught exception handler?

bzy-debug:
Any progress for this PR?

mnxn:
While you were investigating, were you able to see what the uncaught exception was?

If we prevent such an exception from being raised in the first place, that could be another way to solve this problem.

bzy-debug:
In my computer, the uncaught exception is caused by built-in extensions of vscode.

Each time I open vscode, it will invoke several network requests as below

image

Since I'm in China, I have to use a VPN to get access to internet. The instability of my network through VPN might cause these requests failed with exception `ECONNRESET`, which is the uncaught exception.

There are several ways to improve this situation. For example, disable vscode's telemetry, or disable the proxy support for extensions. But those ways damage capacity of extensions relying on internet access (copilot).

mnxn:
Thank you for the additional information.

While `sed` works, I am worried that it is too fragile of a solution. There may also be consequences if `caml_fatal_uncaught_exception` isn't called as expected.

The core of the problem appears to be the `process.exit` call inside of the `uncaughtException` handler. Another option could be to unregister the listener at runtime with `removeAllListeners` and then recreate it without the `process.exit` line.

https://nodejs.org/api/process.html#process-events
https://nodejs.org/api/events.html#emitterremovealllistenerseventname

rgrinberg:
> Another option could be to unregister the listener at runtime with removeAllListeners and then recreate it without the process.exit line.

If we're sharing the extension host with other plugins, it seems dangerous to have this hook at all tbh. Is there a way to remove only our own listener with `removeAllListeners`?

bzy-debug:
Since `process.on` add listener to the end of listeners array, maybe we can do something like this:

```js
const listeners = process.listeners('uncoughtException')
process.removeListener(listeners[listeners.length - 1])
```

https://nodejs.org/api/events.html#emitteroneventname-listener
https://nodejs.org/api/events.html#emitterlistenerseventname
https://nodejs.org/api/events.html#emitterremovelistenereventname-listener

rgrinberg:
> Since process.on add listener to the end of listeners array, maybe we can do something like this:

That would require us to remove the listener right after we've inserted it? I suppose we can do that after the first use of `Base` in our code (base is the one that originally adds this uncaught exception handler)

src/vscode_ocaml_platform.ml
```diff
@@ -54,6 +54,10 @@ let notify_configuration_changes instance =
     ()
 
 let activate (extension : ExtensionContext.t) =
+  let listeners = Process.listeners "uncoughtException" in
+  let _ =
+    Process.removeListener (Array.get listeners (Array.length listeners - 1))
```
rgrinberg, line 59: My thinking that this code should go right after `include Base` in `import.ml`. The reason is that we'd like to remove the listener as soon as it's added to make sure we aren't removing someone else's listener by accident. Do you agree?

Also, a comment explaining the problem would be quite handy.

bzy-debug: move to `import.ml` and add comment

rgrinberg: Awesome. Thanks.

Just to double check, you confirmed that this fixes the issue? I have no way of reproducing this unfortunately.

bzy-debug: Unfortunately, this does not remove the listener as expected.

Screenshot 2023-09-21 at 21 30 48

As you can see from the above debugger output, uncaughtException event listener hasn't been removed, and the `listeners` variable does not contain the expected listener functions.

I have no idea why this happens, maybe it's a bug of jsoo or something?

However, I've figured out another way to do the same thing, and that works. I've pushed the valid way to the branch.

ulugbekna:
1. Hm, vscode replaces `process.exit` implementation with just a console.warn for extensions - [code](https://github.com/microsoft/vscode/blob/384e191d8e3ef20570e2e209ba69dd7986afdd51/src/vs/workbench/api/node/extensionHostProcess.ts#L80-L87).

With a dummy extension that tries process.exit'ing on activate or when one of its commands is run, I get the intended behavior - 

image

2. Furthermore, if I patch the extension sources and remove `process.exit(2)`, I still get the extension host crashes. 

@bzy-debug , could you confirm that removing the process.exit fixes the crashes for you?

bzy-debug:
removing the `process.exit(2)` does not fix the crashes. But remove the listener on `uncaughtException` fixes the crashes

rgrinberg:
@mnxn @ulugbekna I'm going to merge this fix and make a release unless there's any objections.

@bzy-debug thanks for all the hard work

ulugbekna:
No objections from me.

We can make another relase later if we find a better solution. I’m curious
what’s that the uncaughtException handler that we register does that
crashes the ext host. I’ll try investigating

mnxn:
Perhaps the issue is inside the `caml_fatal_uncaught_exception` function that is called by the listener.

https://github.com/ocsigen/js_of_ocaml/blob/170d15e09730563f111350be6cf005ecd13181cc/runtime/sys.js#L85

@bzy-debug: Maybe I missed it, but did you ever check the developer console after the extension host was terminated? You can open the developer console  from the menu `Help > Toggle Developer Tools`.

ulugbekna:
@mnxn either that or extensions are not expected to register
uncaughtException handlers at all

Curious, what would you expect seeing in the dev console because I don’t
see anything interesting on crashes?

I’ll post some screenshots later

bzy-debug:
@mnxn 

This is the developer after the extension host was terminated

Screenshot 2023-09-24 at 00 17 59

bzy-debug:
Also, I'm curious about one thing in the makefile:

https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/Makefile#L75-L77

Why not use `build-release` when packaging the extension for installment?

mnxn:
> Curious, what would you expect seeing in the dev console because I don’t see anything interesting on crashes?

@ulugbekna: I wasn't expecting anything in particular. I still haven't been able to reproduce the issue, so I was just curious.

Based on @bzy-debug's screenshot, it looks like the extension host is terminating before an exception is rethrown with the `throw err` in https://github.com/ocsigen/js_of_ocaml/blob/170d15e09730563f111350be6cf005ecd13181cc/runtime/sys.js#L98

Unfortunately, I don't think that really tells us much, so I can't think of a solution that's better than the current one at this time.

> Why not use build-release when packaging the extension for installment?

@bzy-debug: I believe our GitHub action calls `make build-release` directly. `make pkg` is just for testing.

mnxn (approved):

ulugbekna:
Turns out `caml_fatal_uncaught_exception` called before `process.exit(2)` just rethrows the exception, which was causing the crashes. It's defined by jsoo [here](https://github.com/ocsigen/js_of_ocaml/blob/170d15e09730563f111350be6cf005ecd13181cc/runtime/sys.js#L85-L101). 

Patching the extension sources to remove the final `throw err` from `caml_fatal_uncaught_exception` def fixes the crashes. 

---


This's the dev console after an ext host crash: 

<img width="797" alt="image" src="https://github.com/ocamllabs/vscode-ocaml-platform/assets/16353531/83d40809-c618-4dac-96db-3a1b2ff66744"

ulugbekna:
Starting from vscode 1.86 (released end of January), an exception rethrown from an `uncaughtException` listener registered from an extension will be [ignored](https://github.com/microsoft/vscode/commit/85419a380bc33bdad694b5e07900eefe3a56b6ab), so this hack can be removed then (given we want to bump the vscode engine version for the extension).

bzy-debug:
I've made a PR to remove this hack: https://github.com/ocamllabs/vscode-ocaml-platform/pull/1307

PR #1216, opened 2023-09-09
Update yarn to 3.6.3

smorimoto:


smorimoto:
We have to wait for this... https://github.com/microsoft/vscode-vsce/pull/788

PR #1235, opened 2023-10-20
opam 2.2

smorimoto:


smorimoto:
This should speed things up massively.

PR #1237, opened 2023-10-22
Indentation increase for more let expressions

DarinM223:
Right now if you write a let expression where the identifier has a single quote in it, the indentation level won't be increased. Then when you write an `in` keyword, the level will be decreased resulting in something like:

```ocaml
let test =
  let a' =
  2
in
a' + foo
```

The same thing applies for `let*` and `let+` expressions:
```ocaml
let test =
  let* a =
  bar
in
ret (a + foo)
```

This PR modifies the indentation regex so that all the binding operators for `let` and `and` are matched as well as single quotes in the identifier:

```ocaml
let test =
  let* a' =
    bar
  in
  ret (a' + foo)
```

### DIFF
--- a/languages/ocaml.json
+++ b/languages/ocaml.json
@@ -25,7 +25,7 @@
   ],
   "indentationRules": {
     "decreaseIndentPattern": "^\\s*(end|done|with|in|else)\\b|^\\s*;;",
-    "increaseIndentPattern": "\\b(begin|object)\\s*$|\\blet[\\*\\+]? [a-zA-Z0-9'_-]+( [^ ]+)* =\\s*$|method!?[ \\t]+.*=[ \\t]*$|->[ \\t]*$|\\b(for|while)(_lwt)?[ \\t]+.*[ \\t]+do[ \\t]*$|(\\btry(_lwt)?$|\\bif\\s+.*\\sthen$|\\belse|[:=]\\s*sig|=\\s*struct)\\s*$",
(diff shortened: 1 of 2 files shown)


languages/ocaml.json
```diff
@@ -25,7 +25,7 @@
   ],
   "indentationRules": {
     "decreaseIndentPattern": "^\\s*(end|done|with|in|else)\\b|^\\s*;;",
-    "increaseIndentPattern": "\\b(begin|object)\\s*$|\\blet [a-zA-Z0-9_-]+( [^ ]+)* =\\s*$|method!?[ \\t]+.*=[ \\t]*$|->[ \\t]*$|\\b(for|while)(_lwt)?[ \\t]+.*[ \\t]+do[ \\t]*$|(\\btry(_lwt)?$|\\bif\\s+.*\\sthen$|\\belse|[:=]\\s*sig|=\\s*struct)\\s*$",
+    "increaseIndentPattern": "\\b(begin|object)\\s*$|\\b(let|and)[\\*\\+]? [a-zA-Z0-9'_-]+( [^ ]+)* =\\s*$|method!?[ \\t]+.*=[ \\t]*$|->[ \\t]*$|\\b(for|while)(_lwt)?[ \\t]+.*[ \\t]+do[ \\t]*$|(\\btry(_lwt)?$|\\bif\\s+.*\\sthen$|\\belse|[:=]\\s*sig|=\\s*struct)\\s*$",
```
mnxn, line 28: Binding operators support more operators than just `+` and `*`.

You can use this regex from our OCaml grammar to support all of them: `([$&*+\\-/=>@^|<][!?$&*+\\-/=>@^|%:]*)`

https://github.com/ocamllabs/vscode-ocaml-platform/blob/a506da28c8c521ecf2033603850d084dc2c5e88b/syntaxes/ocaml.json#L379
https://v2.ocaml.org/releases/5.1/htmlman/bindingops.html
https://v2.ocaml.org/releases/5.1/htmlman/lex.html#core-operator-char
https://v2.ocaml.org/releases/5.1/htmlman/indexops.html#dot-operator-char

mnxn (changes requested):

languages/ocaml.json
DarinM223: Attempted in c705a6e5cde7451f0a3993522a5f0af231069f04

mnxn (approved):

mnxn:
Looks good.
Thanks!

PR #1245, opened 2023-10-24
VScode 1.82.0

smorimoto:
VScode v1.82.0 and later uses Electron v25, which uses Node.js v18 internally[^1], so we should be able to safely raise the ECMAScript version[^2].

[^1]: https://github.com/ewanharris/vscode-versions
[^2]: https://node.green

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,5 @@ updates:
     labels:
       - dependencies
       - no changelog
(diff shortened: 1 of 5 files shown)


smorimoto:
I have no strong incentive to do this for myself, but there should be no longer any users using `<1.82`.

mnxn (approved):

PR #1256, opened 2023-10-26
Simplify gen_js_api bindings

mnxn:
Ever since https://github.com/LexiFi/gen_js_api/pull/166, many of the modules that are defined in our `interop` library have become redundant.

In this PR, I replace the custom modules with those available from `gen_js_api`, where possible. For the generic function bindings, it was also possible to have `gen_js_api` generate the module-handling code.

I removed the previous `generic` type aliases since they make the code more confusing for very little added convenience. After doing this, I realized most of the functor applications in the bindings are unnecessary since they are only needed to call methods.

Eventually, I would still like to use [ts2ocaml](https://github.com/ocsigen/ts2ocaml), but these changes make our current bindings simpler for the time being.

### DIFF
--- a/src-bindings/interop/interop.ml
+++ b/src-bindings/interop/interop.ml
@@ -88,6 +88,14 @@ end
 module Js = struct
   type 'a t = (module Ojs.T with type t = 'a)
 
+  module type Generic = sig
+    type 'a t
+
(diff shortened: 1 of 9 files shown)


smorimoto:
This is a really huge work, fantastic!
I'm sure ts2ocaml is already in decent state to use, but we've been busy lately and haven't been able to work on it much.

PR #1274, opened 2023-11-09
support switch between impl and intf in reason

jchavarri:
I tried including `re` extensions in the switch impl/intf functionality, and it works perfectly fine, including Melange projects:

generate-rei

### DIFF
--- a/src/switch_impl_intf.ml
+++ b/src/switch_impl_intf.ml
@@ -2,7 +2,10 @@ open Import
 
 let insert_inferred_intf ~source_uri client text_editor =
   let open Promise.Syntax in
-  match String.is_suffix source_uri ~suffix:".ml" || String.is_suffix source_uri ~suffix:".re" with
+  match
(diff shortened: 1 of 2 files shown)


mnxn (approved):

src/switch_impl_intf.ml
```diff
@@ -2,11 +2,13 @@ open Import
 
 let insert_inferred_intf ~source_uri client text_editor =
   let open Promise.Syntax in
-  (* XXX this seems sketchy. shouldn't it work for reason as well? *)
-  match String.is_suffix source_uri ~suffix:".ml" with
+  match
```
anmonteiro, line 5: what do you think about making a string set of suffixes and checking for presence based on `Filename.extension`?

That could easily support adding e.g. `mlx` or upcoming syntaxes just by adding a string to a set.

smorimoto: Clean-up should be done separately from this PR. But good point!

Issue #1222, opened 2023-09-24
Support `${workspaceFolder}` Within Custom Sandbox Template


han-tyumi:
I use Nix to manage my OCaml projects.
I've looked at [how to use this extension with direnv/nix?](https://github.com/ocamllabs/vscode-ocaml-platform/issues/984#top), but unfortunately this does not work for me because the template is run at the root of my system.

It would be nice if `template` supported `${workspaceFolder}` so that we may specify that as the directory to look for a Nix shell configuration.
Another option could be to make sure that this command is started in the workspace folder.

I've also noticed Haskell's IDE will automatically wait and pickup the environment once it's loaded rather than failing hard.
So that would maybe be an even cleaner option if something like that could also be implemented.
This way you could just use the global sandbox in this scenario which would be even simpler.

This is my current workaround:

```json
{
  "ocaml.sandbox": {
    "kind": "custom",
    "template": "nix develop \"${DIRENV_DIR#-}\" --command $prog $args"
  }
}
```

Ideally I could switch to one of the following based on the options I mentioned above.

>```json
>{
>  "ocaml.sandbox": {
>    "kind": "custom",
>    "template": "nix develop \"${workspaceFolder}\" --command $prog $args"
>  }
>}
>```
>With `${workspaceFolder}` support.

> ```json
> {
>   "ocaml.sandbox": {
>     "kind": "custom",
>     "template": "nix develop --command $prog $args"
>   }
> }
> ```
>With the command definitely running in the workspace folder.

> ```json
> {
>   "ocaml.sandbox": {
>     "kind": "global"
>   }
> }
> ```
>With the extension supporting some form of "waiting" for the expected environment to be loaded.

mnxn:
We do not support the `${workspaceFolder}` variable because workspaces may have multiple folders and sandbox configs can come from different places.

You can use either of the following instead:

1. `${workspaceFolder:folderName}`, which will select the folder based on its name. The syntax comes from VS Code's [[Variables scoped per workspace folder](https://code.visualstudio.com/docs/editor/variables-reference#_variables-scoped-per-workspace-folder)](https://code.visualstudio.com/docs/editor/variables-reference#_variables-scoped-per-workspace-folder). 

    (implemented in #424)
    
2. `${firstWorkspaceFolder}`, which will use the first folder in the workspace.

    (implemented in #1004)

han-tyumi:
[...]

PR #1297, opened 2023-12-04
Fix esy sandbox setup

AriaFallah:
This wasn't working for me, and I eventually found this https://github.com/ocamllabs/vscode-ocaml-platform/issues/1222#issuecomment-1732451170

```
{
  "ocaml.sandbox": {
    "kind": "esy",
    "root": "${firstWorkspaceFolder}"
  }
}
```

Is the correct way to do this by default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## 1.14.0
 
 - Support switching between impl and intf in reason (#1274)
+- Fix incorrect esy sandbox docs (#1297)
 
 ## 1.13.4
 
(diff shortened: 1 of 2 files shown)


smorimoto:
@AriaFallah Could you add a change log entry?

mnxn (approved):
That section of the README is just an example configuration. The old version wasn't _incorrect_, it just has the name of the folder in the variable, which means the example only works in workspaces with the `vscode-ocaml-platform` folder. If you use the sandbox selector UI or commands in VS Code, it should automatically change absolute paths to use the correct variable for your environment.

That being said, changing the example is a good idea if it's misleading users this way.
The README should probably document both ways (`${workspaceFolder:folderName}` and `${firstWorkspaceFolder}`) in the future though.

PR #1307, opened 2023-12-15
Remove exception Hack

bzy-debug:
This PR removes the hack added to avoid throwing uncaughtException, since vscode can handle uncaughtException gracefully after 1.86

### DIFF
--- a/src/import.ml
+++ b/src/import.ml
@@ -25,14 +25,6 @@ let property_exists json property =
 
 include Base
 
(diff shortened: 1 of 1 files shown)


ulugbekna:
This requires vscode engine bump to 1.86 in package.json

And this should be merged in February when 1.86 is out.

ulugbekna (approved):

smorimoto:
Ref: https://github.com/microsoft/vscode/pull/200759

PR #1325, opened 2024-01-05
chore(deps): bump js_of_ocaml to >=5.52

tatchi:
It seems to reduce the package size.

### Master

```
dist/vscode_ocaml_platform.bc.js      5.0mb

du -sh test_extension.vsix
1.7M    test_extension.vsix
```

### This PR

```
dist/vscode_ocaml_platform.bc.js      3.2mb

du -sh test_extension.vsix
1.0M    test_extension.vsix
```


smorimoto (approved):

smorimoto:
I don't think this is going to change anything. Even without this change, opam should use the latest jsoo by default.

tatchi:
Ah right there's no upper bound. Feel free to close if you want 🤗

PR #1330, opened 2024-01-05
remove unnecessary activationEvents entries

tatchi:
image


package.json
```diff
@@ -1170,27 +1170,8 @@
...
   "activationEvents": [
-    "onLanguage:ocaml",
-    "onLanguage:ocaml.interface",
-    "onLanguage:reason",
-    "onLanguage:ocaml.ocamllex",
-    "onLanguage:ocaml.menhir",
-    "onLanguage:dune",
-    "onLanguage:dune-project",
-    "onLanguage:dune-workspace",
     "onLanguage:cram",
```
smorimoto, line 1173: I think we can remove this too by adding a setting for cram to `breakpoints`.

tatchi: Looks like the way to get the warning to remove this line is to add a `languages/cram.json` file. I did that, but I'm not sure about the contents of that file 😅

smorimoto: Ah, indeed, this change may be incorrect for cram...

smorimoto (approved):

PR #1331, opened 2024-01-05
fallback to ocaml.org documentation page

tatchi:
Fallback to the `ocaml.org` package documentation page if there is no `doc` field in the package's opam file. Note that it doesn't currently check whether the page exists on `ocaml.org`, so it may return a `404` error.

https://github.com/ocamllabs/vscode-ocaml-platform/assets/5595092/20b71a45-c7ab-4115-be16-f6f75ea63e25

### DIFF
--- a/src/treeview_sandbox.ml
+++ b/src/treeview_sandbox.ml
@@ -13,11 +13,6 @@ module Dependency = struct
 
   let tooltip t = Sandbox.Package.synopsis t
 
-  let context_value t =
-    match Sandbox.Package.documentation t with
-    | Some _ -> "package-with-doc"
-    | None -> "package"
-
   let icon _ =
     TreeItem.LightDarkIcon.
       { light = `String (Path.asset "number-light.svg" |> Path.to_string)
@@ -38,7 +33,7 @@ module Dependency = struct
     in
     let item = Vscode.TreeItem.make_label ~label ~collapsibleState () in
     Vscode.TreeItem.set_iconPath item icon;
-    TreeItem.set_contextValue item (context_value dependency);
+    TreeItem.set_contextValue item "package";
     let+ _ =
       Promise.Option.iter
         (fun desc -> TreeItem.set_description item (`String desc))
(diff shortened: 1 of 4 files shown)


tatchi:
cc @sabine does it sound like a reasonable idea?

src/treeview_sandbox.ml
```diff
@@ -98,7 +103,9 @@ module Command = struct
             ProgressOptions.create
               ~location:(`ProgressLocation Notification)
               ~title:
-                (Printf.sprintf "Generating documentation for %s" package_name)
+                (Printf.sprintf
+                   "Generating documentation for lol %s"
+                   package_name)
```
smorimoto, line 108: Is this intended change?

src/treeview_switches.ml
```diff
@@ -33,10 +33,7 @@ module Dependency = struct
 
   let context_value = function
     | Switch _ -> "opam-switch"
-    | Package dep -> (
-      match Opam.Package.documentation dep with
-      | Some _ -> "package-with-doc"
-      | None -> "package")
+    | Package _ -> "package"
```
smorimoto, line 36:
```suggestion
    | Package _ -> "opam-package"
```

src/treeview_sandbox.ml
tatchi: Oops, no 🙈

src/treeview_switches.ml
tatchi: Should we make the same change in `treeview-sandbox` ?

https://github.com/ocamllabs/vscode-ocaml-platform/blob/9bfe28d361668f6a360d1da626576e1b3c3e754b/src/treeview_sandbox.ml#L36

smorimoto: Maybe? I don't have a strong opinion about this change 🙂

tatchi: [`98e76f9` (#1331)](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1331/commits/98e76f9e59970027a79924c9cee8c3f5327490ee)

smorimoto:
This PR also needs a change log entry!

rgrinberg:
@tmattio do you have some feedback on this one?

smorimoto:
Do you have any ideas about this? @sabine @tmattio

smorimoto:
Could you add a change log entry?

smorimoto:
@tatchi Could you add a change log entry?

sabine:
These URLs are stable, so linking to them is fine. Sure, we can't guarantee the docs built on OCaml.org, but since it's a fallback that should be fine! @smorimoto

smorimoto:
Thanks 🙌

tatchi:
> @tatchi Could you add a change log entry?

done (sorry for the delay)

PR #1332, opened 2024-01-05
Use Mold linker in CI

smorimoto:
If things go well, it can speed things up, but until we actually try it, we can't figure it out.


smorimoto:
It may not be as powerful in OCaml 🤷‍♂️

PR #1340, opened 2024-01-07
refactor: move ocamllsp settings to extension

rgrinberg:
These aren't any general purpose bindings, they should live in the
extension

Signed-off-by: Rudi Grinberg <me@rgrinberg.com>

### DIFF
--- a/src/ocaml_lsp.ml
+++ b/src/ocaml_lsp.ml
@@ -1,4 +1,38 @@
 open! Import
+open Interop
+
+module OcamllspSettingEnable = struct
+  include Interface.Make ()
+
+  include
+    [%js:
+    val enable : t -> bool or_undefined [@@js.get]
+
+    val create : enable:bool -> t [@@js.builder]]
+end
+
+module OcamllspSettings = struct
+  include Interface.Make ()
+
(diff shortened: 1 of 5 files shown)


rgrinberg:
cc @ulugbekna and @Khady

src/ocaml_lsp.ml
```diff
@@ -1,4 +1,40 @@
...
+module OcamllspSettingEnable = struct
+  include Interface.Make ()
+
+  include
+    [%js:
+    val enable : t -> bool or_undefined [@@js.get]
+
+    val create : enable:bool -> unit -> t [@@js.builder]]
+
+  let create ~enable = create ~enable ()
```
mnxn, line 13: The unit parameter is not necessary for `js.builder` functions: https://github.com/LexiFi/gen_js_api/blob/master/VALUES.md

That parameter can just be removed instead of adding a wrapper function:

```ocaml
val create : enable:bool -> t [@@js.builder]
```

rgrinberg: Thanks. Fixed it.

mnxn (approved):

PR #1342, opened 2024-01-07
Prettier -> Biome

smorimoto:
Biome is almost fully compatible with Prettier and more than a few times faster.


biome.json
```diff
@@ -0,0 +1,26 @@
...
+    "enabled": true,
+    "ignore": [
+      "tests/fixtures/sample-esy/esy.lock",
+      "tests/fixtures/sample-esy/_esy"
+    ],
+    "indentStyle": "space"
+  },
+  "linter": {
+    "enabled": true,
+    "ignore": ["astexplorer"],
```
smorimoto, line 13: There are too many warnings, so take care of them with another PR.

rgrinberg: Isn't that code vendored anyway? In other words, we can't fix the warnings without making it hard for us to update this.

smorimoto: Ah, indeed.

PR #1343, opened 2024-01-08
Update syntax for ocamlformat

smorimoto:


mnxn:
I think the `\b` regex is causing issues since many of the field/values have dashes.
For example, `wrap-or-vertical` is highlighted in fragments because of the dashes:
image

Based on [these stackoverflow answers](https://stackoverflow.com/questions/10196462/regex-word-boundary-excluding-the-hyphen), I think it can be fixed by replacing the first `\\b` in each pattern with `(?<![\\w-])` and the last `\\b` with `(?![\\w-])`.

mnxn:
That looks fine for handling dashes at the end of the string, but I think it also needs to be handled at the start with lookbehind: `(?<!-)`.

smorimoto:
I think it should be the exact match, but doesn't that make it the partial match?

mnxn:
I think I wasn't clear enough, sorry.

Your current version does negative lookahead after the match to prevent any dashes after the string.
However, it does not handle any dashes before the string. That means there's a partial match in the string `abc-wrap`: https://regex101.com/r/47FtVi/1

Changing the regex to `\b(?<!-)(fit-or-vertical|wrap|wrap-or-vertical)(?!-)\b` fixes the issue: https://regex101.com/r/47FtVi/2

smorimoto:
I just got the point! I reflected that. I think we can avoid this kind of work in the future by handling these highlights on the LSP side.

mnxn (approved):

Issue #1014, opened 2022-09-06
Colour of text in .mld files shouldn't be the same as comments in .ml files


Khady:
In .mld files the color of the text is the same as comments in .ml files. For most color schemes it means having something deemed or super flashy. I suggest to change it to the "default" color (often white in dark themes or black in light themes).

PR #1365, opened 2024-01-23
Add syntax color to ocamldoc markup

dlesbre:
A few ocamldoc markup had syntax color, but most didn't. This PR adds color to most markup, using the similar markup in markdown for the textmate scope. For reference, I used the following:
- ocamldoc syntax reference: https://v2.ocaml.org/manual/ocamldoc.html#s%3Aocamldoc-comments
- odoc syntax reference: https://ocaml.github.io/odoc/odoc_for_authors.html
- odoc differences with ocamldoc: https://ocaml.github.io/odoc/ocamldoc_differences.html

## Changes

Before this PR, only few constructs had syntax coloring:
- `@tags`
- `[inline code]` (note that inline code is rendered using ocaml's syntax highlighting, and not semantic highlighting. Depending on theme, the colors may be very different from the rest of the file).

I added the following:
- `{b bold]` `{1 heading]` `{i italic]` `{e emph}` (colored the same as their markdown equivalents)
- `{^ exponent}` `{_ index}`, likely not colored in most themes (named `markup.exponent.ocamldoc`)
- `{m math}` and `{math math-block}`, odoc extension, colored the same as latex math (which is `support.class` for some reason...)
- `{[ocaml code block]}`
- `{v verbatim v}` and `{@lang[ other language code block]}` (odoc extension), colored as markdown code block
- `{{: link}link-text}`
- `{!reference}` and `{!type:reference]`, as well as `{{!reference}alt-text}`. I'm not really satisfied with the later though, as I can't color both `reference` as ocaml code and `alt-text` as markup (tm grammar don't have a `begin`/`middle`/`end` sadly).
- Lists both simple (leading `-`) and complex `{ul {li item}}`
- Target specific formatting `{%latex: latex code%}` and `{%html: html code %}` as well as others (rendered as verbatim code)
- tags specific syntaxes for `@raise Exn`, `@raises Exn`, `@param id`, `@see "document"`, `@see 'file'` and `@see <url>` 

I haven't added the following:
- `{C center}`, `{L left}` and `{R right}` (not supported by odoc)
- `{!indexlist}` (not supported by odoc)
- html tags (not supported by odoc)

## Test comment

Here is a comment showing off the new color highlights that can be used to test them with different themes:
```ocaml
(** This is a test documentation comment:
  @see <https://v2.ocaml.org/manual/ocamldoc.html#s%3Aocamldoc-comments>
       for ocamldoc syntax reference
  @see <https://ocaml.github.io/odoc/odoc_for_authors.html>
       for odoc (implementation of the reference, plus a few extensions, like math)

  Basic markup:
    {1 heading}
    {b bold} {i italic} {e emph} {m math}
    {^ exponent} {_ index} (these are maked, but not colored by most themes)
    {v verbatim v}
    {C center} {R right} {L left} (There are unmarked and uncolored)
    {math
      math block
    }

  Tags:
    @tag
    @raise Exn description    @raises Exn description
    @param id description
    @see 'filename'           @see "document-name"       @see <link>

  Code:
    [let foo x = x]
    {[
      let foo x = x
    ]}
    Other language are an odoc extension. I haven't added any language specific
    code, so this python snippet isn't highlighted as python, only as verbatim.
    {@python[
      def hello() -> None:
          print("hello world!")
    ]}

  References and links:
    [foo] {!foo} {!module:Foo} {!modules: Foo Bar}
    {{: https://example.com}Example link}
    {{!foo}The foo function}
    {{!module:Foo} the Foo module}

  Lists:
    - simple
    - unorderd list
    + simple
    + numbered list
    {ul {- full}
        {li list {b with bold text} }}
    {ol {- ordered}
        {li list}}

  Target specific blocks:
    {% here is some latex \foo $1$ %}
    {%latex: some more latex $x = x+1$ %}
    {%html: <p>hello</p><a href="https://example.com">example link</a> %}
*)
```

### DIFF
--- a/syntaxes/ocamldoc.json
+++ b/syntaxes/ocamldoc.json
@@ -15,7 +15,7 @@
       "patterns": [
         {
           "comment": "ocamldoc documentation tag @raise",
-          "match": "(\\@raises?)\\s+(\\S+)",
+          "match": "(@raises?)\\s+(\\S+)",
           "captures": {
             "1": { "name": "keyword.doc-tag.ocamldoc" },
             "2": {
@@ -25,7 +25,7 @@
         },
         {
           "comment": "ocamldoc documentation tag @param",
-          "match": "(\\@param)\\s+(\\S+)",
+          "match": "(@param)\\s+(\\S+)",
           "captures": {
             "1": { "name": "keyword.doc-tag.ocamldoc" },
             "2": {
@@ -35,7 +35,7 @@
         },
         {
           "comment": "ocamldoc documentation tag @see",
-          "match": "(\\@see)\\s+(<.*?>|\".*?\"|'.*?')",
+          "match": "(@see)\\s+(<.*?>|\".*?\"|'.*?')",
           "captures": {
             "1": { "name": "keyword.doc-tag.ocamldoc" },
             "2": {
@@ -46,7 +46,7 @@
         {
           "comment": "ocamldoc documentation tag",
           "name": "keyword.doc-tag.ocamldoc",
-          "match": "\\@[[:lower:]]+"
+          "match": "@[[:lower:]]+"
         },
         {
           "comment": "embedded ocaml source",
@@ -66,7 +66,7 @@
         },
         {
           "comment": "embedded other language source",
-          "begin": "(?<!\\\\)\\{\\@\\S+?\\[",
+          "begin": "(?<!\\\\)\\{@\\S+?\\[",
           "end": "(?<!\\\\)\\]\\}",
           "name": "markup.inline.raw.odoc",
           "contentName": "source.embedded.odoc"
--- a/syntaxes/ocamldoc.json
+++ b/syntaxes/ocamldoc.json
@@ -95,7 +95,7 @@
         },
         {
           "comment": "ocamldoc heading tag",
-          "begin": "(?<!\\\\){[0-9]+[[:space:]]",
+          "begin": "(?<!\\\\){[[:digit:]]+[[:space:]]",
           "end": "(?<!\\\\)}",
           "name": "markup.heading.ocamldoc",
           "patterns": [{ "include": "#markup" }]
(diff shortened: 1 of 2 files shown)


syntaxes/ocamldoc.json
```diff
@@ -13,6 +13,36 @@
   "repository": {
     "markup": {
       "patterns": [
+        {
+          "comment": "ocamldoc documentation tag @raise",
+          "match": "(\\@raises?)\\s+(\\S+)",
```
mnxn, line 18: I think escaping the `@` character is actually unnecessary. You can correct the old pattern on line 49 too.

mnxn, line 18: For consistency with the other syntaxes, the regex patterns should use [POSIX character classes](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX_Basic_Regular_Expressions#Character_classes).

That is, `\\s` should be written as `[[:space:]]` and`\\S` as `[^[:space:]]`.

mnxn (changes requested):
Thank you for the very detailed PR. The highlighting changes look great!

I only have a few implementation comments:

syntaxes/ocamldoc.json
dlesbre: Done in [164a490](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1365/commits/164a4901e589085273adbc9447e5cb69fc525314)

dlesbre: Done in [feee111](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1365/commits/feee1117d8dcdf5cc9f46cb089ec26d70a1f204d) and [b650be5](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1365/commits/b650be54968aea6ccb4e3253b04a924f2f36b95f)

dlesbre:
I've also added a small change to fix #1014, (simply removing the global scope name `comment.other.ocamldoc` from the ocamldoc.json. It won't changes the color of comments in `.ml`/`.mli` files since its already added in `ocaml.json`, but it removes the color from `mld` files.

mnxn (approved):
Nice work!

PR #1369, opened 2024-01-26
Fix md code block bug

### LINKED ISSUE #1288: Markdown codeblock syntax highlight being influenced
The syntax highlight in codeblock of markdown file is influenced by this extension when code contains: `ml`, `ocaml`, `re`, `reason`, `reasonml` with at least one whitespace character. Example:

Screenshot 2023-11-24 at 3 53 19 AM

See the red curly brackets in the figure.
After looking through the source code, I think probably the issue lies on these two files: `syntaxes/ocaml-markdown-codeblock.json` and `syntaxes/reason-markdown-codeblock.json` where  the regex of `"begin": ...` may see any of these key words(ml, ocaml, re, ...) as the beginning of an ocaml codeblock.

This shouldn't be a big issue. I tried to fix this by using a new regex, the one that is employed in [vscode source code](https://github.com/microsoft/vscode/blob/main/extensions/markdown-basics/syntaxes/markdown.tmLanguage.json) but failed. Maybe other factors account for this issue.  I really hope you can teach me how to fix this. Or maybe I can make a pr if you are willing to give me some hints? :)

Luckily, the markdown preview is not influenced.

dlesbre:
This fixes #1288. I copied code structure from https://github.com/mjbvz/vscode-fenced-code-block-grammar-injection-example. The short description of this fix is: "don't inject `ml <code>` in an md codeblock, instead inject ```` ```ml\n <code>\n``` ```` in md directly".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+- Fix non-ocaml markdown code block syntax highlighting being influenced (issue #1288, #1369)
+
 ## 1.16.1
 
 - No longer color comments starting three stars or more as ocamldoc comments (#1355)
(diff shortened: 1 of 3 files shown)


mnxn (approved):
Thank you for the fix!

PR #1381, opened 2024-02-02
Prepare 1.17.1

smorimoto:
- #1307
- #1379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+## 1.17.1
+
 - Set the VSCode lower bound to 1.86.0 (#1379)
 
 ## 1.17.0
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ocaml-platform",
   "displayName": "OCaml Platform",
-  "version": "1.17.0",
+  "version": "1.17.1",
   "private": true,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 # Unreleased
 
+## 1.17.1
+
+- Remove exception hack (#1307)
```
ulugbekna, line 7: isn't this internal, meaning doesn't need to be in the release notes?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ocaml-platform",
   "displayName": "OCaml Platform",
-  "version": "1.17.0",
```
ulugbekna, line 4: I don't know what's the precedent, but shouldn't the version be 1.18.0 given we're bumping vscode engine version?

ulugbekna (changes requested):

package.json
rgrinberg: Are there any new features introduced by bumping the engine?

CHANGELOG.md
smorimoto: Fair enough.

package.json
smorimoto: There are no new features, just a bug fix, so I thought a point release would be enough.

ulugbekna (approved):

PR #1382, opened 2024-02-05
Highlight the background of the statement sent to REPL

aqjune:
This patch highlights the background of the statement that was sent to REPL.

To set the color, ocaml.repl.evalTextBackgroundColor is newly added as a configuration.

Modifying any character in the editor before or in the selected region will remove the highlighted background.


aqjune:
ping

package.json
```diff
@@ -375,6 +375,11 @@
           "type": "boolean",
           "default": true,
           "description": "Controls whether to use Utop for the REPL if it is installed in the current switch."
+        },
+        "ocaml.repl.evalTextBackgroundColor": {
+          "type": "string",
+          "format": "color-hex",
+          "description": "The background color of the text evaluated in REPL."
```
mnxn, line 382: Without a `default` value, the VS Code settings show an error: 
image

src/repl.ml
```diff
@@ -169,6 +183,13 @@ module Command = struct
     in
     Extension_commands.register ~id:Extension_consts.Commands.open_repl handler
 
+  let evalTextDecorationType : TextEditorDecorationType.t =
+    let bgcolor = get_repl_evalTextBackgroundColor () in
+    let render_options =
+      Vscode.DecorationRenderOptions.create ?backgroundColor:bgcolor ()
+    in
+    Vscode.Window.createTextEditorDecorationType ~options:render_options
```
mnxn, line 191: This should be a function so users can change the selection background color without having to restart the extension. Since it's a value currently, that means it's only updated once when the extension is started.

mnxn (changes requested):
> To remove the highlighted background, selecting an empty range and sending it to REPL works.

This seems a bit unintuitive since doing <kbd>Shift</kbd>+<kbd>Enter</kbd> without a selection will select the entire line. That means removing the highlight specifically requires selecting whitespace or doing <kbd>Shift</kbd>+<kbd>Enter</kbd> on a blank line.

Keeping this functionality is fine, but I think there should also be an additional command to remove the highlighting.

aqjune:
Hi, thanks for your comments! I could not manage to find a time for addressing your comments today. I will reflect them tomorrow.

package.json
aqjune: I updated its type as string or null.

src/repl.ml
aqjune: I redefined it out as the `getEvalTextDecorationType` function.

src/repl.ml
```diff
@@ -199,4 +251,22 @@ let register extension instance =
           Extension_instance.close_repl instance)
       ()
   in
+  Vscode.ExtensionContext.subscribe extension ~disposable;
+  let disposable =
+    Vscode.Workspace.onDidChangeTextDocument () ~listener:(fun event ->
```
aqjune, line 256: I added this `onDidChangeTextDocument` that unsets the highlighted color if there is any update inside or before the evaluated region.

aqjune:
Hi, sorry for my delay. I reflected the comments so far.

smorimoto:
Could you please add a change log entry?

src/repl.ml
```diff
@@ -143,14 +158,39 @@ let get_code text_editor =
...
+    Vscode.TextLine.range line
+  else (selection :> Vscode.Range.t)
 
 let prepare_code code =
   if String.is_suffix (String.strip code) ~suffix:";;" then code
   else code ^ ";;"
 
+(* Highlighted OCaml code *)
+let prevEvalTextDecorationType : TextEditorDecorationType.t option ref =
+  ref None
```
ulugbekna, line 170: to my knowledge, we don't put mutable state in a module but put it in ExtensionContext, possibly within a record that's related to repl functionality.

src/repl.ml
```diff
@@ -143,14 +158,39 @@ let get_code text_editor =
...
+
+let prevEvalTextDecorationRange : Range.t option ref = ref None
```
ulugbekna, line 172: camelCase is used only for vscode bindings (only symbols defined at `src-bindings/`)

src/repl.ml
```diff
@@ -199,4 +251,22 @@ let register extension instance =
...
+        match
+          (!prevEvalTextDecorationRange, Vscode.Window.activeTextEditor ())
+        with
+        | None, _ | _, None -> ()
+        | Some prevRange, Some editor -> (
+          let poss = TextDocumentChangeEvent.contentChanges event in
+          let h = List.hd poss in
+          match h with
+          | None -> ()
+          | Some h ->
```
ulugbekna, line 266: this could be directly a pattern match on the list

ulugbekna (commented):
Frankly, I'm not sure it's useful to color the background for the current behavior of the REPL since the text sent is based either on the selection which is already colored or the current line if selection is empty. The sent text is also visible in the REPL. Do you have cases when such background felt useful?

do others have opinion on this?

aqjune:
> Frankly, I'm not sure it's useful to color the background for the current behavior of the REPL since the text sent is based either on the selection which is already colored or the current line if selection is empty. The sent text is also visible in the REPL. Do you have cases when such background would be useful?

Yes, I felt this feature was necessary when I was evaluating the text in REPL which is taking a long time.
While the text is evaluating, I cannot unselect the text and edit the .ml program because it makes me lose which text I was evaluating before.

Specifically, I am using vscode-ocaml-platform to write an OCaml program that uses HOL Light (https://github.com/jrh13/hol-light/).
OCaml programmer can use HOL Light to define a mathematical theorem and prove it step-by-step.
An HOL Light program is supposed to run on OCaml REPL rather than compile it as a binary because the user needs to interact with the proof state after each step.
This is computation intensive, and evaluating one sentence can easily take several minutes. Therefore evaluating a sentence and writing a code in parallel is important for saving time. It is also important to not lose which statement I had executed before.

An Emacs plugin called ProofGeneral has a similar feature. It is an editor for a language called Coq, which is similar to OCaml but for more mathematical purpose.
https://www.youtube.com/watch?v=l6zqLJQCnzo shows how the evaluated text is colored as blue.

src/repl.ml
aqjune: Thanks for your comments. I addressed this and two following comments in the newest commit.

rgrinberg:
Since this mode isn't enabled by default, I feel there's not much risk in accepting this PR. It doesn't seem like it will add much maintenance overhead or bugs.

PR #1391, opened 2024-02-18
Add new syntax patterns for dune(-project) files

### LINKED ISSUE #1180: Syntax: dune
##### Note: This list was auto-generated based on a loose comparison of the documentation and the syntax file, so the wrong tasks may have been mixed.

https://dune.readthedocs.io/en/stable/dune-files.html

- [ ] accept_alternative_dune_file_name
- [ ] action_stderr_on_success
- [ ] action_stdout_on_success
- [x] alias
- [x] authors
- [x] bug_reports
- [ ] cache-check-probability
- [ ] cache-storage-mode
- [ ] cache
- [x] cinaps
- [ ] config
- [x] context
- [x] copy_files
- [x] coq.theory
- [x] cram
- [x] data_only_dirs
- [x] deprecated_library_name
- [x] dialect
- [x] dirs
- [ ] display
- [x] documentation
- [x] env
- [x] executable
- [ ] executables_implicit_empty_intf
- [x] executables
- [ ] expand_aliases_in_sandbox
- [x] explicit_js_mode
- [ ] external_variant
- [x] foreign_library
- [x] formatting
- [x] generate_opam_files
- [x] generate_sites_module
- [x] homepage
- [x] ignored_subdirs
- [x] implicit_transitive_deps
- [x] include_subdirs
- [x] include
- [x] install
- [x] jbuild_version
- [ ] jobs
- [x] library
- [x] license
- [x] maintainers
- [ ] map_workspace_root
- [x] mdx
- [x] menhir
- [x] name
- [x] ocamllex
- [x] ocamlyacc
- [ ] opam_file_location
- [x] package
- [x] plugin
- [x] profile
- [x] rule
- [ ] sandboxing_preference
- [x] source
- [x] subdir
- [ ] subst
- [ ] terminal-persistence
- [x] test
- [x] tests
- [x] toplevel
- [ ] use_standard_c_and_cxx_flags
- [x] using
- [x] vendored_dirs
- [x] version
- [x] wrapped_executables

ComanderP:
Resolves #1180


ComanderP:
Hi, this is my first time contributing here, sorry if I did something wrong.
I went through most of the dune documentation and added syntax highlighting to most of the stanzas in the list of the issue.
I think most of the stanzas are covered now, with the exception of the stanzas from the `config` section, which I do not know if they're in a separate file or not.

On a side note, I took note of some things while doing this, concerning the syntax highlighting and the documentation:

- I tried to not touch the `executable(s)` and `library/libraries` stanzas, as I didn't want to mess with the existing syntax highlighting that they have in common.
- The `menhir` stanza added to `env` in version 3.0 of the menhir extension was not added as I was not sure if I was allowed to make a separate rule in the file's repository for both the normal `menhir` stanza and the `menhir` stanza in `env`. I can add it if it's okay. I also didn't want to just add the menhir stanza rules inside the `env` stanza rules, as it would lead to code duplication.
- The documentation for the [`subst` stanza](https://dune.readthedocs.io/en/stable/dune-files.html#stanza-subst) seems to be incorrect, I think. It says that the `subst` stanza takes a `<bool>` as an argument, but it actually takes `<enabled|disabled>`. For example, the [`warnings` stanza](https://dune.readthedocs.io/en/stable/dune-files.html#stanza-subst) takes `<enabled|disabled>` and is documented as such. If you want, I can make a PR to fix this.
- Another thing that I noticed is that the different TextMate grammar files have different formatting styles. This isn't that big of a deal, but it's something that I noticed.
- Should `external_variant` be added to the syntax highlighting? I didn't add it as it was an experimental feature removed in version 2.6 of dune, and there's no documentation for it in the dune docs anymore.
- I noticed that there is a `variable.declaration.other.dune` name and a `variable.other.declaration.dune` name in the syntax highlighting files. I don't know if this is intentional or not, and I do not know if there is a difference between the two. I tried both of them and they both seem to work the same way.
- For the `install` stanza, the documentation only mentions 3 fields (files, section, package), but the syntax highlighting file is highlighting way more fields than that. I think there may have been a mistake with the `executable` fields, as there's a mention of the `install` stanza in the documentation for the `executable` stanza followed by the fields for the `executable` stanza. I'm not sure if this is a mistake or not, but I mentioned it just in case.

I hope this is helpful. Let me know if I should change something or if I missed something.

ComanderP:
Hi, thank you for the feedback!
I've added the config stanzas to the dune-workspace file as you said.
I'll try changing some things based on what you said as well before committing.

ComanderP:
Sorry for the delay, got caught up with my studies.
I fixed/changed/added stuff based on your previous insights.
Hopefully, all's fine now.

syntaxes/dune-workspace.json
```diff
@@ -184,6 +184,129 @@
...
+          "begin": "\\([[:space:]]*(sandboxing_preference)\\b",
+          "end": "\\)",
+          "beginCaptures": {
+            "1": { "name": "keyword.language.dune-workspace" }
+          },
+          "patterns": [
+            {
+              "comment": "none/hardlink/copy/simlink",
+              "name": "variable.other.declaration.dune-project",
+              "match": "\\b(none|hardlink|copy|simlink)\\b"
```
mnxn, line 290: I think it's supposed to be `symlink`, not `simlink`: 
https://dune.readthedocs.io/en/stable/reference/config/sandboxing_preference.html

mnxn (changes requested):
Sorry for the delay. I found one typo. Everything else looks fine.

syntaxes/dune-workspace.json
ComanderP: You're right, I've just changed it.

mnxn (approved):

Issue #935, opened 2022-04-19
ocaml.open-repl doesn't work if the repl has been closed previously


Khady:
I have an issue with `OCaml: Open REPL` or `OCaml: Evaluate Selection` when I have previously closed the repl created by one of those commands.

- run `OCaml: Open REPL`, a wild toplevel appears
- close the utop/repl that was just created with something like ^D
- run `OCaml: Open REPL` again, nothing happens, nothing is printed inside the `OCaml Platform Extension` output

When using vscode with remote-ssh this is especially annoying as `OCaml: Restart Language Server` or `Developer: Reload Window` are not enough to fix the situation, I have to do `Remote-SSH: Kill Current VS Code Server` and reconnect

With version v1.10.1

aqjune:
I am experiencing this bug as well.

aqjune:
It seems that this function is not being called at all even if a terminal is closed:

https://github.com/ocamllabs/vscode-ocaml-platform/blob/d3fc5410444e88f7835cb24262be760fe1daac15/src/repl.ml#L198-L199

aqjune:
Diving further, the reason is that even if a user exits the REPL using Ctrl+D it actually does not dispose the Terminal object.

I tried to detour this problem by registering a function to `Vscode.Window.onDidChangeActiveTerminal` so that if the user closes REPL the Terminal object is disposed too.
However, this could not be done because I could not detect when the REPL is terminated. I tried using `Terminal.exitStatus` but no avail. It is returning None even if the REPL process is terminated.

Actually, looking at the implementation of VSCode, `exitStatus` is set to `Some exitcode` when `Terminal.dispose` is called. `$acceptTerminalClosed` at  https://github.com/microsoft/vscode/blob/7c5742f4b958f3c5efb2284ebaf1152d4d1abf27/src/vs/workbench/api/common/extHostTerminalService.ts#L567-L574 is called when it is disposed (https://github.com/microsoft/vscode/blob/7c5742f4b958f3c5efb2284ebaf1152d4d1abf27/src/vs/workbench/api/browser/mainThreadTerminalService.ts#L340-L342).

PR #1395, opened 2024-02-28
Fix issue #935 by periodically checking the REPL process state

aqjune:
This patch resolves issue #935 by periodically checking the state of the REPL process.

When a REPL terminal is closed by the 'x' button or Ctrl+D from a user, the Terminal object is not automatically disposed.
This was the underlying reason why the original code was not working - `Vscode.Window.onDidCloseTerminal` is invoked when the Terminal object is disposed, but since it was not being disposed, it was not being invoked in practice.

This patch detours this problem by running a timer that periodically checking the existence of the REPL process for every 500 ms, and if the process does not exist, invoke `Terminal.dispose` as well as `Extension_instance.close_repl instance`.
To check the existence of REPL process, `Process.kill ~pid:procid ~signal:(`Number 0)` is used. The signal 0 is reserved for a testing signal for its existence (https://nodejs.org/api/process.html#processkillpid-signal).


src/repl.ml
```diff
@@ -191,12 +191,31 @@ module Command = struct
...
-        if String.equal (Vscode.Terminal.name terminal) name then
-          Extension_instance.close_repl instance)
-      ()
+  let timer =
+    Node.setInterval
+      (fun () ->
+        match Extension_instance.repl instance with
+        | None -> ()
+        | Some t ->
+          let _ =
```
rgrinberg, line 200: Could you add a type annotation to the value is being ignored here?

PR #1398, opened 2024-03-02
Fix opening a new REPL terminal after first one was terminated

### LINKED ISSUE #935: ocaml.open-repl doesn't work if the repl has been closed previously

ulugbekna:
Fixes #935 
Alternative to #1395

---

Seems to fix the issue

https://github.com/ocamllabs/vscode-ocaml-platform/assets/16353531/2e6eba50-1c61-4b28-9711-a079456c956d

----

@aqjune @Khady would you mind verifying? You should be able to checkout this branch and run `make install` and reload your vscode to get this branch version of the extension.

----

cc @rgrinberg

### DIFF
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -195,16 +195,17 @@ let register extension instance =
     let listener closed_terminal =
       ignore
         (match Extension_instance.repl instance with
-        | None -> Promise.return ()
-        | Some repl_terminal ->
-          let open Promise.Syntax in
-          let+ repl_terminal_pid, closed_terminal_pid =
-            Promise.all2
-              ( Terminal.processId repl_terminal
-              , Terminal.processId closed_terminal )
-          in
-          if repl_terminal_pid = closed_terminal_pid then
-            Extension_instance.close_repl instance)
+         | None -> Promise.return ()
+         | Some repl_terminal ->
+           let open Promise.Syntax in
+           let+ repl_terminal_pid, closed_terminal_pid =
+             Promise.all2
+               ( Terminal.processId repl_terminal
+               , Terminal.processId closed_terminal )
+           in
+           if repl_terminal_pid = closed_terminal_pid then
+             Extension_instance.close_repl instance
+          : unit Promise.t)
     in
     Vscode.Window.onDidCloseTerminal () ~listener ()
   in
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -194,9 +194,12 @@ let register extension instance =
   let disposable =
     Vscode.Window.onDidCloseTerminal
       ()
-      ~listener:(fun terminal ->
-        if String.equal (Vscode.Terminal.name terminal) name then
-          Extension_instance.close_repl instance)
+      ~listener:(fun closed_terminal ->
+        match Extension_instance.repl instance with
+        | None -> ()
+        | Some repl_terminal ->
+          if phys_equal repl_terminal closed_terminal then
+            Extension_instance.close_repl instance)
       ()
   in
   Vscode.ExtensionContext.subscribe extension ~disposable
--- a/src/repl.ml
+++ b/src/repl.ml
@@ -195,17 +195,16 @@ let register extension instance =
     let listener closed_terminal =
       ignore
         (match Extension_instance.repl instance with
-         | None -> Promise.return ()
-         | Some repl_terminal ->
-           let open Promise.Syntax in
-           let+ repl_terminal_pid, closed_terminal_pid =
-             Promise.all2
-               ( Terminal.processId repl_terminal
-               , Terminal.processId closed_terminal )
-           in
-           if repl_terminal_pid = closed_terminal_pid then
-             Extension_instance.close_repl instance
-          : unit Promise.t)
+        | None -> Promise.return ()
+        | Some repl_terminal ->
+          let open Promise.Syntax in
+          let+ repl_terminal_pid, closed_terminal_pid =
+            Promise.all2
+              ( Terminal.processId repl_terminal
+              , Terminal.processId closed_terminal )
+          in
+          if repl_terminal_pid = closed_terminal_pid then
+            Extension_instance.close_repl instance)
     in
     Vscode.Window.onDidCloseTerminal () ~listener ()
   in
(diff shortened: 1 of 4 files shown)


smorimoto (approved):

src/repl.ml
```diff
@@ -194,9 +194,12 @@ let register extension instance =
...
     Vscode.Window.onDidCloseTerminal
       ()
-      ~listener:(fun terminal ->
-        if String.equal (Vscode.Terminal.name terminal) name then
-          Extension_instance.close_repl instance)
+      ~listener:(fun closed_terminal ->
+        match Extension_instance.repl instance with
+        | None -> ()
+        | Some repl_terminal ->
+          if phys_equal repl_terminal closed_terminal then
```
rgrinberg, line 201: Could you avoid using physical equality somehow? For example, but adding a unique id to `Terminal_sandbox.t`?

aqjune:
Hi, thanks for trying an alternative solution.

I am afraid that this solution doesn't seem to work in case.
In my case, `Vscode.Window.onDidCloseTerminal` wasn't called at all when the terminal is closed with Ctrl+D. So, fixing anything inside the function wasn't a solution to me.
I am using Ubuntu2 on WSL2, and the version of VSCode is 1.86.2 .
It is interesting to know that this patch works in some cases. Which OS/version are you using?

aqjune:
I found that the difference in behavior appeared when using Utop vs. the default REPL of OCaml.
In my case I was using the default OCaml REPL. When Utop is installed, this patch works well.
Using the default OCaml REPL is necessary because I am depending on an OCaml library that uses camlp5. I wrote more details about my use case at [this comment](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1382#issuecomment-1975207895). What would be a good solution in this case?

src/repl.ml
```diff
@@ -192,11 +192,20 @@ end
...
 let register extension instance =
   let disposable =
-    Vscode.Window.onDidCloseTerminal
-      ()
-      ~listener:(fun terminal ->
-        if String.equal (Vscode.Terminal.name terminal) name then
-          Extension_instance.close_repl instance)
-      ()
+    let listener closed_terminal =
+      ignore
```
rgrinberg, line 196: Could you annotate the value that is being ignored here?

src/repl.ml
```diff
@@ -192,11 +192,20 @@ end
...
+        (match Extension_instance.repl instance with
+        | None -> Promise.return ()
+        | Some repl_terminal ->
+          let open Promise.Syntax in
+          let+ repl_terminal_pid, closed_terminal_pid =
+            Promise.all2
+              ( Terminal.processId repl_terminal
+              , Terminal.processId closed_terminal )
+          in
+          if repl_terminal_pid = closed_terminal_pid then
```
rgrinberg, line 206: In general, relying on pid's isn't the most reliable thing because they can always be reused by the OS. So I think that adding our own unique ID would still be helpful. I don't think this is essential though.

rgrinberg (approved):

src/repl.ml
ulugbekna: > In general, relying on pid's isn't the most reliable thing because they can always be reused by the OS.

But this PID would be provided by vscode, so it would require vscode being buggy?

> So I think that adding our own unique ID would still be helpful. 

I didn't quite understand your suggestion tbh because `closed_terminal` has type `Terminal.t` which is a vscode type, which we can't add an ID to?

rgrinberg: > But this PID would be provided by vscode, so it would require vscode being buggy?

PID's are ultimately provided by the the OS. So in theory it's possible that our listener can get called with a terminal that has the same PID but isn't our repl. This can happen if our terminal is closed ,and a new one is immediately opened and closed, and the OS assigned it the same PID. I don't know the vscode api well enough to know if this is indeed a possibility, but it is a reason why relying on dead PIDs is considered fragile.

> I didn't quite understand your suggestion tbh because closed_terminal has type Terminal.t which is a vscode type, which we can't add an ID to?

Yes, that's my mistake. I thought we were dealing with `Terminal_sandbox.t` here, but this is just a type alias to the vscode type.

ulugbekna:
> In my case, Vscode.Window.onDidCloseTerminal wasn't called at all when the terminal is closed with Ctrl+D. So, fixing anything inside the function wasn't a solution to me.

If this event (provided by vscode) doesn't fire, it must a vscode bug? How did you make sure that the event wasn't firing, ie additional logging, etc? I found installing extension version built from source difficult to install on a "remote" machine. 

I'm on macos, not remote setup. 

> I found that the difference in behavior appeared when using Utop vs. the default REPL of OCaml.

With this fix, both behave same for me but again not sure about remote setup.

ulugbekna:
I propose merging this fix as a first step towards fixing this bug at least on a local machine

src/repl.ml
ulugbekna: I think you're right, and relying on PID is a mess because `processId` _can_ return an undefined based on `vscode.d.ts` rather than online docs. 

I had a look at the terminal service implementation in vscode's extension host, and from the looks, terminal objects should be physically equal. Should I go back to it? Why do you think physical eq is bad given most of the js development is based on `===` which is (?) equivalent to physical eq for objects?

aqjune:
> If this event (provided by vscode) doesn't fire, it must a vscode bug? How did you make sure that the event wasn't firing, ie additional logging, etc? I found installing extension version built from source difficult to install on a "remote" machine.

In my case, clicking this arrow button from 'Run and  Debug' worked. It opened a new VSCode window with the plugin in remachine machine installed.
image

> With this fix, both behave same for me but again not sure about remote setup.
This is interesting to know. Then a reasonable conclusion is that running VSCode at the remote setup might be related.

PR #1402, opened 2024-03-05
Update promise_jsoo to 0.4.2

smorimoto:


mnxn (approved):
Unfortunate timing for the release :(

I just pushed a commit to use the `%js.of` syntax so we don't have to write the conversion function manually. The reason `Promise.void` was removed in the first place was because [`gen_js_api` added standard `unit` conversion functions](https://github.com/LexiFi/gen_js_api/blob/v1.0.7/lib/ojs.mli#L42-L43). Those didn't exist in previous versions, which is why `promise_jsoo` had the `Promise.void` helper.

Not sure if it matters, but is the change log supposed to mention dependency update details? It's not something that's visible to users like the other change log entries.

smorimoto:
Indeed!

PR #1449, opened 2024-04-18
Allow selection of Dune context

### LINKED ISSUE #1432: Feature: Allow users to choose which Dune context to use
We are currently experimenting with separate Melange and OCaml libraries and place them in different Dune contexts, in order to improve the developer experience (see https://github.com/melange-re/melange/discussions/694).

Under this multi-context setup, it would be useful as a user to select which Dune context to use in VSCode for code definitions, types and errors.

We are putting out this proposal to gather feedback from maintainers, and if accepted, make sure the implementation plan is ok.

## Current status

It is currently possible to choose the context by editing `dune-workspace` files and adding `(merlin)` to the context we want to work on, but this has a few downsides:
- It requires to manually edit `dune-workspace` every time one wants to switch context
- Can lead to unwanted changes in this file when using git or other source versioning tools
- There is some bug where dune `ocaml-merlin` won't pick up the new context straight away, so it shows errors like `:No config found for file foo.ml. Try calling 'dune build'` until one restarts the whole VSCode window

## Proposal

To improve the experience, the proposal is to add a new command `select-context` to this extension, which will show users a list of contexts to choose from, and record their choice in vscode settings (similar to the current experience with sandbox selection in `select-sandbox`).

## Implementation

For this to happen, besides the new VSCode command `select-context`, we will have to add a few other pieces in ocaml-lsp-server and Dune's `ocaml-merlin`:

### ocaml-lsp-server

- Add a new request `ocamllsp/getDuneContexts` to the lsp server. This request will resolve to calling `ocaml-merlin` behind the scenes (see below)
- Add a new handling logic in `didChangeConfiguration` for a newly added `duneContext` config. This handling will forward the change to `ocaml-merlin` (see below)

### ocaml-merlin

- Add new command `GetContexts` that returns a list of available contexts in the current workspace, so that they can be sent through `ocamllsp/getDuneContexts` and shown in a list inside VSCode extension UI
- Add a new command `SetContext` that allows to define the current context, which will be called from ocaml-lsp-server `didChangeConfiguration` handler

---

Does the above plan make sense? Thanks!

jchavarri:
Fixes #1432.

## Dependencies

This PR relies on features being added in ocamllsp and dune:
- [ocaml-lsp#1238](https://github.com/ocaml/ocaml-lsp/pull/1238/)
- [dune#10324](https://github.com/ocaml/dune/pull/10324)

## Description

The PR adds a new `select-dune-context` command that will show the list of available Dune contexts for the user to choose from. Once a selection is done, the setting will be stored in a new config `ocaml.dune.context`.

Every time the value of the context changes, the LSP server will be killed and a new instance will be created, passing `--context` flag as param (this param is added in the PR referred above).

## Demo

In the demo below, we are switching between a context `melange` and the `default` one. One can observe how all the features in the editor like go to definition, codelenses or PPXs will be shown from the perspective of the currently selected context:

editor-melange-context

## Future improvements

Right now, only a single instance of the LSP server is running, which means that for files that are not in the currently selected context, the editor will show the infamous `No config found for file foo.ml. Try calling 'dune build'` error. In the future, we can improve this by runnning as many LSP server instances as Dune contexts exist, so that the VSCode extension can inspect the result of each, and choose the data from the one that is not failing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+- Allow selection of Dune context (#1449)
+
 ## 1.18.0
 
 - Fallback to `ocaml.org` package documentation page if there is no doc field in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 - Allow selection of Dune context (#1449)
 
+## 1.18.1
+
+- Fix `sideEffects` field in `astexplorer/package.json` (#1427)
+
 ## 1.18.0
 
 - Fallback to `ocaml.org` package documentation page if there is no doc field in
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ yarn-deps:
 .PHONY: deps
 deps: ## Install development dependencies
 	$(MAKE) yarn-deps
-	opam install --deps-only --with-test --with-doc --with-dev-setup --yes .
+	opam install --deps-only --with-test --with-doc --yes .
 
 .PHONY: create_switch
 create_switch:
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ yarn-deps:
 .PHONY: deps
 deps: ## Install development dependencies
 	$(MAKE) yarn-deps
-	opam install --deps-only --with-test --with-doc --yes .
+	opam install --deps-only --with-test --with-doc --with-dev-setup --yes .
 
 .PHONY: create_switch
 create_switch:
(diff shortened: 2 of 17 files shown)


jchavarri:
I talked to @andreypopp and he suggested to remove the `duneContexts` protocol command, and instead call `dune describe contexts` directly from the vscode extension. I applied these changes in https://github.com/ocamllabs/vscode-ocaml-platform/pull/1449/commits/9d348689874890487734bb20b02d1daeb282d7c2 and its companion commit https://github.com/ocaml/ocaml-lsp/pull/1238/commits/9ada8752ddf9fefd220699fcb530491df7fff06d.

smorimoto:
Is this still active?

smorimoto:
Close this for now, but feel free to re-open.

PR #1465, opened 2024-05-03
Odoc markup

dlesbre:
Add syntax color to:
- odoc tables (added in odoc 2.3), 
- odoc headings with labels (`{1:label heading}`) 
- code block with attributes (used by [mdx](https://ocaml.org/p/mdx/latest)).
- also color `{@ocaml attr[` blocks as ocaml instead of verbatim (previously only `@{[` block were colored as ocaml).

Here is an example test comment:
```ocaml
(**

{1:label section}

{table
 {tr
  {td {b hello}}
  {th heading}
 }
}

{t
  | hello | there | header | text |
  | --- | :-- | :---: | --:
  | a | b | c | d
}

{@ocaml skip[
   let x = 5
]}

{@notocaml someattribute[
    this is just colored as verbatim
]}

*)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # Unreleased
 
 - Add new syntax patterns for dune(-project) files (#1391)
+- Add syntax color to odoc tables and headings with labels (#1465)
 
 ## 1.18.1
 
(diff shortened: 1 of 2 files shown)


rgrinberg:
@mnxn could you have a look?

mnxn (approved):
Sorry for the delay. This looks good.

PR #1517, opened 2024-06-24
Fix `variable.other.declaration.dune-project`

smorimoto:


mnxn (commented):
Is it correct to highlight all ASCII characters? I don't think they are all legal in dependency names.

Also, the rule on line 191 needs to be updated with the same changes.

smorimoto:
As long as Git supports all ASCII characters, there should be no problem, but I need to check. CC: @dra27 @rjbou

rgrinberg:
Just a note: some characters are indeed forbidden in package names. For example, `.` is not allowed in names

smorimoto:
Just updated changes to add `_` to allow list instead for now.

PR #1582, opened 2024-09-03
feat: Copy type under cursor

xvw:
The following patch add the command `copy-type-under-cursor` that copy, in the clipboard, the computed type under the cursor. The feature relay on the `TypeEnclosing` custom request.

### DIFF
--- a/src/custom_requests.mli
+++ b/src/custom_requests.mli
@@ -20,3 +20,8 @@ val switchImplIntf : (string, string array) custom_request
 val inferIntf : (string, string) custom_request
 
 val typedHoles : (Uri.t, Range.t list) custom_request
+
+val typeEnclosing :
+  ( Uri.t * [ `Range of Range.t | `Position of Position.t ] * int * int
+  , int * string * Range.t list )
+  custom_request
--- a/package.json
+++ b/package.json
@@ -250,6 +250,11 @@
         "command": "ocaml.goto-closure-code-location",
         "category": "OCaml",
         "title": "Goto Closure Code Location"
+      },
+      {
+        "command": "ocaml.copy-type-under-cursor",
+	"category": "OCaml",
+	"title": "Copy Type Under Cursor"
       }
     ],
     "configuration": {
@@ -1034,6 +1039,10 @@
         {
           "command": "ocaml.goto-closure-code-location",
           "when": "false"
+        },
+	{
+          "command": "ocaml.copy-type-under-cursor",
+          "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex"
         }
       ],
       "editor/title": [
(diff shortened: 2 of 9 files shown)


src/custom_requests.mli
```diff
@@ -20,3 +20,8 @@ val switchImplIntf : (string, string array) custom_request
 val inferIntf : (string, string) custom_request
 
 val typedHoles : (Uri.t, Range.t list) custom_request
+
+val typeEnclosing :
+  ( Uri.t * [ `Range of Range.t | `Position of Position.t ] * int * int
+  , int * string * Range.t list )
```
voodoos, line 26: These types are not very informative... would it be possible / convenient to use records ?

rgrinberg:
@mnxn or @ulugbekna could you try this out?

rgrinberg:
I'll test this out today in the absence of interest from anyone else. @Alizter do you want to give this a try perhaps? Some sanity testing from a vscode user would be appreciated.

rgrinberg:
Okay, I tested this out. Works as advertised. Some feedback:

1. It's hard to know what was actually copied and what was the "effective" identifier the cursor. I suggest you include this information in the "Type copied" message. E.g. `Copied type of X: Y`. `Y` might need truncation if it's larger.

2. The command works incorrectly under a selection even though the underlying custom request supports selection. For example, given the selection of `String.length foo`, I expected the type to be copied to be `int`. At the moment, it's `string` (if the selection is done left to right).

xvw:
> Okay, I tested this out. Works as advertised. Some feedback:
> 
>     1. It's hard to know what was actually copied and what was the "effective" identifier the cursor. I suggest you include this information in the "Type copied" message. E.g. `Copied type of X: Y`. `Y` might need truncation if it's larger.
> 
>     2. The command works incorrectly under a selection even though the underlying custom request supports selection. For example, given the selection of `String.length foo`, I expected the type to be copied to be `int`. At the moment, it's `string` (if the selection is done left to right).

Thanks, I've change the default behaviour of the command in https://github.com/ocamllabs/vscode-ocaml-platform/pull/1582/commits/508b3753b78e5ac77c64ceb587f61f945df4d4ac 

Is that better?

rgrinberg:
Good to go now. I would have liked the message to include the identifier that's under the cursor as well.  It's not obvious what has been typed if you put the cursor over something like a let binding for example. This is certainly usable already though.

xvw:
> hat has been typed if you put the cursor over something like a let binding for example. This is certainly usable already though.

Thanks a lot for the merge. 
I have make some attempt dealing with the "identifier" but in fact, in many many cases, the identifier (or the range) was bloating the notification box.

rgrinberg:
Yeah, I imagined. Maybe you could flash the enclosing selection under the cursor briefly after copying? I think Emacs did something like that.

xvw:
> Yeah, I imagined. Maybe you could flash the enclosing selection under the cursor briefly after copying? I think Emacs did something like that.

Not in my experiences, it just print `Copied $t in the kill-ring`, we can fill an issue to improve the visual feedback of the command btw?

PR #1591, opened 2024-09-11
chore(nix): use newest yarn

rgrinberg:
yarn 1.x no longer works


smorimoto:
We should use corepack instead.

rgrinberg:
If you know nix, I'm happy to take your lead on this one. I just did the quickest thing that got the project to build.

PR #1621, opened 2024-10-17
Expose Node bindings for public use

giltho:
Hello,

I'm using this, your bindings are actually pretty helpful! I do need the Node bindings that are not yet expose, so I forked it, but I thought it might as well be merged.


rgrinberg:
Wouldn't it be better to put this in a separate package?

giltho:
I created its own package at first, but I was thinking that there are already two attempts at binding node from ocaml ([here](https://github.com/besport/ocaml-node) and [here](https://github.com/tmattio/ocaml-node)), and if one wants to ever publish one of those on opam, this could still be published under vscode.node. The advantage and that someone writing bindings for vscode would know these bindings interact well since they are actively used here.
But I'm possibly over-thinking it, happy to move it out to its own package.

rgrinberg:
The main advantage of having a separate package is that users of this library will not need to build and install a bunch of unrelated stuff we use just for vscode.

You could call the package `vscode-node` if you'd like to avoid potential conflicts.

giltho:
Hmm, the issue is that `node` currently depends on `interop` which is exposed as `vscode.interop`, so `node` would depend on `vscode`. Should I also create a package called vscode-interop?

rgrinberg:
You could put it under `vscode-node.interop` if you'd like. Why do the node bindings depend on vscode interop though?

giltho:
It's used for:
- `Class.Make`
- `Dict`
- `or_undefined`
- `Interface.Make`

rgrinberg:
I see.`vscode-interop.interop` sounds like the right way to go

giltho:
Alright done :) And it works on my machine

smorimoto (approved):

PR #1626, opened 2024-10-21
Search By Type or Polarity

PizieDust:
# Search By Type or Polarity

## Introduction
This PR implements the [Search By Type or Polarity Custom Requests](https://github.com/ocaml/ocaml-lsp/pull/1369).

## UI
The UI uses a [QuickInput](https://code.visualstudio.com/api/references/vscode-api#QuickInput) component which provides a textbox for the user to enter the search query. 
The results are displayed in a [QuickPick](https://code.visualstudio.com/api/ux-guidelines/quick-picks) component where upon selection, the value is inserted into the document at the cursor location.

The QuickInput and QuickPick components are ideal for this use case given they are lightweight and offer robust performance as compared to webviews. 
They also offer easy navigation and are suitable for both mouse and keyboard usage.

## UX
The entire workflow is done in 3 steps:
### Step 1:  Triggering the command: 
The command can be triggered using one of three ways:

   - 1. With your cursor in the editor, use the keyboard shortcut `Alt + F`

OR
   
   - 2. From the command pallet, select `OCaml: Search a value by type or polarity` 
image

OR

   - 3. Click a button from the extension treeview under the section `LSP Custom Requests`
image

### Step 2: Entering the search query
Type what you are searching for. You can use type signatures or polarities.
- `int -> string` 
- `-int +string`

image

### Step 3: Selecting a value from the results
After entering the search term and hitting enter, the results are displayed in a `QuickPick` component where you can select the specific value you are looking for.  It also includes a textbox which can be used to filter the values.
Upon selection, the value is inserted into the document at the cursor position.

#### Going back
You can use the back button to go back to the search input box and change your query. If the results are empty, it automatically takes you back to the search input box as well.

image

## Notifications
- `Ocamllsp warning`: If `ocamllsp` is not running at the time the command is invoked, this notification is displayed.
- `No support for requests`: If the version of ocamllsp doesn't publish the `ocamllsp/typeSearch` capability then this notification is displayed. 

## Commands
One command is published `ocaml.search-by-type` which triggers the process.

## Demo

https://github.com/user-attachments/assets/cba6f3ed-21d0-4278-882d-d4659be279e3

## Note
The [Search by Type PR](https://github.com/ocaml/ocaml-lsp/pull/1369) on Ocaml-lsp is unmerged, so to test pin `ocaml-lsp-server` to [https://github.com/PizieDust/ocaml-lsp.git#polarity_search](https://github.com/PizieDust/ocaml-lsp.git#polarity_search)

## References
- QuickPick : [https://code.visualstudio.com/api/ux-guidelines/quick-picks](https://code.visualstudio.com/api/ux-guidelines/quick-picks)
- QuickInput: [https://code.visualstudio.com/api/references/vscode-api#QuickInput](https://code.visualstudio.com/api/references/vscode-api#QuickInput)

cc @voodoos

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ prefix `OCaml:`:
 ![commands](./doc/commands.png)
 
 | Name                           | Description                                 | Keyboard Shortcuts |
-| -------------------------------| ------------------------------------------- | ------------------ |
+| ------------------------------ | ------------------------------------------- | ------------------ |
 | `ocaml.select-sandbox`         | Select sandbox for this workspace           |                    |
 | `ocaml.server.restart`         | Restart language server                     |                    |
 | `ocaml.open-terminal`          | Open a terminal (current sandbox)           |                    |
@@ -264,6 +264,7 @@ prefix `OCaml:`:
 | `ocaml.open-repl`              | Open REPL                                   |                    |
 | `ocaml.evaluate-selection`     | Evaluate Selection                          | `Shift+Enter`      |
 | `ocaml.copy-type-under-cursor` | Copy the type under the cursor              |                    |
+| `ocaml.search-by-type`         | Search a value by type or polarity          | `Alt+F`            |
 
 ## Debugging OCaml programs (experimental)
 
--- a/package.json
+++ b/package.json
@@ -262,6 +262,11 @@
         "command": "ocaml.jump",
         "category": "OCaml",
         "title": "List possible parent targets for jumping"
+      },
+      {
+        "command": "ocaml.search-by-type",
+        "category": "OCaml",
+        "title": "Search a value by type or polarity"
       }
     ],
     "configuration": {
@@ -776,6 +781,11 @@
         "command": "ocaml.switch-hover-mode",
         "key": "Alt+H",
         "when": "editorTextFocus && editorLangId == ocaml || editorLangId == ocaml.interface "
+      },
+      {
+        "command": "ocaml.search-by-type",
+        "key": "Alt+F",
+        "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex"
       }
     ],
     "languages": [
@@ -1066,6 +1076,10 @@
         },
         {
           "command": "ocaml.construct",
+          "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex"
+        },
+        {
+          "command": "ocaml.search-by-type",
           "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex"
         }
       ],
--- a/src/custom_requests.mli
+++ b/src/custom_requests.mli
@@ -64,3 +64,29 @@ module Merlin_jump : sig
   val make : uri:Uri.t -> position:Position.t -> params
   val request : (params, response) custom_request
 end
+
+module Type_search : sig
+  type type_search_result =
+    { name : string
+    ; typ : string
+    ; loc : Range.t
+    ; doc : MarkupContent.t option
+    ; cost : int
+    ; constructible : string
+    }
+
+  type params
+  type response = type_search_result list
+
+  val make
+    :  uri:Uri.t
+    -> position:Position.t
+    -> limit:int
+    -> query:string
+    -> with_doc:bool
+    -> ?doc_format:MarkupKind.t option
+    -> unit
+    -> params
+
+  val request : (params, response) custom_request
+end
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -753,6 +753,203 @@ module MerlinJump = struct
   ;;
 end
 
+module Search_by_type = struct
+  let extension_name = "Search By Type"
+
+  let is_valid_text_doc textdoc =
+    match TextDocument.languageId textdoc with
+    | "ocaml" | "ocaml.interface" | "reason" | "ocaml.ocamllex" -> true
+    | _ -> false
+  ;;
+
+  let ocaml_lsp_doesnt_support_search_by_type ocaml_lsp =
+    not (Ocaml_lsp.can_handle_search_by_type ocaml_lsp)
+  ;;
+
+  let get_search_results ~query ~limit ~with_doc ~position text_editor client =
+    let doc = TextEditor.document text_editor in
+    let uri = TextDocument.uri doc in
+    Custom_requests.(
+      send_request
+        client
+        Type_search.request
+        (Type_search.make ~uri ~position ~limit ~query ~with_doc ()))
+    |> Promise.catch ~rejected:(fun _ -> Promise.return [])
+  ;;
+
+  let input_box =
+    (* Re-using the same instance of the input box allows us to remember the
+       last input. *)
+    let box =
+      InputBox.set
+        (Window.createInputBox ())
+        ~title:"Search By Type"
+        ~ignoreFocusOut:false
+        ~placeholder:"int -> string / -int +string"
+        ~prompt:
+          "Perform a search by type request by providing a type signature to look for"
+        ()
+    in
+    let _disposable =
+      InputBox.onDidChangeValue
+        box
+        ~listener:(fun _ -> InputBox.set_validationMessage box None)
+        ()
+    in
+    box
+  ;;
+
+  let rec display_search_results query results text_editor position client =
+    let format_doc (doc : MarkupContent.t option) =
+      match doc with
+      | Some doc -> doc.value
+      | None -> ""
+    in
+    let quickPickItems =
+      List.map results ~f:(fun (res : Custom_requests.Type_search.type_search_result) ->
+        QuickPickItem.create
+          ~label:res.name
+          ~description:res.typ
+          ~detail:(format_doc res.doc)
+          ())
+    in
+    let module QuickPick = Vscode.QuickPick.Make (QuickPickItem) in
+    let quickPick =
+      QuickPick.set
+        (Window.createQuickPick (module QuickPickItem) ())
+        ~title:"Type/Polarity Search Results"
+        ~activeItems:[]
+        ~busy:false
+        ~enabled:true
+        ~placeholder:"Select an item to insert it to the editor"
+        ~selectedItems:[]
+        ~ignoreFocusOut:false
+        ~items:quickPickItems
+        ~buttons:[ Window.quickInputButtonBack ]
+        ()
+    in
+    let _disposable =
+      QuickPick.onDidTriggerButton
+        quickPick
+        ~listener:(fun _ -> show_query_input text_editor client)
+        ()
+    in
+    let _disposable =
+      QuickPick.onDidAccept
+        quickPick
+        ~listener:(fun () ->
+          match QuickPick.selectedItems quickPick with
+          | Some (item :: _) ->
+            let value = QuickPickItem.label item in
+            let _ =
+              Vscode.TextEditor.edit
+                text_editor
+                ~callback:(fun ~editBuilder ->
+                  Vscode.TextEditorEdit.insert editBuilder ~location:position ~value)
+                ()
+            in
+            QuickPick.hide quickPick
+          | _ -> display_search_results query results text_editor position client)
+        ()
+    in
+    let _disposable =
+      QuickPick.onDidHide quickPick ~listener:(fun () -> QuickPick.dispose quickPick)
+    in
+    QuickPick.show quickPick
+
+  and show_query_input =
+    let previous : Disposable.t option ref = ref None in
+    fun ?(empty_result = false) text_editor client ->
+      let open Promise.Syntax in
+      let () =
+        match !previous with
+        | None -> ()
+        | Some disposable -> Disposable.dispose disposable
+      in
+      let _update_input_box =
+        let validationMessage =
+          if empty_result
+          then
+            Some
+              (InputBoxValidationMessage.create
+                 ~message:"No result found. Check the syntax or use a more general query."
+                 ~severity:Warning
+                 ())
+          else None
+        in
+        InputBox.set_validationMessage input_box validationMessage;
+        InputBox.set_busy input_box false;
+        InputBox.set_enabled input_box true
+      in
+      let onDidAccept_disposable =
+        InputBox.onDidAccept
+          input_box
+          ~listener:(fun () ->
+            match InputBox.value input_box with
+            | Some query ->
+              let () = InputBox.set_busy input_box true in
+              let () = InputBox.set_enabled input_box false in
+              let position = TextEditor.selection text_editor |> Selection.active in
+              ignore
+                (let+ query_results =
+                   get_search_results
+                     ~query
+                     ~with_doc:true
+                     ~limit:100
+                     ~position
+                     text_editor
+                     client
+                 in
+                 match query_results with
+                 | [] -> show_query_input ~empty_result:true text_editor client
+                 | results ->
+                   let results =
+                     List.remove_consecutive_duplicates
+                       ~which_to_keep:`First
+                       ~equal:
+                         (fun
+                           (left : Custom_requests.Type_search.type_search_result)
+                           right ->
+                         String.equal left.name right.name && left.cost = right.cost)
+                       results
+                   in
+                   display_search_results query results text_editor position client)
+            | None -> ())
+          ()
+      in
+      previous := Some onDidAccept_disposable;
+      InputBox.show input_box
+  ;;
+
+  let _search_by_type =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      match Window.activeTextEditor () with
+      | None ->
+        Extension_consts.Command_errors.text_editor_must_be_active
+          extension_name
+          ~expl:
+            "The cursor position is used to determine the correct environment and insert \
+             the result."
+        |> show_message `Error "%s"
+      | Some text_editor when not (is_valid_text_doc (TextEditor.document text_editor)) ->
+        show_message
+          `Error
+          "Invalid file type. This command only works in ocaml files, ocaml interface \
+           files, reason files or ocamllex files."
+      | Some text_editor ->
+        (match Extension_instance.lsp_client instance with
+         | None -> show_message `Warn "ocamllsp is not running"
+         | Some (_client, ocaml_lsp)
+           when ocaml_lsp_doesnt_support_search_by_type ocaml_lsp ->
+           show_message
+             `Warn
+             "The installed version of `ocamllsp` does not support type search"
+         | Some (client, _) -> show_query_input text_editor client)
+    in
+    command Extension_consts.Commands.search_by_type handler
+  ;;
+end
+
 let register extension instance = function
   | Command { id; handler } ->
     let callback = handler instance in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+- Add `ocaml.search-by-type` to search for values using their type signature (#1626)
+
 ## 1.24.0
 
 - Add `ocaml.commands.construct.recursiveCalls` setting to configure construct chaining. (#1673)
--- a/src-bindings/vscode/vscode.ml
+++ b/src-bindings/vscode/vscode.ml
@@ -178,6 +178,49 @@ module Uri = struct
   let equal a b = String.equal (toString a ()) (toString b ())
 end
 
+module LightDarkIcon = struct
+  type t =
+    { light : ([ `String of string | `Uri of Uri.t ][@js.union])
+    ; dark : ([ `String of string | `Uri of Uri.t ][@js.union])
+    }
+  [@@js]
+
+  let t_of_js js_val =
+    let light_js = Ojs.get_prop_ascii js_val "light" in
+    let dark_js = Ojs.get_prop_ascii js_val "dark" in
+    let light =
+      if Ojs.has_property light_js "parse"
+      then `Uri ([%js.to: Uri.t] light_js)
+      else `String ([%js.to: string] light_js)
+    in
+    let dark =
+      if Ojs.has_property dark_js "parse"
+      then `Uri ([%js.to: Uri.t] dark_js)
+      else `String ([%js.to: string] dark_js)
+    in
+    { light; dark }
+  ;;
+end
+
+module ThemeColor = struct
+  include Class.Make ()
+  include [%js: val make : id:string -> t [@@js.new "vscode.ThemeColor"]]
+end
+
+module ThemeIcon = struct
+  include Class.Make ()
+
+  include
+    [%js:
+      val make : id:string -> ?color:ThemeColor.t -> unit -> t
+      [@@js.new "vscode.ThemeIcon"]
+
+      val file : t [@@js.global "vscode.ThemeIcon.File"]
+      val folder : t [@@js.global "vscode.ThemeIcon.Folder"]
+      val id : t -> string [@@js.get]
+      val color : t -> ThemeColor.t or_undefined [@@js.get]]
+end
+
 module TextDocument = struct
   include Interface.Make ()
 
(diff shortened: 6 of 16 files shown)


src/extension_commands.ml
```diff
@@ -534,6 +534,116 @@ module Copy_type_under_cursor = struct
...
+          Extension_consts.Command_errors.text_editor_must_be_active
+            extension_name
+            ~expl:"The command to search for a value by it's type"
+          |> show_message `Error "%s" |> Promise.return
+        | Some _text_editor -> (
+          match Extension_instance.lsp_client instance with
+          | None ->
+            show_message `Warn "ocamllsp is not running" |> Promise.return
+          | Some (_client, ocaml_lsp)
+            when not (Ocaml_lsp.can_handle_search_by_type ocaml_lsp) ->
```
PizieDust, line 583: @voodoos This boolean check results to `true` even when I have the opam switch's version of `ocaml-lsp-server` pinned to the branch that has the `TypeSearch Custom Requests` implementation.

src/extension_commands.ml
```diff
@@ -534,6 +534,116 @@ module Copy_type_under_cursor = struct
...
+                       ~matchOnDescription:false
+                       ~canPickMany:false
+                       ~placeHolder:
+                         "Documentation information will be attached for each \
+                          result."
+                       ~ignoreFocusOut:true
+                       ())
+                  ())
+            in
+            let _get_limit_input =
```
PizieDust, line 624: At the moment, if you disable the boolean check above in the [comment](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1626/files#r1808497434) and run, only this input displays. 

To have a series of inputs we need to implement a way to run each one successively after the other while retaining the results of each step.

An example of this in typescript can be found at: https://github.com/microsoft/vscode-extension-samples/blob/main/quickinput-sample/src/multiStepInput.ts#L169

package.json
```diff
@@ -252,6 +252,11 @@
         "command": "ocaml.copy-type-under-cursor",
         "category": "OCaml",
         "title": "Copy Type Under Cursor"
+      },
+      {
+        "command": "ocaml.search-by-type",
+        "category": "OCaml",
+        "title": "Peform a search by type request"
```
voodoos, line 259:
```suggestion
        "title": "Search a value by type or polarity"
```

src/custom_requests.mli
```diff
@@ -39,3 +39,29 @@ module Type_enclosing : sig
 
   val request : (params, response) custom_request
 end
+
+module Type_search : sig
+  type tpye_search_results =
```
voodoos, line 44:
```suggestion
  type type_search_results =
```

src/custom_requests.mli
```diff
@@ -39,3 +39,29 @@ module Type_enclosing : sig
...
+    ; typ : string
+    ; loc : Range.t
+    ; doc : string option
+    ; cost : int
+    ; constructible : string
+    }
+
+  type params
+
+  type response = tpye_search_results list
```
voodoos, line 55:
```suggestion
  type response = type_search_results list
```

src/extension_commands.ml
voodoos: Does the LSP server advertise the necessary `handleTypeSearch` capability ?

voodoos: Does the LSP server advertise the necessary handleTypeSearch capability ?

PizieDust: No it doesn't. 

```json
Result: {
    "capabilities": {
        "codeActionProvider": {
            "codeActionKinds": [
                "quickfix",
                "refactor.inline",
                "construct",
                "destruct (enumerate cases)",
                "inferred_intf",
                "put module name in identifiers",
                "remove module name from identifiers",
                "remove type annotation",
                "type-annotate"
            ]
        },
        "codeLensProvider": {
            "resolveProvider": false
        },
        "completionProvider": {
            "resolveProvider": true,
            "triggerCharacters": [
                ".",
                "#"
            ]
        },
        "declarationProvider": true,
        "definitionProvider": true,
        "documentFormattingProvider": true,
        "documentHighlightProvider": true,
        "documentSymbolProvider": true,
        "executeCommandProvider": {
            "commands": [
                "ocamllsp/view-metrics",
                "ocamllsp/open-related-source",
                "ocamllsp/merlin-jump-to-target",
                "ocamllsp/show-document-text",
                "ocamllsp/show-merlin-config",
                "dune/promote"
            ]
        },
        "experimental": {
            "ocamllsp": {
                "interfaceSpecificLangId": true,
                "handleSwitchImplIntf": true,
                "handleInferIntf": true,
                "handleTypedHoles": true,
                "handleWrappingAstNode": true,
                "diagnostic_promotions": true,
                "handleHoverExtended": true,
                "handleMerlinCallCompatible": true,
                "handleTypeEnclosing": true,
                "handleGetDocumentation": true,
                "handleConstruct": true
            }
        },
```

src/extension_commands.ml
```diff
@@ -534,6 +534,116 @@ module Copy_type_under_cursor = struct
...
+        (Extension_instance.ocaml_version_exn instance)
+    with
+    | Ok () -> ()
+    | Error (`Msg msg) ->
+      show_message
+        `Warn
+        "The installed version of `ocamllsp` does not support type search. %s"
+        msg
+
+  let _get_search_results ~query ~limit ~with_doc text_editor client =
```
voodoos, line 553: It's uncommon to have names starting with a `_` unless during debugging to prevent unused warnings. (the `_copy_type` one in this file is probably an oversight)

src/extension_commands.ml
PizieDust: Yes, I just found the bug :), more of an omission

src/extension_commands.ml
voodoos: Yes, but do we really need all these ?

I expect only one input to shows: the actual query that the user needs to enter.

If we feel like we want to make other parameter configurable, we can have real options for that.

Anyway, there is an issue right now since the request is performed without waiting fro the user input.

src/extension_commands.ml
PizieDust: Okay, will correct that. I was following the convention.

src/extension_commands.ml
PizieDust:
```json
"experimental": {
            "ocamllsp": {
                "interfaceSpecificLangId": true,
                "handleSwitchImplIntf": true,
                "handleInferIntf": true,
                "handleTypedHoles": true,
                "handleWrappingAstNode": true,
                "diagnostic_promotions": true,
                "handleHoverExtended": true,
                "handleMerlinCallCompatible": true,
                "handleTypeEnclosing": true,
                "handleGetDocumentation": true,
                "handleConstruct": true,
                "handleTypeSearch": true
            }
        },
```

src/extension_commands.ml
PizieDust: This works great.

src/extension_commands.ml
```diff
@@ -635,8 +635,22 @@ module Search_by_type = struct
...
             in
-            show_message `Info "Type search has been performed"
-            |> Promise.return)
+            match (_get_query_input, _get_doc_input, _get_limit_input) with
+            | Some query, Some with_doc, Some limit ->
+              let* _type_search_results =
+                get_search_results
+                  ~query
+                  ~with_doc:
+                    (if String.equal with_doc "yes" then true else false)
```
rgrinberg, line 670: `String.equal with_doc "yes"`

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+        send_request
+          client
+          Type_search.request
+          (Type_search.make ~uri ~position ~limit ~query ~with_doc ()))
+    in
+    match res with
+    | [] ->
+      show_message `Info "Empty results";
+      [] |> Promise.return
+    | res -> remove_duplicates res |> Promise.return
```
voodoos, line 574: This function is purely utilitary, we should always remove duplicates and handle the empty case at it's call-site.

Additionally you can use `let+` so that you don't need all these `Promise.return`

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+              let* type_search_results =
+                get_search_results
+                  ~query
+                  ~with_doc:true
+                  ~limit:100
+                  ~position
+                  text_editor
+                  client
+              in
+              let* type_picker = display_search_results type_search_results in
```
voodoos, line 646: We should not display the results if the list is empty.

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+                    ~callback:(fun ~editBuilder ->
+                      TextEditorEdit.insert
+                        editBuilder
+                        ~location:position
+                        ~value:text)
+                    ()
+                in
+                if not type_inserted then
+                  show_message `Error "Unable to insert %s" text
+                else show_message `Info "Inserted %s" text
```
voodoos, line 661: For this command, messages should only appear to signal errors, the user will see in the editor that the insertion worked.

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+              | None -> show_message `Info "No selection made" |> Promise.return
```
voodoos, line 662: If the user decided to discard the picker and not make a selection, she doesn't need to be reminded of that choice.

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+              )
+            | _ ->
+              show_message `Error "Invalid input for type search"
+              |> Promise.return))
```
voodoos, line 666: My guess is that this only happens if the user decides to cancel his actions, I don't think this is an error, we should just silently do nothing.

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+              Vscode.(
+                Window.showInputBox
+                  ~options:
+                    (InputBoxOptions.create
+                       ~title:"Search By Type"
+                       ~prompt:
+                         "Perform a search by type request by providing a type \
+                          signature to look for"
+                       ~placeHolder:"int -> string"
+                       ~value:"int -> string"
```
voodoos, line 629: I don't think we want a default value here. Additionally, maybe we should advertise the polarity-based syntax too.

The place holder could show both: `int -> string / -int +string`

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+        ())
+
+  let _search_by_type =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      let search_by_type () =
+        match Window.activeTextEditor () with
+        | None ->
+          Extension_consts.Command_errors.text_editor_must_be_active
+            extension_name
+            ~expl:"The command to search for a value by it's type"
```
voodoos, line 604: Looking at the behavior of `text_editor_must_be_active`, the current `~expl`anation doesn't seem adapted:
```ocaml
module Command_errors = struct
  let text_editor_must_be_active ~expl cmd_name =
    Printf.sprintf
      "The command \"OCaml: %s\" should be run only with a file open in the \
       editor. %s"
      cmd_name
      expl
end
```

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+          (Type_search.make ~uri ~position ~limit ~query ~with_doc ()))
+    in
+    match res with
+    | [] ->
+      show_message `Info "Empty results";
+      [] |> Promise.return
+    | res -> remove_duplicates res |> Promise.return
+
+  let display_search_results results =
+    Vscode.(
```
voodoos, line 577: We can remove some indentation by using `let open Vscode in` (same latter)

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+      Window.showQuickPickItems
+        ~choices:
+          (List.map
+             results
+             ~f:(fun (res : Custom_requests.Type_search.type_search_result) ->
+               ( Vscode.QuickPickItem.create
```
voodoos, line 583:
```suggestion
               (QuickPickItem.create
```

src/extension_commands.ml
voodoos: The correct explanation is that the command needs the cursor position to deduce the environment and know were to insert the value.

package.json
```diff
@@ -1194,6 +1207,10 @@
       {
         "view": "ocaml-sandbox",
         "contents": "In order to manage your sandbox, you can open a folder containing an Opam switch, or select an Opam switch as a sandbox.\n[Open Folder](command:vscode.openFolder)\n[Select Sandbox](command:ocaml.select-sandbox)"
+      },
+      {
+        "view": "ocaml-custom-requests",
+        "contents": "Type/Polarity Search \nSearch for value by type or polarity.\n[Search by Type or Polarity](command:ocaml.search-by-type)"
```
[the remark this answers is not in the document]
PizieDust: The idea was to have all the custom requests in a dedicated section, but this change will be made in the latest commit. I think it makes more sense for all the commands to be in the command section.

src/extension_commands.ml
```diff
@@ -534,6 +534,143 @@ module Copy_type_under_cursor = struct
...
+      show_message
+        `Warn
+        "The installed version of `ocamllsp` does not support type search. %s"
+        msg
+
+  let rec remove_duplicates = function
+    | (a : Custom_requests.Type_search.type_search_result) :: (b :: _ as t) ->
+      if String.equal a.name b.name && a.cost = b.cost then remove_duplicates t
+      else a :: remove_duplicates t
+    | unique -> unique
```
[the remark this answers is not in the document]
PizieDust: The results appear to always be sorted by cost but I added a sort just in case. wdyt?

README.md
```diff
@@ -263,6 +263,7 @@ prefix `OCaml:`:
 | `ocaml.open-repl`              | Open REPL                                   |                    |
 | `ocaml.evaluate-selection`     | Evaluate Selection                          | `Shift+Enter`      |
 | `ocaml.copy-type-under-cursor` | Copy the type under the cursor              |                    |
+| `ocaml.search-by-type`         | Search a value by type or polarity          |                    |
```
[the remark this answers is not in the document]
PizieDust: how about <kbd>Ctrl + Alt + p</kbd> ?
.
.
.
p for polarity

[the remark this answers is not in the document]
voodoos: "Polarity" is probably the keyword related to that feature that user are less familiar with. I think `S` for search or `F` for find would be better. Most users are used to `Ctrl+F` and `Ctrl+Maj+F` to Find things. Maybe we could go with a chord like `Ctrl-K Ctrl-F`? 

@smorimoto do you have some knowledge about how best we can choose these ?

src/extension_commands.ml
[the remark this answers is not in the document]
voodoos: The list, is, by essence, sorted by cost, so there is no need sorting it again with this criteria.

src/extension_commands.ml
```diff
@@ -643,27 +628,39 @@ module Search_by_type = struct
...
-                    ~callback:(fun ~editBuilder ->
-                      TextEditorEdit.insert
-                        editBuilder
-                        ~location:position
-                        ~value:text)
-                    ()
+              match type_search_results with
+              | [] ->
+                show_message `Info "Empty type/polarity search results";
+                search_by_type ~query ()
```
voodoos, line 634: There is two parts in the large `search_by_type` function:
1. Check that the `text_editor` and  the server are compatible with the feature
2. Use UI element to interact with the users and provide feedback

When the results are empty, or when we restart the workflow for whatever reason, we don't need to redo the checks already made at step 1. It would make the code clearer, and the behavior simpler, to extract phase 2 to a separate function that will be the one looping until the user is satisfied.

The part to extract would be:
```
            let position =
              TextEditor.selection text_editor |> Selection.active
            in
            let* query_input = get_query_input ~previous_query:query () in
            match query_input with
            | Some query -> (
              let* type_search_results =
                get_search_results
                  ~query
                  ~with_doc:true
                  ~limit:100
                  ~position
                  text_editor
                  client
              in
              match type_search_results with
              | [] ->
                show_message `Info "Empty type/polarity search results";
                search_by_type ~query ()
              | results -> (
                let* type_picker =
                  display_search_results
                    (remove_duplicates
                       ~cmp:(fun
                           (left :
                             Custom_requests.Type_search.type_search_result)
                           right
                         ->
                         String.equal left.name right.name
                         && left.cost = right.cost)
                       results)
                in
                match type_picker with
                | Some text ->
                  let+ type_inserted =
                    TextEditor.edit
                      text_editor
                      ~callback:(fun ~editBuilder ->
                        TextEditorEdit.insert
                          editBuilder
                          ~location:position
                          ~value:text)
                      ()
                  in
                  if not type_inserted then
                    show_message `Error "Unable to insert %s" text
                | None -> Promise.return ()))
```

I don't have a good name for that function right now. Something like `start_user_interaction`,  or `wait_for_user_input_and_display_results`.

README.md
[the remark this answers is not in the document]
voodoos: Maybe we could have chords that start with O for OCaml ? I don't know if that's allowed. Could be `Ctrl-O Ctrl-F`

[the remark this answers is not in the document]
voodoos: Or simply `Alt+F` looks like it's available.

src/extension_commands.ml
```diff
@@ -534,6 +534,180 @@ module Copy_type_under_cursor = struct
...
+  let extension_name = "Search By Type"
+
+  let ocaml_lsp_doesnt_support_search_by_type ocaml_lsp =
+    not (Ocaml_lsp.can_handle_search_by_type ocaml_lsp)
+
+  let rec _remove_duplicates ~cmp = function
+    | a :: (b :: _ as t) ->
+      if cmp a b then _remove_duplicates ~cmp t
+      else a :: _remove_duplicates ~cmp t
+    | remaining -> remaining
```
voodoos, line 547: Ah, I must have missed it the first time I looked, but there is a utility to do that in the `List` module (that comes from `Base`): `List.remove_consecutive_duplicates`

src/extension_commands.ml
```diff
@@ -534,6 +534,193 @@ module Copy_type_under_cursor = struct
...
+          QuickPickItem.create
+            ~label:res.name
+            ~description:res.typ
+            ~detail:(Option.value ~default:"" res.doc)
+            ())
+    in
+    let module QuickPick = Vscode.QuickPick.Make (QuickPickItem) in
+    let quickPick =
+      QuickPick.set
+        (Window.createQuickPick (module QuickPickItem) ())
```
voodoos, line 610: Do we really need functors and first class modules to make a binding for `QuickPick` and `createQuickPick` ?  I suspect there is a simpler approach.

PizieDust: So the api indicates that we need to give it a type when invoking it (`QuickPick`)
https://code.visualstudio.com/api/references/vscode-api#QuickPick&lt;T&gt;. 
Looking through the codebase functors are what was used to write the bindings

src/extension_commands.ml
```diff
@@ -534,6 +534,193 @@ module Copy_type_under_cursor = struct
...
+  let _search_by_type =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      let search_by_type () =
+        match Window.activeTextEditor () with
+        | None ->
+          Extension_consts.Command_errors.text_editor_must_be_active
+            extension_name
+            ~expl:
+              "Search by type/polarity can only be run with a valid file open \
+               in the editor."
```
voodoos, line 703:
```suggestion
              "The cursor position is used to determine the correct \
                environment and insert the result."
```

src/extension_commands.ml
```diff
@@ -534,6 +534,193 @@ module Copy_type_under_cursor = struct
...
+
+  let rec display_search_results query results text_editor position client =
+    let quickPickItems =
+      List.map
+        results
+        ~f:(fun (res : Custom_requests.Type_search.type_search_result) ->
+          QuickPickItem.create
+            ~label:res.name
+            ~description:res.typ
+            ~detail:(Option.value ~default:"" res.doc)
```
voodoos, line 604: Can't we pass the option directly ?

src/extension_commands.ml
```diff
@@ -534,6 +534,193 @@ module Copy_type_under_cursor = struct
...
+        ~items:quickPickItems
+        ~buttons:[ Window.quickInputButtonBack ]
+        ()
+    in
+    let _disposable =
+      QuickPick.onDidTriggerButton
+        quickPick
+        ~listener:(fun _ ->
+          let _ = handle_search ~previous_query:query text_editor client in
+          ())
```
voodoos, line 627:
```suggestion
        ~listener:(fun _ -> handle_search ~previous_query:query text_editor client)
```

src/extension_commands.ml
```diff
@@ -534,6 +534,193 @@ module Copy_type_under_cursor = struct
...
+                ~callback:(fun ~editBuilder ->
+                  Vscode.TextEditorEdit.insert
+                    editBuilder
+                    ~location:position
+                    ~value)
+                ()
+            in
+            QuickPick.hide quickPick
+          | _ ->
+            display_search_results query results text_editor position client)
```
voodoos, line 649: We could just do nothing here (`()`) and it will keep the results open. I also guess this case should never happen ?

voodoos (commented):
Thanks Pizie, that looks very good !

I made a few tests and there are still a few non-optimal behaviors:

1. When the query fails, the box remains busy forever (until the user clicks elesewhere). For example, calling type-search in the package JSON file causes this server error:
```
[Trace - 10:22:28] Received response 'ocamllsp/typeSearch - (103)' in 1ms. Request failed: no document found with uri: file:///Users/ulysse/tmp/vscode-ocaml-platform/package.json (-32600).
```
Hopefully there is a way to catch that and tell the user ?
Additionally, is there a way to prevent the command to be called in non-ocaml files ?

2. Ideally, we would always show the previous answer when reopening the prompt. (I am working on this one 🙂).

src/extension_commands.ml
```diff
@@ -534,6 +534,215 @@ module Copy_type_under_cursor = struct
     command Extension_consts.Commands.copy_type_under_cursor handler
 end
 
+module Search_by_type = struct
+  let extension_name = "Search By Type"
+
+  let is_valid_file file_uri =
```
PizieDust, line 540: @voodoos here we perform the check if it's a valid file
will this be sufficient for the purpose?

voodoos: I don't think checking the extension is a good idea. You should use the "language id" from the TextDocument that: https://vscode-api.js.org/interfaces/vscode.TextDocument.html#languageId

It's surprising that vscode doesn't always enforce the `when` clause in the package.json:
`"when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex"`

src/extension_commands.ml
```diff
@@ -555,7 +556,8 @@ module Search_by_type = struct
 
   let input_box =
     (* Re-using the same instance of the input box allows us to remember the
-       last input. *)
+       last input. show_message `Error "Invalid file type. This command can only
+       work in a valid \ Ocaml .ml or .mli file" *)
```
voodoos, line 560:
```suggestion
       last input. *)
```

src/extension_commands.ml
```diff
@@ -534,6 +534,212 @@ module Copy_type_under_cursor = struct
...
+                    editBuilder
+                    ~location:position
+                    ~value)
+                ()
+            in
+            QuickPick.hide quickPick
+          | _ ->
+            display_search_results query results text_editor position client)
+        ()
+    in
```
voodoos, line 633: I think we should also `dispose` the quickpick when `onDidHide` to be on the safe side.

PizieDust:
> Thanks @PizieDust, this look very good.
> 
> There is one last thing that would be nice to have: the current documentation is shown in ocamldoc syntax which is not always ideal. There is two way to fix that:
> 
> 1. Add the possibility to specify the returned format of the documentation to the custom request, simlilarly as what was done for the "get_documentation" syntax. We could have multiple options: ocamldoc, markdown, html or plaintext. @xvw would that be useful for other clients too ?
> 2. Use odoc-parser in the vscode plugin to reformat. It's easy for this client because it is written in OCaml, but more complex for other clients.
> 
> What do you think ?

Yes indeed having a new field to specify the format of the documentation will be a cool addition. 
I will update the PR in `ocaml-lsp` and then this one.

xvw:
From my POV, Markdown seems "nice" for documentation!

xvw (approved):

xvw:
(but since https://github.com/ocaml/ocaml-lsp/pull/1369 is not merged, the PR should be flagged as Draft?)

voodoos:
The new custom request has now been merged and released, we can rebase and undraft this PR @PizieDust  :-)

smorimoto:
Could you review this? @voodoos

smorimoto (approved):

voodoos:
I will do one last pass tomorrow !

voodoos (approved):

PR #1642, opened 2024-11-11
Make it a warning if ocamlc is missing

gridbugs:
If you're using the dune developer preview then ocamlc won't be globally installed. It's a bit offputting to have the plugin pop up with an error every time vscode starts, especially since the plugin seems to work fine without it. This downgrades it to a warning and adds an "Ignore" button.

### DIFF
--- a/src/extension_instance.ml
+++ b/src/extension_instance.ml
@@ -372,7 +372,7 @@ let update_ocaml_info t =
     | `Ocamlc_missing ->
       let (_ : unit Promise.t) =
         let+ maybe_choice =
-          Window.showErrorMessage
+          Window.showWarningMessage
             ~message:
               "OCaml bytecode compiler `ocamlc` was not found in the current \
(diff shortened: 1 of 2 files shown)


src/extension_instance.ml
```diff
@@ -385,6 +385,7 @@ let update_ocaml_info t =
                         ~args:[]
                     in
                     () )
+              ; ("Ignore", fun _ -> ())
```
smorimoto, line 388: I don't think this is needed because there is a default close button.

gridbugs: You're correct, it's not needed. I intentionally added it because I thought it might reassure people that this warning isn't a big deal. But happy to remove it if that's the consensus.

smorimoto: If you can remove, it would be nice 🙂

gridbugs: Done!

smorimoto (approved):

PR #1646, opened 2024-11-14
Construct custom request implementation

PizieDust:
cc @voodoos

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -534,6 +534,108 @@ module Copy_type_under_cursor = struct
     command Extension_consts.Commands.copy_type_under_cursor handler
 end
 
+module Construct = struct
+  let extension_name = "Construct"
+
+  let is_valid_text_doc textdoc =
+    match TextDocument.languageId textdoc with
+    | "ocaml" | "ocaml.interface" | "reason" | "ocaml.ocamllex" -> true
+    | _ -> false
+
+  let ocaml_lsp_doesnt_support_construct ocaml_lsp =
+    not (Ocaml_lsp.can_handle_construct ocaml_lsp)
+
+  let get_construct_results position text_editor client =
+    let doc = TextEditor.document text_editor in
+    let uri = TextDocument.uri doc in
+    Custom_requests.(
+      send_request
+        client
+        Construct.request
+        (Construct.make ~uri ~position ~depth:None ~with_values:None ()))
+
+  let rec display_results ~text_editor:_ ~position:_
+      (results : Custom_requests.Construct.response) =
+    let quickPickItems =
+      List.map results.result ~f:(fun res ->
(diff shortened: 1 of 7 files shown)


src/extension_commands.ml
```diff
@@ -534,6 +556,105 @@ module Copy_type_under_cursor = struct
...
+    let open Promise.Syntax in
+    let* res = get_construct_results position text_editor client in
+    let* selected_result = display_results ~text_editor ~position res in
+    match selected_result with
+    | Some (value, range) -> (
+      let* value_inserted = insert_to_document text_editor range value in
+      match value_inserted with
+      | true ->
+        let* new_range =
+          Holes_commands.hole_position text_editor client `Prev
```
voodoos, line 603: This will go to the last hole if there are multiple ones in the result, which is not very natural. I think you should have ` Holes_commands.hole_position` take a position as parameter and not a text_editor. That way you can call it with a position that corresponds to the beginning of the inserted range:

```
Holes_commands.hole_position (Range.start range) client `Next
```

PizieDust: Yes indeed it goes to the last hole before coming to the hole before it

PizieDust: `Hole_commands.hole_position` calls ocamllsp to get the list of hole positions, which is why it's receiving `text_editor` and `client`

src/treeview_commands.ml
```diff
@@ -43,7 +43,31 @@ let terminal_item =
...
-let items = [ select_sandbox_item; terminal_item ]
+
+  let construct_item =
+    let icon =
+      `LightDark
+        Vscode.TreeItem.LightDarkIcon.
+          { light = `String (Path.asset "collection-light.svg" |> Path.to_string)
+          ; dark = `String (Path.asset "collection-dark.svg" |> Path.to_string)
+          }
+    in
```
voodoos, line 54: That's not a great icon. Here is a suggestion, but feel free to chose [another one](https://code.visualstudio.com/api/references/icons-in-labels#icon-listing):
```suggestion
    let icon = `ThemeIcon (Vscode.ThemeIcon.make ~id:"tools" ()) in
```

src/treeview_commands.ml
```diff
@@ -43,7 +43,31 @@ let terminal_item =
...
+    let label =
+      `TreeItemLabel
+        (Vscode.TreeItemLabel.create ~label:"Construct" ())
```
voodoos, line 57: We should provide a better description as a command name, something like "List values that can fill the selected typed-hole`. This should be done in other places too.

voodoos (commented):
Thanks Pizie, it's already in a very good state :-)

voodoos (approved):
Thanks for the changes, this looks good to go!

PR #1648, opened 2024-11-18
Refactor

smorimoto:
- remove unused fields
- update dune to 3.16
  - newer dune has the functionality to detect unused fields

### DIFF
--- a/src/ocaml_lsp.ml
+++ b/src/ocaml_lsp.ml
@@ -40,16 +40,14 @@ end
 
 module Experimental_capabilities = struct
   type t =
-    { interfaceSpecificLangId : bool
-    ; handleSwitchImplIntf : bool
+    { handleSwitchImplIntf : bool
     ; handleInferIntf : bool
     ; handleTypedHoles : bool
     ; handleTypeEnclosing : bool
     }
 
   let default =
-    { interfaceSpecificLangId = false
-    ; handleSwitchImplIntf = false
+    { handleSwitchImplIntf = false
     ; handleInferIntf = false
     ; handleTypedHoles = false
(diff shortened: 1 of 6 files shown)


src/ocaml_lsp.ml
```diff
@@ -63,17 +59,10 @@ module Experimental_capabilities = struct
...
-      let handleInferIntf = has_capability "handleInferIntf" in
       let handleTypedHoles = has_capability "handleTypedHoles" in
       let handleTypeEnclosing = has_capability "handleTypeEnclosing" in
-      { interfaceSpecificLangId
-      ; handleSwitchImplIntf
-      ; handleInferIntf
-      ; handleTypedHoles
-      ; handleTypeEnclosing
-      }
+      { handleSwitchImplIntf; handleTypedHoles; handleTypeEnclosing }
```
voodoos, line 65: Actually, these probably means errors somewhere else since they should be used to check a feature availability before using it.

And there is indeed a copy-pasting error line 241:
```ocaml
let can_handle_infer_intf t = t.experimental_capabilities.handleSwitchImplIntf
```

It should be `t.experimental_capabilities.handleInferIntf`.

`interfaceSpecificLangId` seems to be a remnant of the past however, it looks like it was used at some point to check the lsp version ?

smorimoto: It certainly seems to have been introduced for that purpose. But I think it's no longer necessary, so I think it can be removed from the LSP side as well.

- https://github.com/ocaml/ocaml-lsp/pull/189
- https://github.com/ocamllabs/vscode-ocaml-platform/pull/209

voodoos (approved):

PR #1666, opened 2024-12-02
Navigating Typed Holes

PizieDust:
# Navigating Typed Holes

## Introduction
This PR implements navigation of typed holes.

## UI
Results are displayed in a [QuickPick](https://code.visualstudio.com/api/ux-guidelines/quick-picks) component where upon selection, the cursor is moved to the selected typed hole.

## UX
This command can be triggered from the command pallet "Navigate to different typed holes"

## Notifications
- `Ocamllsp warning`: If `ocamllsp` is not running at the time the command is invoked, this notification is displayed.
- `No support for requests`: If the version of ocamllsp doesn't publish the `ocamllsp/typedHoles` capability then this notification is displayed. 

## Commands
One command is published `ocaml.navigate-typed-holes` which triggers the process.

## Demo

https://github.com/user-attachments/assets/b14aa44e-ea00-494c-9164-83464bee0f05

## References
- QuickPick : [https://code.visualstudio.com/api/ux-guidelines/quick-picks](https://code.visualstudio.com/api/ux-guidelines/quick-picks)

cc @voodoos cc @awilliambauer

### DIFF
--- a/package.json
+++ b/package.json
@@ -267,6 +267,11 @@
         "command": "ocaml.search-by-type",
         "category": "OCaml",
         "title": "Search a value by type or polarity"
+      },
+      {
+        "command": "ocaml.navigate-typed-holes",
+        "category": "OCaml",
+        "title": "Lists typed holes in the file for navigation."
       }
     ],
     "configuration": {
@@ -310,6 +315,11 @@
           "type": "boolean",
           "default": false,
           "markdownDescription": "Enable/Disable syntax documentation"
+        },
+        "ocaml.commands.typedHoles.constructAfterNavigate": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "When enabled, list values that can fill a typed hole after navigating to it."
         },
         "ocaml.commands.construct.recursiveCalls": {
           "type": "boolean",
@@ -1081,6 +1091,10 @@
         {
           "command": "ocaml.search-by-type",
           "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason || editorLangId == ocaml.ocamllex"
+        },
+        {
+          "command": "ocaml.navigate-typed-holes",
+          "when": "editorLangId == ocaml || editorLangId == ocaml.interface || editorLangId == reason"
         }
       ],
       "editor/title": [
--- a/package.json
+++ b/package.json
@@ -271,7 +271,7 @@
       {
         "command": "ocaml.navigate-typed-holes",
         "category": "OCaml",
-        "title": "Lists typed holes in the file for navigation."
+        "title": "List typed holes in the file for navigation."
       }
     ],
     "configuration": {
--- a/package.json
+++ b/package.json
@@ -271,7 +271,7 @@
       {
         "command": "ocaml.navigate-typed-holes",
         "category": "OCaml",
-        "title": "List typed holes in the file for navigation."
+        "title": "List typed holes in the file for navigation"
       }
     ],
     "configuration": {
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -993,15 +993,9 @@ module Navigate_holes = struct
     move_to_hole range text_editor
   ;;
 
-  let display_results
-        current_position
-        (results : Range.t list)
-        text_editor
-        client
-        instance
-    =
+  let display_results (results : Range.t list) text_editor client instance =
     let open Promise.Syntax in
-    let selected_item = ref None in
+    let selected_item = ref false in
     let text_document = TextEditor.document text_editor in
     let quickPickItems =
       List.map results ~f:(fun res ->
@@ -1062,7 +1056,7 @@ module Navigate_holes = struct
                   | Some range ->
                     let range = Range.t_of_json range in
                     let* () = jump_to_hole range text_editor in
-                    selected_item := Some range;
+                    selected_item := true;
                     QuickPick.hide quickPick;
                     (match
                        Settings.(get server_typedHolesConstructAfterNavigate_setting)
@@ -1080,31 +1074,33 @@ module Navigate_holes = struct
         ()
     in
     let _disposable =
+      let initial_selection = TextEditor.selection text_editor in
       QuickPick.onDidHide
         quickPick
         ~listener:(fun () ->
-          match !selected_item with
-          | None ->
+          if !selected_item
+          then ()
+          else
             ignore
-              (let* () =
-                 jump_to_hole
-                   (Range.makePositions ~start:current_position ~end_:current_position)
-                   text_editor
-               in
-               QuickPick.dispose quickPick |> Promise.return)
-          | Some _ -> QuickPick.dispose quickPick)
+              (TextEditor.set_selection text_editor initial_selection;
+               TextEditor.revealRange
+                 text_editor
+                 ~range:(Selection.to_range initial_selection)
+                 ~revealType:TextEditorRevealType.InCenterIfOutsideViewport
+                 ());
+          QuickPick.dispose quickPick)
         ()
     in
     QuickPick.show quickPick
   ;;
 
-  let handle_hole_navigation current_position text_editor client instance =
+  let handle_hole_navigation text_editor client instance =
     let open Promise.Syntax in
     let doc = TextEditor.document text_editor in
     let+ hole_positions = send_request_to_lsp client doc in
     match hole_positions with
     | [] -> show_message `Info "No typed holes found in the file."
-    | holes -> display_results current_position holes text_editor client instance
+    | holes -> display_results holes text_editor client instance
   ;;
 
   let _holes =
@@ -1131,8 +1127,7 @@ module Navigate_holes = struct
              `Warn
              "The installed version of `ocamllsp` does not support typed hole navigation"
          | Some (client, _) ->
-           let current_position = TextEditor.selection text_editor |> Selection.active in
-           let _ = handle_hole_navigation current_position text_editor client instance in
+           let _ = handle_hole_navigation text_editor client instance in
            ())
     in
     command Extension_consts.Commands.navigate_typed_holes handler
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -968,14 +968,7 @@ module Navigate_holes = struct
     Custom_requests.send_request client Custom_requests.typedHoles uri
   ;;
 
-  let jump_to_hole range text_editor =
-    let open Promise.Syntax in
-    let+ _ =
-      Window.showTextDocument
-        ~document:(TextEditor.document text_editor)
-        ~preserveFocus:true
-        ()
-    in
+  let move_to_hole range text_editor =
     let new_selection =
       let anchor = Range.start range in
       let active = Range.end_ range in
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -306,6 +311,11 @@
           "default": false,
           "markdownDescription": "Enable/Disable syntax documentation"
         },
+        "ocaml.server.typedHoles.constructAfterNavigate": {
```
voodoos, line 314:
```suggestion
        "ocaml.commands.typedHoles.constructAfterNavigate": {
```

package.json
```diff
@@ -262,6 +262,11 @@
         "command": "ocaml.jump",
         "category": "OCaml",
         "title": "List possible parent targets for jumping"
+      },
+      {
+        "command": "ocaml.navigate-typed-holes",
+        "category": "OCaml",
+        "title": "Navigate to different typed holes"
```
voodoos, line 269:
```suggestion
        "title": "Lists typed holes in the file for navigation."
```

src/extension_commands.ml
```diff
@@ -766,6 +766,121 @@ module MerlinJump = struct
...
+  let _holes =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      let hole_navigation () =
+        match Window.activeTextEditor () with
+        | None ->
+          Extension_consts.Command_errors.text_editor_must_be_active
+            extension_name
+            ~expl:
+              "The cursor position is used to determine the correct \
+               environment and perform the jump."
```
voodoos, line 857: This is not the reason we need an active editor here, we need it because the request is based on the content of the editor.

src/extension_commands.ml
```diff
@@ -766,6 +766,121 @@ module MerlinJump = struct
...
+            (Range.end_ range)
+            text_editor
+            client
+            instance
+        | Some false -> Promise.return ())
+      | None -> Promise.return ())
+
+  let _holes =
+    let handler (instance : Extension_instance.t) ~args:_ =
+      let hole_navigation () =
```
voodoos, line 850: Since `hole_navigation` returns `unit` and is called right after, I don't think there is any benefit in giving it a name.
```suggestion
```

src/extension_commands.ml
```diff
@@ -766,6 +766,121 @@ module MerlinJump = struct
...
+    let* hole_positions = send_request_to_lsp client doc in
+    match hole_positions with
+    | [] ->
+      show_message `Info "No typed holes found in the file.";
+      Promise.return ()
+    | holes -> (
+      let* selected_hole = display_results holes doc in
+      match selected_hole with
+      | Some (range, ()) -> (
+        let* _ = jump_to_hole range text_editor in
```
voodoos, line 835:
```suggestion
        let* () = jump_to_hole range text_editor in
```

src/extension_commands.ml
```diff
@@ -766,6 +766,121 @@ module MerlinJump = struct
...
+        match
+          Settings.(get server_typedHolesConstructAfterNavigate_setting)
+        with
+        | Some true | None ->
+          Construct.process_construct
+            (Range.end_ range)
+            text_editor
+            client
+            instance
+        | Some false -> Promise.return ())
```
voodoos, line 845: My personal feeling is that triggering automatically construct should be off by default.

src/extension_commands.ml
```diff
@@ -766,6 +766,121 @@ module MerlinJump = struct
...
+      List.map results ~f:(fun res ->
+          let line = Position.line @@ Range.end_ res in
+          ( QuickPickItem.create
+              ~label:
+                (Printf.sprintf
+                   "Line %d: %s"
+                   line
+                   (TextLine.text @@ TextDocument.lineAt ~line text_document))
+              ()
+          , (res, ()) ))
```
voodoos, line 795: Maybe we should use the `details`of the quick pick item to show the line ?

Also, I am wondering whether it would be better to navigate as soon as the user selects a quickpick (without validating with Enter) that would allows the user to see the context around the hole more easily before making a decision.

src/treeview_commands.ml
```diff
@@ -73,7 +73,32 @@ let jump_item =
   Vscode.TreeItem.set_command item command;
   item
 
-let items = [ select_sandbox_item; terminal_item; construct_item; jump_item ]
+let navigate_holes_item =
+  let icon = `ThemeIcon (Vscode.ThemeIcon.make ~id:"arrow-up" ()) in
```
voodoos, line 77: I think the `activate-breakpoints` icon would symbolize holes nicely, what do you think ?

PizieDust: The icon looks great. It's literally holes :)

src/extension_commands.ml
smorimoto: Do you mean to pass the line number to the `~label` and the content to the `~detail`?

PizieDust: It's passing the content of the line to `~detail`.
After https://github.com/ocamllabs/vscode-ocaml-platform/pull/1626 is merged, there are some additional modules that can allow us use the low level `QuickPick` component which will improve the functionality of navigating typed holes much better.

smorimoto: Merged.

PizieDust: Thanks, I'll do the necessary updates and then this PR will be ready for merging :)

src/extension_commands.ml
```diff
@@ -950,6 +950,220 @@ module Search_by_type = struct
...
+                     with
+                     | Some true ->
+                       let+ _ =
+                         Construct.process_construct
+                           (Range.end_ range)
+                           text_editor
+                           client
+                           instance
+                       in
+                       ()
```
voodoos, line 1096:
```suggestion
                       Construct.process_construct
                         (Range.end_ range)
                         text_editor
                         client
                         instance
```

It already returns a `unit Promise.t`.

src/extension_commands.ml
```diff
@@ -950,6 +950,220 @@ module Search_by_type = struct
...
+            let range_string = QuickPickItem.description item in
+            ignore
+              (match range_string with
+               | Some r ->
+                 (match Jsonoo.try_parse_opt r with
+                  | Some range ->
+                    let range = Range.t_of_json range in
+                    let* () = jump_to_hole range text_editor in
+                    selected_item := Some range;
+                    QuickPick.hide quickPick;
```
voodoos, line 1084: I would expect the quickpick to hide without the need for this.

src/treeview_commands.ml
```diff
@@ -84,8 +84,32 @@ let type_search_item =
   item
 ;;
 
+let navigate_holes_item =
+  let icon = `ThemeIcon (Vscode.ThemeIcon.make ~id:"breakpoints-activate" ()) in
+  let label =
+    `TreeItemLabel
+      (Vscode.TreeItemLabel.create ~label:"Navigate to different typed holes" ())
```
voodoos, line 91:
```suggestion
      (Vscode.TreeItemLabel.create ~label:"Navigate between typed holes" ())
```

src/extension_commands.ml
```diff
@@ -950,6 +950,220 @@ module Search_by_type = struct
...
+                | Some range ->
+                  let range = Range.t_of_json range in
+                  move_to_hole range text_editor
+                | None -> ())
+             | _ -> ())
+          | _ -> ())
+        ()
+    in
+    let _disposable =
+      QuickPick.onDidChangeActive
```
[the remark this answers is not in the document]
PizieDust: I think what we want is `onDidChangeActive`

src/extension_commands.ml
PizieDust: for some reason this is not the behavior I observe. It doesn't close if I don't manually hide it.

src/extension_commands.ml
```diff
@@ -950,6 +950,220 @@ module Search_by_type = struct
...
+    let quickPickItems =
+      List.map results ~f:(fun res ->
+        let line = Position.line @@ Range.end_ res in
+        QuickPickItem.create
+          ~label:(Printf.sprintf "Line %d" line)
+          ~detail:
+            (Printf.sprintf
+               "%s"
+               (TextLine.text @@ TextDocument.lineAt ~line text_document))
+          ~description:(Range.json_of_t res |> Jsonoo.stringify)
```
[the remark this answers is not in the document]
PizieDust: At the moment I don't think the vscode api gives us much room to wiggle. I have to embed the range in the description so we can fetch it later when the user selects an item.

package.json
```diff
@@ -267,6 +267,11 @@
         "command": "ocaml.search-by-type",
         "category": "OCaml",
         "title": "Search a value by type or polarity"
+      },
+      {
+        "command": "ocaml.navigate-typed-holes",
+        "category": "OCaml",
+        "title": "List typed holes in the file for navigation."
```
smorimoto, line 274: to have the same format as other titles:

```suggestion
        "title": "List typed holes in the file for navigation"
```

smorimoto:
Does this work now with lsp 1.20.1? It does not work in my environment.

PizieDust:
> Does this work now with lsp 1.20.1? It does not work in my environment.

I'm not sure why. 
It works for me:

The switch I just tested it in `opam list` outputs:

```
| Package               | Version   | Description                           |
|-----------------------|-----------|---------------------------------------|
| jsonrpc               | 1.20.1    | Jsonrpc protocol implementation       |
| lsp                   | 1.20.1    | LSP protocol implementation in OCaml |
| merlin-lib            | 5.3-502   | Merlin's libraries                   |
| ocaml                 | 5.2.1     | The OCaml compiler (virtual package)  |
| ocaml-base-compiler   | 5.2.1     | Official release 5.2.1               |
| ocaml-lsp-server      | 1.20.1    | LSP Server for OCaml                 |
| ocamlformat           | 0.27.0    | Auto-formatter for OCaml code        |
| ocamlformat-lib       | 0.27.0    | OCaml Code Formatter                 |


```

src/extension_commands.ml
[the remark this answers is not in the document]
voodoos: I think we could have custom `QuickPickItem`s, but there is also a simpler way around:

When the active item change you can get the list of all items and check what is the index of the item in that list.
Since the list is in the same order as the list of holes you can then lookup the corresponding range at that index. (you might want to turn the range list into an array)

src/extension_commands.ml
```diff
@@ -950,6 +950,185 @@ module Search_by_type = struct
...
+              (let range = item.range in
+               let* () = jump_to_range range text_editor in
+               selected_item := true;
+               match Settings.(get server_typedHolesConstructAfterNavigate_setting) with
+               | Some true ->
+                 Construct.process_construct
+                   (Range.end_ range)
+                   text_editor
+                   client
+                   instance
```
voodoos, line 1075: Testing this feature highlighted an issue with `process_construct` when there is no result, instead of showing a message it displays an empty quick pick which is very confusing. We should improve that behaviour in a separate PR :-)

src/extension_commands.ml
```diff
@@ -950,6 +950,185 @@ module Search_by_type = struct
...
+    let quickPick =
+      QuickPick.set
+        (Window.createQuickPick (module QuickPickItemWithRange) ())
+        ~title:"Typed Holes"
+        ~activeItems:[]
+        ~busy:false
+        ~enabled:true
+        ~placeholder:"Use arrow keys to preview / Select to jump to it"
+        ~selectedItems:[]
+        ~ignoreFocusOut:true
```
voodoos, line 1041:
```suggestion
        ~ignoreFocusOut:false
```

This option is very disruptive for the user's flow, we should only enable it when there is a real need for it. We discussed it at some point, but I think it's not needed in the end isn't it ?

voodoos (approved):
I think this is good to go, thanks @PizieDust 

@smorimoto it should work on 1.20.1 and a large number of older version since it does not rely on a new custom request. Were you able to fix your setup ? Or is there a bug in the latest release ?

smorimoto:
Let's go as long as it works on your end!

PR #1673, opened 2024-12-05
Make construct chaining configurable

PizieDust:
This PR provides a configurable option for construct chaining. 
The default behavior let's construct automatically compute values for the next typed hole. This option can be used to turn this behavior off.

```json
"ocaml.server.construct.chainConstruct": true
```

image

cc @voodoos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+- Add `ocaml.server.construct.chainConstruct` setting to configure construct chaining. (#1673)
+
 ## 1.23.0
 
 - Add `ocaml.jump` to jump to a specific target. (#1654)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # Unreleased
 
-- Add `ocaml.server.construct.chainConstruct` setting to configure construct chaining. (#1673)
+- Add `ocaml.commands.construct.recursiveCalls` setting to configure construct chaining. (#1673)
 
 ## 1.23.0
 
--- a/package.json
+++ b/package.json
@@ -308,7 +308,7 @@
         },
         "ocaml.server.construct.chainConstruct": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "markdownDescription": "Enable/Disable chaining construct to display values for subsequent holes."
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+- Add `ocaml.server.construct.chainConstruct` setting to configure construct chaining. (#1673)
```
voodoos, line 5: I think we could have a new group to distinguish options that are used to configure the server and options that are used to configure the client's commands behavior. Maybe `ocaml.commands.construct.recursiveCalls`

package.json
```diff
@@ -306,6 +306,11 @@
           "default": false,
           "markdownDescription": "Enable/Disable syntax documentation"
         },
+        "ocaml.server.construct.chainConstruct": {
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Enable/Disable chaining construct to display values for subsequent holes."
```
voodoos, line 312:
```suggestion
          "markdownDescription": "When enabled, the construct command will execute again on the next hole after a value has been chosen."
```

vscode.opam
```diff
@@ -22,7 +22,7 @@ depends: [
   "gen_js_api" {= "1.1.2"}
   "promise_jsoo" {>= "0.3.1"}
   "jsonoo" {>= "0.2.1"}
-  "ocamlformat" {= "0.26.2" & with-dev-setup}
+  "ocamlformat" {= "0.27.0" & with-dev-setup}
```
voodoos, line 25: Why is this part of this PR ? I think you should rebase on master, the bump was made recently.

voodoos (approved):
Thanks, apart from some nitpicking on the name choices this looks good to me.

vscode.opam
smorimoto: Fixed

CHANGELOG.md
voodoos: @PizieDust did you forget this one ? After that I think the PR is ready to be merged :-)

PizieDust: Oh my bad. let me fix that

voodoos (approved):
Thanks, looks good to me

PR #1675, opened 2024-12-05
New command: Type of dynamic selection

voodoos:
This PR adds three new commands that can be used to get the type of a selection, and the ones of growing enclosings of this selection. At each step it is possible to ask for more verbose results.

The following videos show how the types of the growing selection are shown in a hover popup, with the selection growing, then shrinking, and lastly with an increased verbosity:

https://github.com/user-attachments/assets/dca1c995-45bc-45c7-adee-ce83e4dbb702

The type history is also logged to an output channel which can be configured to show up automatically and clear itself to show only the last result.

I have been through a lot of back and forth to get to this result, because it is not possible (to the extent of my knowledge) to easily disable the standard hover provider from the client side due to how the official LSP client from Microsoft behaves. I tried multiple alternatives but I know think having a configuration option to mute the output of the standard hover from the server to be the most robust one. 

## Todo

- [x] This feature relies on a new server option to temporary disable the standard hover results
- [x] Rationalize the new configuration options

cc @awilliambauer

### DIFF
--- a/src/import.ml
+++ b/src/import.ml
@@ -253,3 +253,31 @@ module Ocaml_version = struct
 
   include Ocaml_version
 end
+
+exception User_error of string
+
+module Handlers = struct
+  let unpwrap = function
+    | `Ok () -> ()
+    | `Error err_msg -> show_message `Error "%s" err_msg
+  ;;
+
+  let w1 f x =
+    try `Ok (f x) with
+    | User_error e -> `Error e
+  ;;
+
+  let ws f x y =
+    match f x with
+    | `Ok f' ->
+      (try `Ok (f' y) with
+       | User_error e -> `Error e)
+    | `Error e -> `Error e
+  ;;
+
+  let w2 f = ws (w1 f)
+  let w3 f x = ws (w2 f x)
+  let w4 f x y = ws (w3 f x y)
+  let w5 f x y z = ws (w4 f x y z)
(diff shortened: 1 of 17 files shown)


voodoos:
@xvw, @PizieDust  these changes are ready for review (but I cannot request it via the UI).

src/import.ml
```diff
@@ -253,3 +253,31 @@ module Ocaml_version = struct
...
+       | User_error e -> `Error e)
+    | `Error e -> `Error e
+  ;;
+
+  let w2 f = ws (w1 f)
+  let w3 f x = ws (w2 f x)
+  let w4 f x y = ws (w3 f x y)
+  let w5 f x y z = ws (w4 f x y z)
+  let _w6 f x y z w = ws (w5 f x y z w)
+end
```
xvw, line 283: Are these standard names? I'm having trouble understanding what `ws` and `wX` mean?

xvw (commented):
Nice, Just a little stressed by `Handlers`.

src/import.ml
voodoos: I just move it from elsewhere, and I agree it's a bit puzzling. We didn't use it in the other commands anyway, I will probably move to do the same.

voodoos: `wrap` I guess

smorimoto (approved):

PR #1679, opened 2024-12-07
ocamlformat: use janestreet profile

smorimoto:
Use the same profile as the OCaml LSP to keep the code style consistent between the LSP and the Client.


smorimoto:
Ref: https://github.com/ocaml/ocaml-lsp/pull/1337

PR #1685, opened 2024-12-11
Add reason language to ast_editor

pedrobslisboa:
## Description 

This PR adds reasonml to ast_editor to be able to see AST and Pp

## Screenshots

Working with both ocaml and reason

Captura de Tela 2024-12-11 às 17 45 26
Captura de Tela 2024-12-11 às 18 24 14

## How to test

- Follow this steps: https://github.com/ocamllabs/vscode-ocaml-platform/blob/88cbda69761298bde70803d5792c90b497a4256f/CONTRIBUTING.md#L27
- On the debug view, create a `.re` file
- To see the pp you need to build it to have the `.pp.ml` file

### DIFF
--- a/src/ast_editor.ml
+++ b/src/ast_editor.ml
@@ -45,16 +45,16 @@ let send_msg t value ~(webview : WebView.t) =
 
 module Pp_path : sig
   type kind =
-    | Structure
-    | Signature
+    | Structure of [ `Ocaml | `Reason ]
+    | Signature of [ `Ocaml | `Reason ]
     | Unknown
 
   val get_kind : document:TextDocument.t -> kind
   val get_pp_path : document:TextDocument.t -> string
 end = struct
   type kind =
-    | Structure
-    | Signature
+    | Structure of [ `Ocaml | `Reason ]
+    | Signature of [ `Ocaml | `Reason ]
     | Unknown
 
   let relative_document_path ~document =
@@ -66,8 +66,10 @@ end = struct
   let get_kind ~document =
     let relative = relative_document_path ~document in
     match Stdlib.Filename.extension relative with
-    | ".ml" -> Structure
-    | ".mli" -> Signature
+    | ".ml" -> Structure `Ocaml
+    | ".re" -> Structure `Reason
+    | ".mli" -> Signature `Ocaml
+    | ".rei" -> Signature `Reason
     | _ -> Unknown
   ;;
 
@@ -80,8 +82,12 @@ end = struct
       let fname_opt =
         match get_kind ~document with
         | Unknown -> None
-        | Structure -> Some (String.chop_suffix_exn ~suffix:".ml" relative ^ ".pp.ml")
-        | Signature -> Some (String.chop_suffix_exn ~suffix:".mli" relative ^ ".pp.mli")
+        | Structure `Ocaml ->
+          Some (String.chop_suffix_exn ~suffix:".ml" relative ^ ".pp.ml")
+        | Signature `Ocaml ->
+          Some (String.chop_suffix_exn ~suffix:".mli" relative ^ ".pp.mli")
+        | Structure `Reason -> Some (relative ^ ".pp.ml")
+        | Signature `Reason -> Some (relative ^ ".pp.mli")
       in
       (match fname_opt with
        | Some fname ->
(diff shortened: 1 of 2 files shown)


src/ast_editor.ml
```diff
@@ -330,10 +336,16 @@ let open_pp_doc instance ~document =
   match fetch_pp_code ~document with
   | Error e -> Promise.return (Error e)
   | Ok pp_pp_str ->
+    let file_name =
+      match Stdlib.Filename.extension (TextDocument.fileName document) with
+      | ".re" -> (String.chop_suffix_exn ~suffix:".re" (TextDocument.fileName document) ^ ".ml")
+      | ".rei" -> (String.chop_suffix_exn ~suffix:".rei" (TextDocument.fileName document) ^ ".mli")
+      | _ -> TextDocument.fileName document
+    in
```
pedrobslisboa, line 344: In the future instead of providing a .ml for pp document we could convert ocaml code to reasonml.

src/ppx_tools/dumpast.ml
```diff
@@ -329,11 +334,17 @@ let transform source kind =
...
-      | `Impl ->
-        let v = Parse.implementation (Lexing.from_string source) in
+      | `Impl Ocaml ->
+        let (v, _) = ML.implementation_with_comments (Lexing.from_string source) in
+        parse_ast#structure v
+      | `Intf Ocaml ->
+        let (v, _) = ML.interface_with_comments (Lexing.from_string source) in
+        parse_ast#signature v
+      | `Impl Reason ->
+        let (v, _) = RE.implementation_with_comments (Lexing.from_string source) in
```
pedrobslisboa, line 344: Using reason instead of Astlib to parse reason and ml files

rgrinberg:
Can we think of a way of having this feature without making the plugin on reason? In the past, I've found the the dependency on reason to be rather disruptive to development. I would much prefer if we had a way to extend this plugin so that you could reason support externally.

rgrinberg:
The problem is that the reason package is not actively maintained, and the version in opam has historically lagged behind. This is particularly painful for a package that depends on the internals of the compiler.

I don't really have a concrete idea on how to work around this. I was hoping that you would think of something. One simple approach is to call out to a reason binary to dump the parsetree

davesnx:
Oh, reason has been very maintained this past years (since 2022), and we plan to keep it that way: https://github.com/reasonml/reason/releases releases are pushed to opam only.

Sure, there might be other options, but it faces problems with compatibility (that a package manager like opam fixes for us) and other issues.

Happy to commit to keep ocaml-platform up-to-date as dependency of Reason.

pedrobslisboa:
@rgrinberg, it turned out that I was overcomplicating it; Ast can handle `.re.`
Just changed it.

src/ast_editor.ml
```diff
@@ -321,9 +327,19 @@ let open_pp_doc instance ~document =
...
+    let file_name =
+      match Pp_path.get_kind ~document with
+      | Structure `Reason ->
+        String.chop_suffix_exn ~suffix:".re" (TextDocument.fileName document)
+        ^ ".ml"
+      | Signature `Reason ->
+        String.chop_suffix_exn ~suffix:".rei" (TextDocument.fileName document)
+        ^ ".mli"
+      | _ -> TextDocument.fileName document
+    in
```
pedrobslisboa, line 339: In nthe future we can use the refmt executable to convert it from ocaml to reasonml.

src/ast_editor.ml
```diff
@@ -321,9 +327,19 @@ let open_pp_doc instance ~document =
...
   | Ok pp_pp_str ->
+    let file_name =
+      match Pp_path.get_kind ~document with
+      | Structure `Reason ->
+        String.chop_suffix_exn ~suffix:".re" (TextDocument.fileName document)
+        ^ ".ml"
+      | Signature `Reason ->
+        String.chop_suffix_exn ~suffix:".rei" (TextDocument.fileName document)
+        ^ ".mli"
+      | _ -> TextDocument.fileName document
```
davesnx, line 338: Since this relies on dune and isn't straight forward to understand reading the code, would you add a comment explaining that .ml is present at the same location as the .re file, etc?

pedrobslisboa: This part is related to creating the PP doc preview. It establishes a preview file with ocaml syntax, but as the current file can be a `.re`, I change it here to fit the ocaml syntax.
Captura de Tela 2024-12-12 às 13 02 31

As I said in [this comment](https://github.com/ocamllabs/vscode-ocaml-platform/pull/1685#discussion_r1882423107), later we can convert this content from ocaml syntax to reasonml, and this change of extension will not be needed anymore.

package.json
```diff
@@ -1269,6 +1269,7 @@
     "@vscode/vsce": "3.2.1",
     "esbuild": "0.24.0",
     "ovsx": "0.10.1",
+    "parcel": "2.13.2",
```
pedrobslisboa, line 1272: Missadded it, going to remove.

smorimoto:
@pedrobslisboa You have to add a change log entry as well.

pedrobslisboa:
> @pedrobslisboa You have to add a change log entry as well.

True, thank you. I'm going to fix it as well

pedrobslisboa:
@smorimoto Done, ready for review

pedrobslisboa:
@rgrinberg @smorimoto 🏓

rgrinberg (approved):

PR #1704, opened 2025-01-07
Jump enhance

PizieDust:
This PR fixes the empty list bug (displays a message rather than an empty quick pick) and also allows the user to preview the different jump targets using the arrow keys without jumping to them.

cc @voodoos


PizieDust:
Closed as the commit history is a bit unclear. I may have pulled changes from older branches unrelated

PR #1705, opened 2025-01-07
Improve quick peek with highlighting in both the jump and holes commands

PizieDust:
This PR 
- fixes the empty list bug (displays a message rather than an empty quick pick) 
- allows the user to preview the different jump targets using the arrow keys without jumping to them.

Match first word from position of interest

https://github.com/user-attachments/assets/cda4a8af-82e5-437b-80b6-377acf9c7085

cc @voodoos @awilliambauer

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -804,9 +804,6 @@ module MerlinJump = struct
                 (TextDocument.getWordRangeAtPosition
                    text_document
                    ~position:(Range.start range)
-                   ~regex:
-                     (Js_of_ocaml.Regexp.regexp
-                        "\\(?\\b(let|fun|match|module|module\\s*type|\\w+)(?=\\s*(?:->|\\s|\\)|$))")
                    ())
                 ~default:range
             in
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -779,7 +779,7 @@ module MerlinJump = struct
       QuickPick.onDidChangeActive
         quickPick
         ~listener:(function
-          | { position; target } :: _ ->
+          | { position; target; _ } :: _ ->
             let range =
               Range.makePositions
                 ~start:position
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -195,14 +195,7 @@ let show_selection selection text_editor =
     text_editor
     ~range:(Selection.to_range selection)
     ~revealType:TextEditorRevealType.InCenterIfOutsideViewport
-    ();
-  let range =
-    TextLine.range
-      (TextDocument.lineAtPosition
-         (TextEditor.document text_editor)
-         ~position:(Selection.start selection))
-  in
-  set_decoration text_editor range
+    ()
 ;;
 
 module Holes_commands : sig
@@ -739,17 +732,20 @@ module MerlinJump = struct
     type t =
       { item : QuickPickItem.t
       ; position : Position.t
+      ; target : string
       }
 
     let t_of_js js =
       let position = Ojs.get_prop_ascii js "position" |> Position.t_of_js in
       let item = QuickPickItem.t_of_js js in
-      { item; position }
+      let target = Ojs.get_prop_ascii js "target" |> Ojs.string_of_js in
+      { item; position; target }
     ;;
 
     let t_to_js t =
       let item = QuickPickItem.t_to_js t.item in
       Ojs.set_prop_ascii item "position" (Position.t_to_js t.position);
+      Ojs.set_prop_ascii item "target" (Ojs.string_to_js t.target);
       item
     ;;
   end
@@ -762,7 +758,7 @@ module MerlinJump = struct
     let quickPickItems =
       List.map results ~f:(fun (target, position) ->
         let item = (QuickPickItem.create ~label:("Jump to nearest " ^ target)) () in
-        { QuickPickItemWithPosition.item; position })
+        { QuickPickItemWithPosition.item; position; target })
     in
     let quickPick =
       QuickPick.set
@@ -783,10 +779,19 @@ module MerlinJump = struct
       QuickPick.onDidChangeActive
         quickPick
         ~listener:(function
-          | { position; _ } :: _ ->
+          | { position; item; target } :: _ ->
+            let range =
+              Range.makePositions
+                ~start:position
+                ~end_:
+                  (Position.make
+                     ~character:(Position.character position + String.length target)
+                     ~line:(Position.line position))
+            in
             show_selection
               (Selection.makePositions ~anchor:position ~active:position)
-              text_editor
+              text_editor;
+            set_decoration text_editor range
           | _ -> ())
         ()
     in
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -779,7 +779,7 @@ module MerlinJump = struct
       QuickPick.onDidChangeActive
         quickPick
         ~listener:(function
-          | { position; item; target } :: _ ->
+          | { position; target } :: _ ->
             let range =
               Range.makePositions
                 ~start:position
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -736,40 +736,37 @@ module MerlinJump = struct
     show_selection (Selection.makePositions ~anchor:position ~active:position) text_editor
   ;;
 
-  module QuickPickItemWithPosition = struct
+  module JumpQuickPickItem = struct
     type t =
       { item : QuickPickItem.t
       ; position : Position.t
-      ; target : string
       }
 
     let t_of_js js =
       let position = Ojs.get_prop_ascii js "position" |> Position.t_of_js in
       let item = QuickPickItem.t_of_js js in
-      let target = Ojs.get_prop_ascii js "target" |> Ojs.string_of_js in
-      { item; position; target }
+      { item; position }
     ;;
 
     let t_to_js t =
       let item = QuickPickItem.t_to_js t.item in
       Ojs.set_prop_ascii item "position" (Position.t_to_js t.position);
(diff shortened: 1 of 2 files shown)


src/extension_commands.ml
```diff
@@ -699,23 +725,109 @@ module MerlinJump = struct
...
+      Ojs.set_prop_ascii item "position" (Position.t_to_js t.position);
+      Ojs.set_prop_ascii item "target" (Ojs.string_to_js t.target);
+      item
+    ;;
+  end
+
+  module QuickPick = Vscode.QuickPick.Make (QuickPickItemWithPosition)
+
+  let display_results (results : Custom_requests.Merlin_jump.response) text_editor =
+    let open Promise.Syntax in
```
xvw, line 756: WDYT about moving the opening of `Promise.Syntax` inside the function given to the `listener` at line `804`?

src/extension_commands.ml
```diff
@@ -699,23 +725,109 @@ module MerlinJump = struct
...
+      QuickPick.onDidAccept
+        quickPick
+        ~listener:(fun () ->
+          match QuickPick.selectedItems quickPick with
+          | Some (item :: _) ->
+            ignore
+              (let position = item.position in
+               let* () = jump_to_position text_editor position in
+               selected_item := true;
+               Promise.return ())
```
xvw, line 808:
```ocaml
ignore (
  let open Promise.Syntax in
  let+ () = jump_to_position text_editor in
  selected_item := true)
```

xvw (approved):
Just a small nitpick. Feel free to discard it!

PizieDust:
> Just a small nitpick. Feel free to discard it!

It's done in the latest commit :)

xvw (approved):

src/extension_commands.ml
```diff
@@ -158,6 +158,48 @@ let _open_ocamllsp_output_pane, _open_ocaml_platform_ext_pane, _open_ocaml_comma
       (handler Output.command_output_channel) )
 ;;
 
+let decorationType =
```
voodoos, line 161: This name should describe what does this decoration looks like and/or what are correct use cases for it such as `highlighting_decoration` or `quick_peek_highlighting_decoration`.

src/extension_commands.ml
```diff
@@ -158,6 +158,48 @@ let _open_ocamllsp_output_pane, _open_ocaml_platform_ext_pane, _open_ocaml_comma
...
+  TextEditor.setDecorations
+    text_editor
+    ~decorationType
+    ~rangesOrOptions:(`Ranges [ range ])
+;;
+
+let remove_decoration text_editor =
+  let rangesOrOptions = `Ranges [] in
+  TextEditor.setDecorations text_editor ~decorationType ~rangesOrOptions
+;;
```
voodoos, line 192: I think you should move the decoration related code to a sub-module `Decorations` and rename `set_decoration` to a more specific name like `highlight_range` and `remove...` to `remove_all_highlightings`.

src/extension_commands.ml
```diff
@@ -699,23 +727,134 @@ module MerlinJump = struct
...
-      (Selection.makePositions ~anchor:position ~active:position);
-    TextEditor.revealRange
-      text_editor
-      ~range:(Range.makePositions ~start:position ~end_:position)
-      ~revealType:TextEditorRevealType.InCenterIfOutsideViewport
-      ()
+    show_selection (Selection.makePositions ~anchor:position ~active:position) text_editor
+  ;;
+
+  module QuickPickItemWithPosition = struct
```
voodoos, line 733: There is also a target in there :-)

```suggestion
  module JumpQuickPickItem = struct
```

src/extension_commands.ml
```diff
@@ -699,23 +727,134 @@ module MerlinJump = struct
...
+    in
+    let _disposable =
+      let initial_selection = TextEditor.selection text_editor in
+      QuickPick.onDidHide
+        quickPick
+        ~listener:(fun () ->
+          if !selected_item then () else show_selection initial_selection text_editor;
+          remove_decoration text_editor;
+          QuickPick.dispose quickPick)
+        ()
```
voodoos, line 847: I am a bit worried that if the users decides not-to-choose a jump target in the list and clicks somewhere in the texteditor the cursor won't go to the location that was clicked on but to the locations before calling the jump command. Could you check if that is the case or not ?

src/extension_commands.ml
```diff
@@ -699,23 +727,134 @@ module MerlinJump = struct
...
+              text_editor;
+            let text_document = TextEditor.document text_editor in
+            let range =
+              Option.value
+                (TextDocument.getWordRangeAtPosition
+                   text_document
+                   ~position:(Range.start range)
+                   ~regex:
+                     (Js_of_ocaml.Regexp.regexp
+                        "\\(?\\b(let|fun|match|module|module\\s*type|\\w+)(?=\\s*(?:->|\\s|\\)|$))")
```
voodoos, line 803: What happens if you don't use that regex at all ?

src/extension_commands.ml
voodoos: Also, do you use the `item` or the `target` fields` at all ?

src/extension_commands.ml
PizieDust: Indeed by default, if the use doesn't select an item the cursor will always return to the location before the jump command was called.

src/extension_commands.ml
PizieDust: Just checked and we can indeed get rid of it. The default behavior of `getWordRangeAtPosition` will grab the first word of the line.

src/extension_commands.ml
PizieDust: We don't need target. I had added it when I was experimenting with the highlights.

PizieDust: Given that item is the `QuickPickItem` itself, we need it so the results is rendered in the quickpick.
Without it we get:

image

Moving the arrow keys we can see the highlights of the target but the quickpick will be blank

src/extension_commands.ml
PizieDust: We can use a `TextEditorSelectionChangeEvent` to detect mouse clicks and set the position of where the user clicks. 

https://code.visualstudio.com/api/references/vscode-api#TextEditorSelectionChangeKind.Mouse

I can see that you already implemented the modules in the TypeSelection PR https://github.com/ocamllabs/vscode-ocaml-platform/pull/1675

voodoos: Yes, you can copy paste it

src/extension_commands.ml
PizieDust: I will reinstate the regex but a more simpler version, just to be able to match when there are paranthesis. 

the match in `(match` won't be highlighted without the regex.

src/extension_commands.ml
```diff
@@ -841,7 +866,9 @@ module MerlinJump = struct
       QuickPick.onDidHide
         quickPick
         ~listener:(fun () ->
-          if !selected_item then () else show_selection initial_selection text_editor;
+          if !selected_item
+          then ()
+          else Disposable.dispose (move_cursor_after_quickpick_close initial_selection ());
```
voodoos, line 871: I think this does nothing: you start listening to an event and then dispose of it immediately.

I'm not sure of the best way to do it myself, but maybe we should, when the quickpick is being shown, watch for such events so that we set a flag when the user changed the selection to handle the onDidHide event accordingly.

src/import.ml
```diff
@@ -253,3 +253,31 @@ module Ocaml_version = struct
...
+       | User_error e -> `Error e)
+    | `Error e -> `Error e
+  ;;
+
+  let w2 f = ws (w1 f)
+  let w3 f x = ws (w2 f x)
+  let w4 f x y = ws (w3 f x y)
+  let w5 f x y z = ws (w4 f x y z)
+  let _w6 f x y z w = ws (w5 f x y z w)
+end
```
voodoos, line 283: We don't use it for other listeners, maybe we don't need it ?

src/extension_commands.ml
```diff
@@ -158,6 +158,71 @@ let _open_ocamllsp_output_pane, _open_ocaml_platform_ext_pane, _open_ocaml_comma
...
+(**  If the user hits the ESC key, this should go back to the initial_selection,
+     otherwise the current position of the click is used *)
+let move_cursor_after_selection_change event_fired () =
+  let onDidChangeTextEditorSelection_listener event =
+    event_fired := true;
+    let selections = TextEditorSelectionChangeEvent.selections event in
+    match selections with
+    | [ selection ] ->
+      show_selection selection (TextEditorSelectionChangeEvent.textEditor event)
+    | _ -> ()
```
voodoos, line 217: In my view of the control flow, we don't need to do anything here, simply update the flag that says the selection changed.

src/extension_commands.ml
```diff
@@ -677,45 +742,145 @@ module MerlinJump = struct
...
+      let initial_selection = TextEditor.selection text_editor in
+      let selection_listener_disposable =
+        move_cursor_after_selection_change event_fired ()
+      in
+      QuickPick.onDidHide
+        quickPick
+        ~listener:(fun () ->
+          if !event_fired || !selected_item
+          then ()
+          else show_selection initial_selection text_editor;
```
voodoos, line 869:
```suggestion
          if not(!event_fired || !selected_item)
          then show_selection initial_selection text_editor;
```

src/extension_commands.ml
```diff
@@ -677,45 +742,145 @@ module MerlinJump = struct
...
+          Decorations.remove_all_highlights text_editor;
+          QuickPick.dispose quickPick;
+          Disposable.dispose selection_listener_disposable)
```
voodoos, line 872:
```suggestion
          Disposable.dispose selection_listener_disposable);
          QuickPick.dispose quickPick
```

PR #1719, opened 2025-01-31
Detect dune pkg and auto configure it

PizieDust:
This PR can detect that we are using Dune Pkg by checking the presence of a `dune.lock` directory.
If the user has not run `dune pkg lock` yet, the editor will use the "Default" sandbox. On the next time they open the editor, it'll configure a `Custom sandbox` for Dune Pkg  automatically. 

## Test

- Initialize a new dune project
```
dune init proj hello
```

- Lock packages
```
dune pkg lock
```

- Open project in (this) vscode
you should see a notification informing you of Dune Pkg being active.

cc @voodoos

### DIFF
--- a/package.json
+++ b/package.json
@@ -1331,7 +1331,7 @@
     "@vscode/test-electron": "2.5.2",
     "@vscode/vsce": "3.3.2",
     "esbuild": "0.25.2",
-    "ovsx": "0.10.1",
+    "ovsx": "0.10.2",
     "typescript": "5.8.3"
   },
   "packageManager": "yarn@4.9.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4588,7 +4588,7 @@ __metadata:
     "@vscode/test-electron": "npm:2.5.2"
     "@vscode/vsce": "npm:3.3.2"
     esbuild: "npm:0.25.2"
-    ovsx: "npm:0.10.1"
+    ovsx: "npm:0.10.2"
     polka: "npm:1.0.0-next.28"
     sirv: "npm:3.0.1"
     typescript: "npm:5.8.3"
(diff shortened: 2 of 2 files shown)


src/sandbox.ml
```diff
@@ -282,13 +282,28 @@ let detect_opam_sandbox ~project_root opam () =
...
+  let open Promise.Syntax in
+  (* Path to the dune.lock file *)
+  let dune_lock_path = Path.join project_root (Path.of_string "dune.lock") in
+  let+ exists = Fs.exists (Path.to_string dune_lock_path) in
+  if exists
+  then (
+    show_message
+      `Info
+      "Dune Pkg detected. Configuring a custom sandbox for the workspace.";
+    Some (Custom "$prog $args"))
```
voodoos, line 295: A note that this work because dune package management install a shim script for `ocamllsp`. In the future we probably want to do the same thing as in that script:

```
#!/bin/sh

set -eu

# Script to choose an appropriate instance of ocamllsp. If the project has a
# lockdir then ocamllsp will be installed and run as a dev-tool. Otherwise the
# script will try to launch an opam installation of ocamllsp.

if [ -d dune.lock ]; then

    OCAMLFORMAT_TARGET="_build/_private/default/.dev-tool/ocamlformat/ocamlformat/target"

    if [ ! -f $OCAMLFORMAT_TARGET/cookie ]; then
        # Make sure that the ocamlformat dev tool is installed as it's needed by
        # ocamllsp. There's currently no command that just installs ocamlformat so
        # we need to run it and ignore the result.
        dune tools exec ocamlformat -- --help > /dev/null
    fi

    # Add ocamlformat to the environment in which ocamllsp runs so ocamllsp can invoke ocamlformat.
    export PATH="$PWD/$OCAMLFORMAT_TARGET/bin:$PATH"

    # Build and run ocamllsp.
    dune tools exec ocamllsp -- "$@"
else
...
```

And we should find a better solution for the server's use of ocamlformat.

cc @gridbugs @maiste

voodoos, line 295: Custom was a handy workaround so far, but we really need to introduce a new variant `Dune_pkg` now.

PR #1725, opened 2025-02-10
Automatically install or update ocaml-lsp-server 

PizieDust:
Currently, when `ocaml-lsp-server` is not found in a sandbox, the language server displays an error and fails to start. 
This PR improves users experience by providing a UI that asks if they will like to install or update `ocaml-lsp-server` or select another sandbox.

If the user wants to select another sandbox, a quickpick will be displayed with various switches. 
If the user decides to install `ocaml-lsp-server`, it will be installed in the background and they can continue with their activities in the editor. 

This PR will streamline the process for complete beginners and also reduce the time to terminal for experienced users who often will have to switch to the terminal to install `ocaml-lsp-server` in their sandbox.

cc @voodoos

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -59,7 +59,7 @@ let _select_sandbox =
 let _install_ocaml_lsp_server =
   let handler (instance : Extension_instance.t) ~args:_ =
     let open Promise.Syntax in
-    let (_ : unit Promise.t) =
+    let _ =
       let sandbox = Extension_instance.sandbox instance in
       let* ocamllsp_present = Extension_instance.check_ocaml_lsp_available sandbox in
       match ocamllsp_present with
@@ -80,7 +80,7 @@ let _install_ocaml_lsp_server =
 
 let _restart_language_server =
   let handler (instance : Extension_instance.t) ~args:_ =
-    let (_ : unit Promise.t) = Extension_instance.start_language_server instance in
+    let _ = Extension_instance.start_language_server instance in
     ()
   in
   command Extension_consts.Commands.restart_language_server handler
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -70,8 +70,7 @@ let _install_ocaml_lsp_server =
         show_message `Info "Installing the latest release of ocaml-lsp-server...";
         let* () = Extension_instance.install_ocaml_lsp_server sandbox in
         show_message `Info "Installation of ocaml-lsp-server completed successfully.";
-        let+ () = Extension_instance.start_language_server instance in
-        ()
+        Extension_instance.start_language_server instance
     in
     ()
   in
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -67,7 +67,6 @@ let _install_ocaml_lsp_server =
         show_message `Info "ocaml-lsp-server is already installed and up to date."
         |> Promise.return
       | Error _ ->
-        show_message `Info "Installing the latest release of ocaml-lsp-server...";
         let* () = Extension_instance.install_ocaml_lsp_server sandbox in
         show_message `Info "Installation of ocaml-lsp-server completed successfully.";
         Extension_instance.start_language_server instance
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -64,11 +64,11 @@ let _install_ocaml_lsp_server =
       let* ocamllsp_present = Extension_instance.check_ocaml_lsp_available sandbox in
       match ocamllsp_present with
       | Ok () ->
-        show_message `Info "ocaml-lsp-server is already installed and up to date."
+        show_message `Info "OCaml-LSP server is already installed."
         |> Promise.return
       | Error _ ->
         let* () = Extension_instance.install_ocaml_lsp_server sandbox in
-        show_message `Info "Installation of ocaml-lsp-server completed successfully.";
+        show_message `Info "Installation of OCaml-LSP server completed successfully.";
         Extension_instance.start_language_server instance
     in
     ()
@@ -92,7 +92,7 @@ let _upgrade_ocaml_lsp_server =
         let+ _ = Extension_instance.start_language_server instance in
         show_message
           `Info
-          "ocaml-lsp-server upgraded successfully to version %s"
+          "OCaml-LSP server upgraded successfully to version %s"
           latest_version
     in
     ()
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -287,7 +287,9 @@ end = struct
       show_message
         `Warn
         "The installed version of `ocamllsp` does not support typed holes. %s"
-        msg
+        msg;
+      let _ = Extension_instance.suggest_to_upgrade_ocaml_lsp_server () in
+      ()
   ;;
 
   let current_cursor_pos text_editor =
@@ -603,7 +605,9 @@ module Copy_type_under_cursor = struct
       show_message
         `Warn
         "The installed version of `ocamllsp` does not support type enclosings. %s"
-        msg
+        msg;
+      let _ = Extension_instance.suggest_to_upgrade_ocaml_lsp_server () in
+      ()
   ;;
 
   let get_enclosings text_editor client =
@@ -1143,7 +1147,9 @@ module Search_by_type = struct
            when ocaml_lsp_doesnt_support_search_by_type ocaml_lsp ->
            show_message
              `Warn
-             "The installed version of `ocamllsp` does not support type search"
+             "The installed version of `ocamllsp` does not support type search";
+           let _ = Extension_instance.suggest_to_upgrade_ocaml_lsp_server () in
+           ()
          | Some (client, _) -> show_query_input text_editor client)
     in
     command Extension_consts.Commands.search_by_type handler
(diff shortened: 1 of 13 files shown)


src/extension_commands.ml
```diff
@@ -56,6 +56,28 @@ let _select_sandbox =
   command Extension_consts.Commands.select_sandbox handler
 ;;
 
+let _install_ocaml_lsp_server =
+  let handler (instance : Extension_instance.t) ~args:_ =
+    let open Promise.Syntax in
+    let (_ : unit Promise.t) =
```
xvw, line 62: I think that type anotation is not useful?

src/extension_instance.ml
```diff
@@ -180,44 +213,48 @@ end = struct
...
-        "An error occurred starting the language server `ocamllsp`. %s"
-        s
+       | Error s ->
+         show_message
+           `Error
+           "An error occurred starting the language server `ocamllsp`. %s"
+           s)
+    | Error _ ->
+      let+ () = suggest_to_install_ocaml_lsp_server () in
+      ()
```
voodoos, line 257:
```suggestion
      suggest_to_install_ocaml_lsp_server ()
```

voodoos (commented):
Thanks Pizie, this looks quite reasonable and works great.

I understand now the "notifications" you were talking about earlier:

image

Maybe we could improve the message of the "Installing sandbox packages" to also mention the name of the packages so that we could remove the additional "Installing the latest release of...." notification ?

PizieDust:
> Thanks Pizie, this looks quite reasonable and works great.
> 
> I understand now the "notifications" you were talking about earlier:
> 
> image
> Maybe we could improve the message of the "Installing sandbox packages" to also mention the name of the packages so that we could remove the additional "Installing the latest release of...." notification ?


Thank you. This is done in the latest commit.

src/extension_commands.ml
xvw: @smorimoto 
Ok, since it is wrapped into a comprehension (via `let*`), inference works pretty well, so in this review, FMPOV, it is noise, but my goal is not to change the full code-base :)

voodoos:
I have one concern with the current behavior: what install package does is:

```
$ update --switch=5.3.0
$ install --switch=5.3.0 -y ocaml-lsp-server
```

There is nothing fundamentally wrong with that, but I think the second command will also upgrade every upgradable package in the switch. This might be something we want to skip to prevent potentially breaking user projects ? (not sure if that is doable however)

src/extension_commands.ml
```diff
@@ -56,9 +56,29 @@ let _select_sandbox =
...
 
+let _install_ocaml_lsp_server =
+  let handler (instance : Extension_instance.t) ~args:_ =
+    let open Promise.Syntax in
+    let _ =
+      let sandbox = Extension_instance.sandbox instance in
+      let* ocamllsp_present = Extension_instance.check_ocaml_lsp_available sandbox in
+      match ocamllsp_present with
+      | Ok () ->
+        show_message `Info "ocaml-lsp-server is already installed and up to date."
```
voodoos, line 67: Actually, we don't know here that the server is "up to date", only that it is installed.

There is `Ocaml_lsp.is_version_up_to_date` that checks that and warn the user. We should probably use it and make it actionable.

PizieDust: I just modified the PR to also prompt the user to update `ocaml-lsp-server` to the latest version (as per the vscode platform extension).

voodoos: Still, at that point we don't know that it is up to date, the message should simply states that the server is installed.

src/extension_instance.ml
```diff
@@ -150,21 +167,68 @@ end = struct
...
+  let suggest_to_upgrade_ocaml_lsp_server err =
+    let open Promise.Syntax in
+    let upgrade_lsp_text = "Upgrade OCaml-LSP server" in
+    let select_different_sanbox = "Select a different Sandbox" in
+    let+ selection =
+      Window.showInformationMessage
+        ~message:("OCaml-LSP server is not up to date: " ^ err)
+        ~choices:
+          [ upgrade_lsp_text, upgrade_lsp_text
+          ; select_different_sanbox, select_different_sanbox
```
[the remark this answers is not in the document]
PizieDust: For the `Don't show this message again`, should this apply for the specific sandbox or this should be apply for all sandbox?

voodoos (commented):
Looks very good. I have a few suggestions left for the upgrades 🙂

src/extension_commands.ml
```diff
@@ -247,7 +287,9 @@ end = struct
       show_message
         `Warn
         "The installed version of `ocamllsp` does not support typed holes. %s"
-        msg
+        msg;
+      let _ = Extension_instance.suggest_to_upgrade_ocaml_lsp_server () in
+      ()
```
voodoos, line 292: It seems like this would be done in `Ocaml_lsp.is_version_up_to_date` if that is possible, that would remove the need to handle each of these cases.

src/extension_instance.ml
```diff
@@ -108,6 +108,45 @@ let stop_server t =
...
+      ()
+  in
+  match selection with
+  | Some `Update_lsp ->
+    let (_ : Ojs.t option Promise.t) =
+      Vscode.Commands.executeCommand
+        ~command:Extension_consts.Commands.upgrade_ocaml_lsp_server
+        ~args:[]
+    in
+    ()
```
voodoos, line 145: I think we should wait for that promise to resolve.

src/extension_instance.ml
```diff
@@ -150,21 +189,35 @@ end = struct
...
+          ~args:[]
+      in
+      ()
+    | Some `Select_sandbox ->
+      let (_ : Ojs.t option Promise.t) =
+        Vscode.Commands.executeCommand
+          ~command:Extension_consts.Commands.select_sandbox
+          ~args:[]
+      in
+      ()
```
voodoos, line 219: Same here, what happens if we do wait for these promises to fulfill instead of ignoring them ?

src/extension_instance.ml
```diff
@@ -239,6 +296,38 @@ let documentation_server_info () =
...
+    Vscode.Commands.executeCommand
+      ~command:Extension_consts.Commands.refresh_switches
+      ~args:[]
+  in
+  let (_ : Ojs.t option Promise.t) =
+    Vscode.Commands.executeCommand
+      ~command:Extension_consts.Commands.refresh_sandbox
+      ~args:[]
+  in
+  ()
```
voodoos, line 312: Same here, is there a reason we're not waiting for these promises to fulfill ?

voodoos (commented):
Thanks a lot Pizie, that looks quite close to being ready :-)

src/extension_commands.ml
PizieDust: if we do that it'll cause a `Error: dependency cycle between modules` error

src/extension_instance.ml
```diff
@@ -239,6 +272,38 @@ let documentation_server_info () =
...
```
voodoos, line 304:
```suggestion
  let* () = Sandbox.upgrade_packages sandbox ~packages:[ "ocaml-lsp-server" ] in
  let* (_ : Ojs.t option) =
    Vscode.Commands.executeCommand
      ~command:Extension_consts.Commands.refresh_switches
      ~args:[]
  in
  let+ (_ : Ojs.t option) =
    Vscode.Commands.executeCommand
      ~command:Extension_consts.Commands.refresh_sandbox
      ~args:[]
  in
  ()
```

src/extension_instance.ml
voodoos: It turns out that right now the binding to execute vscode commands does not allows us to really wait for that command to complete, but that's something we should consider fixing later.

src/ocaml_lsp.ml
```diff
@@ -222,6 +244,7 @@ let is_version_up_to_date t ocaml_v =
             else Error (`Newer_available (Some v, available.(last)))))
   in
   Result.map_error res ~f:(fun err ->
+    let _ = suggest_to_upgrade_ocaml_lsp_server () in
```
voodoos, line 247:
```suggestion
```

Hum I now realize that this introduce a bad circle: `suggest` calls `upgrade_ocaml_lsp_server` which calls `is_version_up_to_date` :-) 

Let's keep it simple for now, and suggest to upgrade after calls to `is_version_up_to_date` when it makes sense.

voodoos (approved):
Just made a few more remarks, but once fixed I think this is ready to go.

In my opinion this hints toward a long-awaited refactoring of how we check for available commands. It could be unified so that we have the same behavior for every command: Not available ? Then do you want to upgrade ? Nice. Arg.

But that's the subject of another PR.

src/extension_commands.ml
```diff
@@ -1274,7 +1316,9 @@ module Navigate_holes = struct
            ->
            show_message
              `Warn
-             "The installed version of `ocamllsp` does not support typed hole navigation"
+             "The installed version of `ocamllsp` does not support typed hole navigation";
+           let _ = Ocaml_lsp.suggest_to_upgrade_ocaml_lsp_server () in
```
voodoos, line 1320: We're missing the same thing in the `type_selection.ml` module :-)

voodoos: Also having two messages shown is not ideal:

image

Maybe we could have an option to customize the message of the `suggest_to_upgrade_ocaml_lsp_server` function ?

voodoos (approved):
Thanks Pizie, I think the current state is all-right for introducing this new functionality.

We might want to do some refactoring / cleanup of the mechanism to detect missing features in the future.

PR #1741, opened 2025-03-03
Fix odoc syntax color for math block with braces

dlesbre:
Fix a small bug in odoc syntax coloring. The math mode (`{m ...}` and `{math ... }` blocks) used to terminate at the first closing brace `}`. However, math mode uses LaTeX syntax and thus, can contain a lot of braces (eg. `2^{256}`). This PR fixes that issue.

Here is a small comment that shows the difference:
```ocaml
(**
- Inline math: {m 2^{256}+2^{128}}
- Display math:  {math
2^{256}+2^{128}
}
*)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+- Fix syntax color for odoc math blocks containing braces (#1741)
+
 ## 1.28.0
 
 - Add `.re` file extension support. (#1685)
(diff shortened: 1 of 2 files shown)


dlesbre:
@mnxn I think you're the one who usually reviews these PRs.

smorimoto (approved):

smorimoto:
I merge this because this looks fine to me. If @mnxn finds the issue later, we can fix it at that point.

PR #1750, opened 2025-03-11
Map Alt+D to the correct action kind in package.json

dalps:
The extension assigns the key binding Alt+D to the very useful "destruct" code action, but currently the key binding doesn't do anything (also see https://github.com/ocaml/ocaml-lsp/issues/1301).

I did some digging and discovered that the definition of the key binding is passing the `destruct` kind, but ocaml-lsp actually names the kind `destruct (enumerate cases)`.

Now the keybinding is correctly mapped to the destruct action, and I also see the keys suggested in the code actions context menu:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Update `destruct` code action command to use `destruct (enumerate cases)` (#1750)
+
 ## 1.28.1
 
 - Fix syntax color for odoc math blocks containing braces (#1741)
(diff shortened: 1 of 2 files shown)


smorimoto:
@voodoos I think this should be fixed on the LSP side, what do you think?

voodoos:
> @voodoos I think this should be fixed on the LSP side, what do you think?

I don't think so, the action was renamed and split into two in https://github.com/ocaml/ocaml-lsp/pull/1283
This PR looks good to me.

voodoos (approved):

smorimoto:
Oh, I didn't know it was renamed. Let's add a change log entry and go ahead!

PR #1791, opened 2025-04-18
Automatically configure Dune Package Management

PizieDust:
This PR attempts to automatically configure dune pkg management by detecting the presence of a `dune.lock` directory or the user selects `Dune Package Management` as a sandbox. 
If they do, and there's  no `dune.lock`, we should ask if they wish to run `dune pkg lock` or select a different sandbox.


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 - Fix syntax color for odoc math blocks containing braces (#1741)
 
+- Detect Dune Pkg and auto-configure a Custom sandbox for it. (#1719)
```
smorimoto, line 15: Could you move this to `Unreleased` section!

src/sandbox.ml
```diff
@@ -99,12 +103,14 @@ module Kind = struct
...
     | Global
     | Custom
+    | Dune_pkg
 
   let of_string = function
     | "opam" -> Some Opam
     | "esy" -> Some Esy
     | "global" -> Some Global
     | "custom" -> Some Custom
+    | "Dune Package Manager" -> Some Dune_pkg
```
smorimoto, line 113:
```suggestion
    | "dune" -> Some Dune
```

src/sandbox.ml
```diff
@@ -121,6 +127,7 @@ module Kind = struct
     | Esy -> "esy"
     | Global -> "global"
     | Custom -> "custom"
+    | Dune_pkg -> "Dune Package Manager"
```
smorimoto, line 130:
```suggestion
    | Dune -> "dune"
```

src/sandbox.ml
```diff
@@ -91,6 +94,7 @@ let to_pretty_string t =
     print_opam project_name
   | Global -> "Global OCaml"
   | Custom _ -> "Custom OCaml"
+  | Dune_pkg _ -> "Dune Package Manager"
```
smorimoto, line 97:
```suggestion
  | Dune _ -> "Dune Package Manager"
```

src/sandbox.ml
```diff
@@ -76,6 +78,7 @@ let to_string = function
   | Opam (_, switch) -> Printf.sprintf "opam(%s)" (Opam.Switch.name switch)
   | Global -> "global"
   | Custom _ -> "custom"
+  | Dune_pkg _ -> "Dune Package Manager"
```
smorimoto, line 81:
```suggestion
  | Dune _ -> "dune"
```

smorimoto (commented):
I feel we should just use 'dune' here.

smorimoto (approved):

src/extension_instance.ml
```diff
@@ -167,35 +224,48 @@ end = struct
       LanguageClient.Executable.create ~command ~args ~options ()
   ;;
 
-  let suggest_to_install_ocaml_lsp_server () =
+  let suggest_to_install_ocaml_lsp_server t () =
```
voodoos, line 227: I don't think we should modify `suggest_to_install_ocaml_lsp_server` or if we do we should rename it.

It makes sense in an opam or esy sandbox, not when using dune pkg management.

Instead, when the server fails to start in Extension_instance.start_language_server`, we should look at the sandbox: if it is Dune, then propose to run dune lock (if that is the actual correct solution). If it's Opam or another: propose to install ocaml-lsp-server.

src/extension_instance.ml
```diff
@@ -434,39 +504,51 @@ let update_ocaml_info t =
...
-         Option.iter maybe_choice ~f:(fun f -> f ())
+    (match t.sandbox with
+     | Dune _ ->
+       let open Promise.Option.Syntax in
+       let _ =
+         let+ project_root = Sandbox.workspace_root () |> Promise.return in
+         let* is_dune_pkg_locked = Sandbox.detect_dune_lock_dir ~project_root () in
+         match is_dune_pkg_locked with
+         | Dune _ -> Some () |> Promise.return
+         | _ -> Some (suggest_to_run_dune_pkg_lock t) |> Promise.return
```
voodoos, line 515: Suggesting to run dune pkg lock here won't actually fix the fact that the plugin wasn't able to detect OCaml's version, right ? Wouldn't it be more correct to do nothing here ? And later, when we will have a way to probe OCaml's version via Dune package management, then we will actually fix the check for real.

src/extension_instance.ml
```diff
@@ -167,35 +224,48 @@ end = struct
...
-    | Some `Install_lsp ->
-      let+ (_ : Ojs.t option) =
-        Vscode.Commands.executeCommand
-          ~command:Extension_consts.Commands.install_ocaml_lsp_server
-          ~args:[]
+    match t.sandbox with
+    | Dune _ ->
+      let _ =
+        let open Promise.Option.Syntax in
+        let* project_root = Sandbox.workspace_root () |> Promise.return in
```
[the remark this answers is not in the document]
voodoos: Also, we should definitely store the project root path in the `| Dune _` variant and use that directly.

src/sandbox.ml
```diff
@@ -282,13 +296,22 @@ let detect_opam_sandbox ~project_root opam () =
   Opam (opam, switch)
 ;;
 
+let detect_dune_lock_dir ~project_root () =
+  let open Promise.Syntax in
+  (* Path to the dune.lock file *)
+  let dune_lock_path = Path.join project_root (Path.of_string "dune.lock") in
+  let+ exists = Fs.exists (Path.to_string dune_lock_path) in
+  if exists then Some (Dune "$prog $args") else None
```
voodoos, line 304: Again, you never use the information that you store in the `Dune` constructor: `"$prog $args"` you should put nothing, or better, store the project root.

src/sandbox.mli
```diff
@@ -30,11 +30,14 @@ type t =
...
   | Global
   | Custom of string
+  | Dune of string
 
+val workspace_root : unit -> Path.t option
 val equal : t -> t -> bool
 val to_string : t -> string
 val to_pretty_string : t -> string
 val of_settings : unit -> t option Promise.t
+val detect_dune_lock_dir : project_root:Path.t -> unit -> t option Promise.t
```
voodoos, line 40: This function really should return true or false.

src/extension_instance.ml
```diff
@@ -125,6 +125,63 @@ let check_ocaml_lsp_available sandbox =
...
 ;;
 
+let suggest_to_run_dune_pkg_lock t =
+  let open Promise.Syntax in
+  let (_ : unit Promise.t) =
+    let+ maybe_choice =
+      Window.showWarningMessage
+        ~message:
+          "Dune Package Manager is detected but no lock file is present. Do you want to \
+           run `dune pkg lock`"
```
voodoos, line 135: "Dune Package Manager is selected as the active sandbox, but no lock file is present. Do you want to  run `dune pkg lock`?"

src/extension_instance.ml
```diff
@@ -125,6 +125,63 @@ let check_ocaml_lsp_available sandbox =
...
+        ~choices:
+          [ ( "Run dune pkg lock"
```
voodoos, line 137:
```suggestion
          [ ( "Generate lockfile"
```

src/extension_instance.ml
```diff
@@ -125,6 +125,63 @@ let check_ocaml_lsp_available sandbox =
...
+                  in
+                  let task ~progress:_ ~token:_ =
+                    let+ result =
+                      let _ =
+                        let terminal = Terminal_sandbox.create t.sandbox in
+                        let _ = Terminal_sandbox.show ~preserveFocus:true terminal in
+                        Terminal_sandbox.send terminal "dune pkg lock"
+                      in
+                      Ok () |> Promise.return
+                    in
```
voodoos, line 155: None of these `Terminal_sandbox` functions return a promise, so why would you ?

src/extension_instance.ml
```diff
@@ -125,6 +125,63 @@ let check_ocaml_lsp_available sandbox =
...
+                    match result with
+                    | Ok () -> Ojs.null
+                    | Error err ->
+                      show_message
+                        `Error
+                        "An error occured while running dune pkg lock: %s"
+                        err;
+                      Ojs.null
```
voodoos, line 163: `result` is always `Ok ()` so this never happen.

And I mean, this is a shame, we should be able to wait for the command to finish and actually check the result, isn't there a way to do that ? I think this is what happen when installing a package.

voodoos (changes requested):
Thanks for this work @PizieDust !

You should be careful when introducing new types, such as the new constructor `Dune` for sandboxes. You decided to store a string (that is always the template "$prog $arg") in it but you never use that information anywhere. That's a sign that it's useless. Instead, you should store the path to the project root that would simplify your life later one. 

When running commands that you don't want to show to the user, you should not do that in a terminal. You should do it in the same way than it is done for installing opam packages for example.

src/dune_pkg.ml
smorimoto, line 1: Rename this to `dune.ml`

voodoos:
> @voodoos Can you review this again?

Yes, we are working on it with Pizie, there are still changes to be made :-)

smorimoto:
Fantastic! Thanks a lot!

voodoos:
@PizieDust is this ready for another review or not yet ?

PizieDust:
> @PizieDust is this ready for another review or not yet ?

Yes, it's ready. :)

src/sandbox.ml
```diff
@@ -9,32 +9,38 @@ module Package = struct
   type t =
     | Opam of Opam.Package.t
     | Esy of Esy.Package.t
+    | Dune of Dune.Package.t
```
voodoos, line 12: Since we don't actually handle anything package related for dune (it's all failwith !), do we really have to add this case now ?

src/sandbox.ml
```diff
@@ -231,6 +259,14 @@ let of_settings () : t option Promise.t =
          None))
   | Some Global -> Promise.return (Some Global)
   | Some (Custom template) -> Promise.return (Some (Custom template))
+  | Some (Dune dune) ->
+    let open Promise.Syntax in
+    let* dune = (available ~root:dune ()).dune in
```
voodoos, line 264: I don't think this is correct, if Dune is installed, then dune pkg is available, whatever the project root is.

There is two different things: detecting that dune pkg is installed or not which is what we want to do here. And latter check if the project root does contain a dune pkg project (lock file).

PizieDust: So here, I think `availbale_sandboxes` checks if Dune.make returns something or None (which usually means Dune pkg is not installed).

src/extension_instance.ml
```diff
@@ -108,21 +108,68 @@ let stop_server t =
...
+  match t.sandbox with
+  | Dune dune ->
+    let open Promise.Syntax in
+    let+ dune_lsp_present = Dune.detect_dune_ocamllsp (Dune.root dune) () in
+    if dune_lsp_present
+    then Ok ()
+    else (
+      let (_ : Ojs.t option Promise.t) =
+        Vscode.Commands.executeCommand
+          ~command:Extension_consts.Commands.install_dune_lsp
```
pitag-ha, line 154: Here, inside your `check_ocaml_lsp_available` function, you're executing the  `install_dune_lsp` command.

src/extension_commands.ml
```diff
@@ -96,6 +96,89 @@ let _upgrade_ocaml_lsp_server =
...
+let _install_dune_lsp_server =
+  let handler (instance : Extension_instance.t) ~args:_ =
+    let open Promise.Syntax in
+    let _ =
+      match Extension_instance.sandbox instance with
+      | Dune dune ->
+        let* is_dune_locked = Dune.detect_dune_lock_dir (Dune.root dune) () in
+        if is_dune_locked
+        then
+          let* ocamllsp_present = Extension_instance.check_ocaml_lsp_available instance in
```
pitag-ha, line 108: Here, inside your `install_dune_lsp` command, you're calling `check_ocaml_lsp_available`.

src/extension_instance.ml
```diff
@@ -108,21 +108,68 @@ let stop_server t =
...
-  Cmd.output ?cwd (ocaml_lsp_version sandbox)
-  |> Promise.Result.fold
-       ~ok:(fun (_ : string) -> ())
-       ~error:(fun (_ : string) ->
-         "Sandbox initialization failed: `ocaml-lsp-server` is not installed in the \
-          current sandbox.")
+  ()
+;;
+
+let check_ocaml_lsp_available t =
```
pitag-ha, line 144: You're changing the parameter from type `sandbox` to type `instance`. However, the only field of `instance` you use is `sandbox`. Why did you make that change?

pitag-ha (changes requested):
Thanks, @PizieDust! It'll be very nice to have VSCode support while using dune package management!

I will make a bigger review with more comments on Monday. Here already the first thing that struck me: You have a few indirect mutual recursion cycles that I assume are unintentional. For example, the one I'm pointing out below between `check_ocaml_lsp_available` and `install_dune_lsp`.

src/sandbox.ml
pitag-ha: @PizieDust , `Dune.make` currently returns `None` if dune is not installed, right? That's an important start! Additionally to that, we also need to check if the installed dune version supports package management, i.e. if the `dune pkg` and `dune tools exec <tool>` commands are available.

src/extension_instance.ml
```diff
@@ -108,21 +108,68 @@ let stop_server t =
...
+  | Dune dune ->
+    let open Promise.Syntax in
+    let+ dune_lsp_present = Dune.detect_dune_ocamllsp dune in
+    if dune_lsp_present
+    then Ok ()
+    else (
+      let (_ : Ojs.t option Promise.t) =
+        Vscode.Commands.executeCommand
+          ~command:Extension_consts.Commands.install_dune_lsp
+          ~args:[]
```
pitag-ha, line 155: Here, the plug-in installs the dune lsp server without asking the user if they want to do that. Have you discussed this UX design choice? I personally would prefer to always ask the user before installing things on their computer.

src/extension_instance.ml
```diff
@@ -108,21 +108,68 @@ let stop_server t =
...
+      in
+      Ok ())
```
pitag-ha, line 157: This function is meant to check if the lsp server is installed. If the chosen sandbox is not dune pkg management, it returns an error if the server not installed. This case is then handled by the caller, which in that case suggests to install the server. That seems like a good control flow! However, if the sandbox is dune pkg, this function directly installs the server and then returns Ok.

I'd suggest to have the same behavior for all sandboxes: If the server is not installed, this function returns an Err and the rest is handled by the caller. Does that sound good to you?

pitag-ha (commented):
There's a lot of difference in the control flow depending on whether the chosen sandbox is dune package management or anything else.

The way anything else is handled is very smooth! It's as simple as (up to details): `start_language_server` -> Is lsp available (return Ok or Err). In case of Ok, register capabilities and start server. In case of Err, suggest to install the server. I think you've implemented that code as well, @PizieDust , right? It's a very nice piece of code!

However, when dune is selected as the package manager, the code is very tangled, has a few loops. The entry point of the tangles is my comment below.

What do you think about handling all sandboxes in a parallel way? Concretely, in the way the non-dune sandboxes are handled?

src/dune.ml
```diff
@@ -0,0 +1,38 @@
...
+  let open Promise.Syntax in
+  (* Should we do something specific for Windows ? *)
+  let spawn = { Cmd.bin = binary; args = [] } in
+  let+ spawn = Cmd.check_spawn spawn in
+  match spawn with
+  | Ok bin -> Some { bin; root }
+  | Error _ -> None
+;;
+
+let detect_dune_lock_dir t =
```
pitag-ha, line 20: Naming nitpick:

```suggestion
let is_project_locked t =
```

(Starting the name of a function that returns a bool with "is"/"has" or similar is quite a common naming convention. Simplifies a bit reading through the code later.)

src/dune.ml
```diff
@@ -0,0 +1,38 @@
...
+  (* Path to the dune.lock dir *)
+  let dune_lock_path = Path.join t.root (Path.of_string "dune.lock") in
+  Fs.exists (Path.to_string dune_lock_path)
+;;
+
+let detect_dune_ocamllsp t =
```
pitag-ha, line 26:
```suggestion
let is_ocamllsp_present t =
```

(same convention as above)

src/dune.ml
```diff
@@ -0,0 +1,38 @@
...
+  (* Path to the ocaml-lsp-server.pkg file *)
+  let ocamllsp =
+    Path.join
+      t.root
+      (Path.join (Path.of_string "dev-tools.locks") (Path.of_string "ocaml-lsp-server"))
+  in
+  Fs.exists (Path.to_string ocamllsp)
+;;
+
+let exec t ~args = Cmd.Spawn (Cmd.append t.bin args)
```
pitag-ha, line 36: This function is very generic. It's meant to execute any dune command. In my opinion, it'd be better to have different functions: One to execute a tool via dune (`<dune bin> exec tools <tool>`) and one to lock a project. The reasons:
- If the current preview dune API changes, we only need to adapt to that here in dune.ml and not all over the code base.
- That'd match the abstraction of the other sandboxes.

src/dune.mli
```diff
@@ -0,0 +1,13 @@
...
+  { root : Path.t
+  ; bin : Cmd.spawn
+  }
+
+val detect_dune_lock_dir : t -> bool Promise.t
+val detect_dune_ocamllsp : t -> bool Promise.t
+val exec : t -> args:string list -> Cmd.t
+val equal : t -> t -> bool
+val make : root:Path.t -> unit -> t option Promise.t
+val root : t -> Path.t
```
pitag-ha, line 13: If you could add a bit of documentation here (similar to the documentation in `opam.mli` and `esy.mli` etc), that'd be amazing.

src/extension_commands.ml
```diff
@@ -96,6 +96,92 @@ let _upgrade_ocaml_lsp_server =
...
+                  "An error occured while running `dune tools exec ocamllsp`: %s"
+                  err;
+                Ojs.null
+            in
+            let+ _ = Vscode.Window.withProgress (module Ojs) ~options ~task in
+            let _ =
+              Dune.exec dune ~args:[ "tools"; "exec"; "ocamllsp" ]
+              |> Cmd.output ~cwd:(Dune.root dune)
+            in
+            let _ = Extension_instance.start_language_server instance in
```
pitag-ha, line 142:
```suggestion
            let _ = Extension_instance.start_language_server instance in
```

I don't think you need to spawn the server manually here. IIUC, the `start_langauge_server` function will create a language client, which will spawn the srver as a child process. Let me know if you understand this differently than me. I'm new to this code base as well.

src/extension_commands.ml
```diff
@@ -96,6 +96,92 @@ let _upgrade_ocaml_lsp_server =
...
+            ())
+        else Extension_instance.suggest_to_run_dune_pkg_lock () |> Promise.return
+      | _ -> Promise.return ()
```
pitag-ha, line 145: This case is expected never to happen, right? I mean one would only install the dune lsp server if dune is selected as the sandbox. So I'd log something here to facilitade debugging if this case ends up being hit.

src/extension_commands.ml
```diff
@@ -96,6 +96,92 @@ let _upgrade_ocaml_lsp_server =
...
+          match result with
+          | Ok _ -> Ojs.null
+          | Error err ->
+            show_message `Error "An error occured while running dune pkg lock: %s" err;
+            Ojs.null
+        in
+        let+ _ = Vscode.Window.withProgress (module Ojs) ~options ~task in
+        let _ = Extension_instance.start_language_server instance in
+        ()
+      | _ -> Promise.return ()
```
pitag-ha, line 178: Same as above: I'd log something here to facilitate debugging if this case ends up being hit.

src/extension_commands.ml
PizieDust: Thanks for catching this. This case can be hit easily if the command is declared in `package.json` which should be case, and we should let the user know that it doesn't work in none dune-pkg environments. I have added a log but I think a toast pop-up warning will be a good approach to give the user instant feedback because just logging means the user should be savvy enough to check in the output panel (and also switch the type of output to the ocaml platform extension)

PizieDust:
@pitag-ha it's ready for a review :)

pitag-ha:
Perfect, thanks for the changes, @PizieDust ! I was meaning to review directly today, but in the end didn't get around to doing it. I'll do it asap!

pitag-ha (commented):
Thanks for the changes, @PizieDust  ! I think we're close to merging! Here are only a few more comments, several of them are minor comments.

One general thing: I wouldn’t set Dune Package Management (DPM) as the default sandbox. The VSCode plugin currently uses a nice heuristic to detect the sandbox that's most likely to work well. With this PR, DPM becomes the default for anyone with dune installed (i.e. almost everyone). However, only very few people use DPM at the moment. And in fact, for most people it doesn't work yet (they don't use the preview). I think instead of _always_ setting DPM as the default, it should be set as the default only if the project is locked. What do you think?

src/sandbox.ml
```diff
@@ -56,18 +56,22 @@ module Package = struct
   ;;
 end
 
+let empty_root = Path.of_string ""
+
```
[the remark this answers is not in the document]
PizieDust: I think there was an earlier implementation that used `workspace_root` directly but was changed. Please refer here:
https://github.com/ocamllabs/vscode-ocaml-platform/pull/1791#discussion_r2056386213

src/extension_instance.ml
```diff
@@ -147,9 +188,13 @@ end = struct
       ()
   ;;
 
-  let server_options sandbox =
+  let server_options t =
     let args = Settings.(get server_args_setting) |> Option.value ~default:[] in
-    let command = Sandbox.get_command sandbox "ocamllsp" args in
+    let command =
+      match t.sandbox with
+      | Dune dune -> Dune.exec_tool ~tool:"ocamllsp" ~args dune
```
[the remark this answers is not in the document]
PizieDust: I think this shim script checks if the project is locked, and then it runs `dune tools exec ocamlformat` and then proceeds to run `dune tools exec ocamllsp`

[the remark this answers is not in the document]
pitag-ha: Yes, it adds `ocamlformat` to the PATH of `ocamllsp`'s process. And that's exactly what we need.

I've already chatted with @gridbugs about this. He has confirmed what I've written above: The dune installation script adds this shim to the user's PATH. There's no opt-out for that. So it's safe for the VSCode plug-in to directly spawn ocamllsp via calling `ocamllsp`. Directly calling `ocamllsp` will run the shim, which will check if the project is locked. If so, it'll add `ocamlformat` to the PATH and then spawn `ocamllsp`.

So the conclusion is indeed: We need to spawn the LSP server by calling `ocamllsp` (which will execute the shim), not by calling `dune tools exec ocamllsp`.

PizieDust:
I pushed some more changes (refactoring `Sandbox.get_command` and also given that we now use the ocamllsp shim script this means ocamllsp can be available without a `dune-dev.locks` directory.

pitag-ha:
As a heads-up: I won't be able to review tomorrow or beginning of next week. I'll do so mid next week, though! I think the only point left to discuss is
> given that we now use the ocamllsp shim script this means ocamllsp can be available without a dune-dev.locks directory.

Using this assumption could easily lead to a mismatch between the compiler (installed via dune pkd management) and ocaml-lsp (installed via opam and found by the ocamllsp shim).

pitag-ha:
As an update for smorimoto : Pizie and I are going to discuss this in a call. We just need to use the ocamllsp shim correctly, and then this should be ready for merge.

pitag-ha (approved):
Heya, my slowness in reviewing (I'm sorry for this. I've also been sick these days) is blocking this to move forward. So let me suggest to merge this even though it isn't perfect and to fine-tune it in further PRs. Wdyt, @PizieDust and @smorimoto ? Hopefully, it'll be fixed before the next `vscode-ocaml-platform` release. And even if it isn't: DPM (dune package management) is still WIP and it's expected that things can still go wrong.

The main problems I see atm are:
1. With the current implementation, choosing DPM (dune package management) will either work or silently go wrong, depending on how dune got installed: 
    - if dune is not installed at all when selecting the sandbox, dune package management won't show up as an option  ✔️ (unlikely as really _most_ OCamlers use dune as their build system) 
    - if dune is installed via the binary distribution, dune pkg management shows up as an option and things will work correctly ✔️ 
    - if dune is installed via opam or nix or anything else than the binary distribution, dune package management will show up as an option, but when selected things will silently go wrong: vscode will spawn ocamllsp by calling `ocamllsp` directly, thinking that will run the shim. However, the shim doesn't exist in this case, so instead whichever ocamllsp is in the path will be spawned (e.g. the one from the global opam switch). ❌ Unfortunately, this is the most common case: most people have dune installed via opam.
2. There's currently no version check for dune. We've already decided to fix this in a separate PR (see issue #1833)

But the rest looks good! :heavy_check_mark: Thank you, @PizieDust !

Btw, one of the CI checks is failing, right? Have you had a look?

@PizieDust, thanks a lot for this work and for sticking both with my slow reviews and with the dune API changes! This has been (and will still be) complex work due to building on top of dune package management's WIP work.

PizieDust:
@smorimoto Can we re-run the workflows that are failing? The errors seem unrelated to this PR

smorimoto (approved):

PR #1802, opened 2025-05-06
mlx syntax highlight 

davesnx:
This PR adds ocaml.mlx as a language and pushes a new syntax file based on `ocaml.json` with JSX nodes. I managed to look into `reason.json` syntax to ensure tokens are named correctly.

At the start, I wanted to make use of injectTo and injectSources but I couldn't make it to work, so instead I "fork" the current ocaml syntax. Happy to revisit this decision, since I think the diff would be smaller and the future maintenance painless.

If injecting isn't working, also I could reuse the syntax with `{"include": "ocaml#xxx"}`

image


voodoos (approved):
I am not very knowledgeable in these areas, but the changes looks ok to me and are a clear improvement because they bring support for files that had none.

It looks like vscode is finally going to get tree-sitter support in the following months, it will probably be a good time to improve the "injection" story: https://github.com/microsoft/vscode/issues/210475

Meanwhile, I am curious about your last comment:
> If injecting isn't working, also I could reuse the syntax with {"include": "ocaml#xxx"}

You make it sounds like an easy way to remove a lot of code duplication, what's the catch 🙂 ?

davesnx:
Yeah, the catch is that I define the jsx-expressions and they only work in the "top-level" so, jsx-expressions inside let bindings, modules or attributes/extensions do not work.

Here's the commit to showcase the simplicity of the code, but it doesn't work as expected. If someone more experienced can recommend a better approach, I'm all ears, maybe @mnxn?

davesnx:
After more debugging, I think it's a limitation on how textmate works. Once I `include source.ocaml#xxx` and those contain $self, there's no way to take it back. It's like closed recursion, if that make sense. I might be wrong, but so far is what I discover

I could only define the places where I care like, re-create module or bindings, but it looks more fragile like vendoring the entire repository in mlx.json

voodoos:
That's fine for me. could you add a changelog entry ?

davesnx:
Done @voodoos

davesnx:
Thanks @smorimoto

PR #1861, opened 2025-06-23
Statically type commands

### LINKED ISSUE #965: Binding for `vscode.commands.registerCommand`: callback should return Ojs.t rather than unit
The ts type for `vscode.commans.registerCommand` looks like 

```ts
registerCommand(command: string, callback: (args: any[]) => any, thisArg?: any): [Disposable](https://code.visualstudio.com/api/references/vscode-api#Disposable)
```

so in our binding, which is `val registerCommand: command:string -> callback:(args:Js.Any.t list -> unit) -> Disposable.t`, `callback` should return `Js.Any.t` or `Ojs.t`, which are equivalent. 

It would be nice however to somehow witness the return type of a callback, so that a callback caller can work with the real return type rather than Ojs.t

### LINKED ISSUE #1778: [Bindings] Enable waiting for command execution
Our current binding for `registerCommand` restricts callbacks to returning `unit` that prevents obtaining a promise to wait for when calling `executeCommand`. This could be helpful to have in the future to implement more complex workflows.

Official API:
https://vscode-api.js.org/modules/vscode.commands.html#registerCommand
https://vscode-api.js.org/modules/vscode.commands.html#executeCommand

mnxn:
Adds a `('a, 'b) Command_api.handle` type to associate command IDs with argument and return types.

Instead of using the `registerCommand`/`executeCommand` functions from the VS Code API directly, the `Extension_commands.register` and `Command_api.execute` functions are used to typecheck and automatically convert values from/to JS types.

Previously, commands were called like this:
```ocaml
Vscode.Commands.executeCommand
  ~command:"vscode.openWith"
  ~args:
    [ Uri.t_to_js uri
    ; Ojs.string_to_js "ast-editor"
    ; ViewColumn.t_to_js ViewColumn.Beside
    ]
```

Now, it is simplified to:
```ocaml
Command_api.(execute Vscode.open_with) (uri, "ast-editor", ViewColumn.Beside)
```

Types have been added for the built-in commands the extension uses and most internal commands. Types haven't been added for internal commands with parameters, so those still accept an "untyped" `Ojs.t list` argument.

Closes #965, closes #1778


smorimoto:
@mnxn Could you rebase this?

mnxn:
Rebasing was becoming tedious, so I did a merge instead. Hopefully that's not a problem.

PR #1871, opened 2025-07-01
Reintroduce quoted string literal highlighting with correct ID handling

### LINKED ISSUE #1868: Highlighting issue when quoted strings contain quotes
Maybe related to https://github.com/ocamllabs/vscode-ocaml-platform/issues/930

There are some display issues when quoted string  `{|` contain regular quotes `"`: 

Image

Notice that here:
- the list separators `;` are all highlighted as strings (bright green in my theme), when they should not be.
- the final delimiter `|}` is not highlighted as a string.

The issue gets even worse if you have mismatched quotes, see here, when I add a single quote to the last string, the string coloration leaks into the rest of the file...

Image

Example code for quick testing:
```ocaml
String.concat ""
  [
    {|<img src="|};
    link;
    {|" alt="|};
    alt;
    {|" title="|};
    title;
    {|" >"|};
  ]
```

mnxn:
Closes #1868

The flawed grammar + semantic highlighting was more reliable than no grammar rules and semantic highlighting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add support for quoted string literal brackets (`{|`, `|}`) syntax highlighting and auto-closing pairs
+- Revert quoted string literal highlighting removal (#1871)
 
 ## 1.30.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Add support for quoted string literal brackets (`{|`, `|}`) syntax highlighting and auto-closing pairs
-- Revert quoted string literal highlighting removal (#1871)
+- Reintroduce quoted string literal highlighting with correct ID handling (#1871)
 
 ## 1.30.1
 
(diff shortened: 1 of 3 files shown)


dlesbre:
Is there a reason why the quoted string literal blocks using `{|` don't set the `name`, wheras the ones using `"` do? 

It seems to me that means that, if you disable semantic highlighting, only the standard quoted string will have color.

mnxn:
If you're looking at the diff, the JSON objects without `name` fields are for the string recognition within comments. This comment is unclosed for example:
```ocaml
(* {| *)
```
But this is ok:
```ocaml
(* {| |} *)
```
https://stackoverflow.com/a/20795424

The rules within the `strings` object have the `string.quoted.braced.ocaml` name.

mnxn:
Making this a draft while I look at backreferences again.

dlesbre:
I've pushed a commit that uses the backreferences on my fork: https://github.com/dlesbre/vscode-ocaml-platform/commit/28093351fd2953e673180324ebcd1983446fed4b, quick testing shows it seems to work:
image

mnxn:
Looks like that fixed it. Thanks, @dlesbre! I guess I just never found this method last time I tried.

I added your commit to this PR.

smorimoto (approved):

PR #1887, opened 2025-07-14
Upgrade to ppxlib 0.36.0

smorimoto:


src/ppx_tools/traverse_ast2.ml
```diff
@@ -741,37 +741,6 @@ class virtual ['res] lift2 =
           self#constr
             "Pexp_let"
             [ "rec_flag", a; "value_binding list", b; "expression", c ]
-        (*Caused by: [fun (t39 : t39 Js_of_ocaml.Js.t) -> ...] *)
```
patricoferris, line 744: I'm not exactly sure what these special cases are here for, but they are still able to be represented with the new 5.2 AST. So maybe they need to stay?

smorimoto: You're right that these special cases are still needed. However, since `Pexp_fun` was removed in ppxlib 0.36, I think we should implement similar compatibility cases using the new `Pexp_function` syntax instead. I've added equivalent patterns for `Pexp_function` wrapped in `Pexp_apply`, `Pexp_constraint`, and `Pexp_poly` to maintain the same compatibility coverage. But I'm not sure if this is correct.

smorimoto: Could you review again?

smorimoto: Thanks a lot!

PR #1903, opened 2025-07-31
Start the ocamllsp server as a dev tool

PizieDust:
This Pr no longer relies on the shim script to start ocamllsp server and rather starts it as a dune dev tool.

cc @pitag-ha

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Start ocamllsp server for DPM using dune tools exec ocamllsp (#1903)
+
 ## 1.31.0
 
 - Automatically configure Dune Package Management (#1791)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Start ocamllsp server for DPM using dune tools exec ocamllsp (#1903)
```
pitag-ha, line 5:
```suggestion
- Fix DPM (Dune Package Management) support for users with Dune installed via
  Opam. Previously, DPM support worked only when Dune was installed using
  the installation script for the Dune binary distribution. (#1903)
```

For the release notes, it's better to point out the user impact of the change than the technical change itself.

pitag-ha (approved):
LGTM. Thanks again!

Btw, for reference for others: When `ocamllsp` is spawned via `dune tools exec`, other dune dev tools (in partucular `ocamlformat`) are now included in its PATH. (see https://github.com/ocaml/dune/pull/12065 )

PR #1926, opened 2025-08-30
Standardise UI messages to use quotes instead of Markdown backticks

smorimoto:
Replace Markdown-style backticks in notifications and message popups with double quotes, since VSCode does not render Markdown in `showInformationMessage`/`showWarningMessage`/`showErrorMessage`.


smorimoto:
FYI: @pitag-ha @PizieDust

PR #1930, opened 2025-09-01
Change wording of dune notifications

PizieDust:
When vscode is launched the platform extension tries to find all sandboxes which are available. Since we want users to be more aware of DPM, we have a notification which reads `Dune Package Management selected with dune from [dune|opam], [1.20]`. This is confusing, as it displays regardless of if the user has selected DPM or not. 

We should change the wording to `Dune Package Management is available with dune from [dune|opam], [1.20]`

cc @pitag-ha @smorimoto

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Reword DPM notification to inform users DPM is available and not selected. (#1930)
+
 ## 1.32.1
 
 - Fix DPM error when invoking `ocamlc --version` without a working directory by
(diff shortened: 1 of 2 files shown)


PizieDust:
@smorimoto will this need a changelog?

smorimoto:
Yes, I guess!

smorimoto (approved):

RblSb:
I see this notification every time i open project after updating deps, but i'm too stupid to understand what it wants from me. Can you guys give some context about "selecting" DPM?
Upd: not checked issues first, sorry, there is https://github.com/ocamllabs/vscode-ocaml-platform/pull/2037

PR #1990, opened 2025-10-27
Implement the new `codelens.forNestedBindings` setting

Tim-ats-d:
Implement the new setting provided by this PR https://github.com/ocaml/ocaml-lsp/pull/1567


smorimoto (approved):
LGTM

Tim-ats-d:
The field `forNestedBindings` is now marked as optional in the JSON setting decoding. This no longer create incompatibility error when testing with previous OCaml LSP server version that does not handle the `forNestedBindings` option.

src/settings.mli
```diff
@@ -36,6 +36,7 @@ val substitute_workspace_vars : string -> string
 val server_extraEnv : unit -> string Interop.Dict.t option
 val server_args_setting : string list setting
 val server_codelens_setting : bool setting
+val server_codelens_for_nested_bindings_setting : bool option setting
```
voodoos: It feels wrong to have a `bool option setting`, I don't think vscode has UI for that.

Let's stick with a non-nullable bool and default to `false` when decoding if the field is absent using Jsonoo's `try_default` instead of `try_optional`). WDYT ?

src/settings.ml
```diff
@@ -120,11 +120,19 @@ let server_args_setting =
...
     ~of_json:Jsonoo.Decode.bool
     ~to_json:Jsonoo.Encode.bool
 ;;
 
+let server_codelens_for_nested_bindings_setting =
+  create_setting
+    ~scope:ConfigurationTarget.Workspace
+    ~key:"ocaml.server.codelens"
+    ~of_json:Jsonoo.Decode.(try_default false (field "forNestedBindings" bool))
+    ~to_json:Jsonoo.Encode.(bool)
```
mlantas: This can just be

```suggestion
    ~to_json:Jsonoo.Encode.bool
```

mlantas (approved):

PR #2027, opened 2025-11-28
Add `melange.emit` and `melange.runtime_deps` in `dune/suntax.json`

davesnx:


syntaxes/dune.json
```diff
@@ -1731,6 +1731,168 @@
             }
           ]
         },
+        {
+          "comment": "melange.emit",
+          "begin": "\\([[:space:]]*(melange\\.emit)\\b",
```
mlantas: Would it make sense to recognize `melange`, `.`, and `emit` as separate entities?
Then `.` could be a `keyword.operator.dune`.

I'd be interested to know how the official dune parser reads this.

mlantas (approved):
Just one comment, but non-blocking.

syntaxes/dune.json
davesnx: Done, I could do coq as well in another PR

PR #2131, opened 2026-04-12
Remove reliance on external which command in Dune.make

smorimoto:
## Problem

On native Windows, the `which` programme is not available. When `dune` is not found on PATH, `Dune.make` falls back to running `opam exec -- which dune` to locate it within the current opam switch. This fails on Windows with `[ERROR] Command not found 'which'`, preventing the extension from discovering dune installed via opam.

Reported in #2105.

## Solution

Rather than using `which` (or `where.exe` on Windows) to resolve the dune binary path and then invoking it directly, store `opam exec -- dune` as the `Dune.t.bin` spawn command. All command-generating functions (`command`, `exec`, `tools`) use `Cmd.append t.bin args`, so subsequent dune invocations are correctly issued as `opam exec -- dune <args>`.

This approach:
- Eliminates the need for any external path-resolution command (`which`, `where.exe`)
- Ensures the full opam environment is available for all subsequent dune commands
- Simplifies the fallback logic (removes an intermediate `which` step and its result parsing)

## Verification

- `dune build` succeeds
- `dune build @runtest` passes
- `dune build @fmt` passes
- Locally verified that `opam exec -- dune --version` correctly returns the version when run from a directory with a local opam switch

Closes #2105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Fix dune detection failing on Windows when `which` is unavailable by
+  running dune directly through `opam exec` instead. (#2131)
+
 ## 2.1.0
 
 - Fix version parsing for nightly and dev builds of dune. This resolves
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -121,7 +121,6 @@ let make root () =
   let dune_version_output bin root =
     command { root; bin } ~args:[ "--version" ] |> Cmd.output ~cwd:root
   in
-  (* Should we do something specific for Windows ? *)
   match root with
   | Some root ->
     let spawn = { Cmd.bin = binary; args = [] } in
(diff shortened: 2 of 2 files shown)


rgrinberg:
Don't we have `Cmd.which` already? I'm pretty sure we have a way to get the env of a switch as well to extract the PATH.

smorimoto:
@rgrinberg For getting the switch's PATH: there's no `opam env` helper, but `Switch.path` gives us the switch root, so we can derive the bin directory as `Switch.path opam switch / "bin"` and pass it to `Cmd.which`.

The main change needed would be threading `Opam.t` and `Switch.t` into `Dune.make` (it currently only takes `root : Path.t option`), since the callers in `sandbox.ml` already have access to the opam instance.

Does that match what you had in mind?

rgrinberg:
There might not be a helper, but `opam env --sexp` certainly exists. We might need to wrap this and run this only once per switch, but that seems like a good thing.

smorimoto:
Thanks for the suggestion! I went ahead and implemented the `opam env --sexp` approach, but a concern came up and I'd like to reconsider the direction.

The output of `opam env --sexp` contains the following variables beyond `PATH`:

- `OPAM_SWITCH_PREFIX`
- `CAML_LD_LIBRARY_PATH`
- `OCAML_TOPLEVEL_PATH`
- `OCAMLTOP_INCLUDE_PATH`

`opam exec --` sets all of these on the child process, but if we only extract `PATH` and resolve the dune binary via `Cmd.which`, the rest are left at whatever values the bare process environment provides. dune itself might still work, but subprocesses it spawns (ocamlc, ocamlfind, ocamllsp, etc.) are likely to depend on them, so paths like `dune exec` or `dune tools exec` (e.g. `dune tools exec ocamllsp`) could break.

Propagating the full env properly would mean adding an `env` field to `Cmd.spawn` and updating record literals across several places (`opam.ml`, `esy.ml`, `ocaml_windows.ml`, etc.) — a refactor beyond the scope of this PR.

With that in mind, I'd like to propose going back to the original approach of this PR (wrapping with `opam exec -- dune`):

- Consistent with the existing wrap pattern used by `Opam.exec` / `Esy.exec`.
- Env handling is delegated to `opam`, which is safer.
- Removing the external `which` dependency (the actual goal of this PR) is already achieved.
- `Dune.make` only runs at extension startup and on sandbox switch, so the `opam exec` overhead is negligible.

I think the env-extraction + propagation idea would be healthier as a separate PR with its own scope. What do you think?

rgrinberg:
Sure, if this is the best that can be done for now, then let's go with this. Let's just be mindful of process spawning on windows being notoriously slow.

PR #2184, opened 2026-06-22
Add e2e tests where we activate the extension

davesnx:
Last 2 versions: 2.2.0 and 2.2.1 contain runtime errors that are easily catchable by CI.

This activation test ensures no runtime errors occur at the binding boundary or any external issues with compatibility with the node_module dependencies.


.github/workflows/main.yml
```diff
@@ -22,15 +22,15 @@ jobs:
...
+          ocaml-compiler: 5.4
 
       - name: Set-up Mise
         uses: jdx/mise-action@v4
 
       - name: Install dependencies
         run: |
           bun install --frozen-lockfile
-          opam install . --deps-only --with-test
+          opam install . --deps-only --with-test --with-dev-setup
```
davesnx, line 33: This is needed to install ocamllsp as part of the CI

davesnx:
The CI is failing due to:

- Dependency on log-symbols@4.1.0 does require("is-unicode-supported")
- bun resvoles into is-unicode-supported@2.1.0, which is ESM-only
- vscode electron runs a require(...) and it's a module, not a function

davesnx:
Pushed an overlay that fixed the package "is-unicode-supported" under log-symbols (it's a bit hacky, but it's the only way I can think of when using bun + mocha, which doesn't currently support it fully https://github.com/mochajs/mocha/issues/5108)

smorimoto (approved):
Thanks a lot!

PR #2190, opened 2026-06-25
First draft of "Run file as standalone bytecode executable" feature.

Tim-ats-d:
This PR adds support for running standalone file or dune executable as byte code.
It adds a run button on the editor bench. 
image
When clicked, a dropdown listing executables discovered in the current workspace is displayed:
image

The way executables are discovered depends on the workspace setup. If the user uses Dune and has a `dune-project` at the root of the workspace, we try to parse the output of `dune describe` in order to get the list of defined executables. Otherwise, when no Dune setup is found, the list of `.ml` discovered in the workspace is displayed (ignoring `_build/`).

When the user chooses an executable to launch, execution takes place in a "Run OCaml" terminal output appearing at the bottom of the screen. Depending on the detected project context, the executable is run via `dune exec <exec_path>` or with `ocaml <filename>`.
image
If the current text editor active file is chosen, it is saved before it launched.

Finally, this PR also provide new task rule to launch executable with Dune. To use it : `Ctrl + Shift + P` -> `Tasks: Run Task` -> `Dune`. The following video demonstrates what if we try to execute a file that has error:
https://github.com/user-attachments/assets/3d06bf45-eff7-47f4-9673-f4c1904e5655
And this one, show executing a successful build:
https://github.com/user-attachments/assets/5a7abce8-1f51-460d-92e4-f1b99c6307d7


src/standalone_file.ml
```diff
@@ -0,0 +1,180 @@
...
+
+(*
+   Résumé:
+   - boutton qui s'affiche que sur les éditeurs ocaml actifs
+   - approximation optimiste
+   - Si dune-project repéré à la racine du workspace alors exécuter dune exec le nom du binaire ? dune describe
+
+   TODO:
+   - Tester sur macOS et windows avec et sans Opam
+*)
```
xvw: I guess that you can remove that :)

package.json
```diff
@@ -325,6 +325,20 @@
         "command": "ocaml.install-dune-lsp",
         "category": "OCaml",
         "title": "Start the Ocaml-LSP server for Dune Package Management"
+      },
+      {
+        "command": "ocaml.view-metrics",
+        "category": "OCaml",
+        "title": "View metrics"
+      },
```
xvw: Sorry for the naive nature of my question but what is the purpose of this command?

xvw (commented):
And I guess that the PR deserve a change-entry!

Tim-ats-d:
I think this can be reviewed now cc @xvw @voodoos

voodoos:
Can you update the PR description (or describe the current design in a new comment) and share a few screenshots of the UI to help with the review ? Thanks !

src/dune_task_provider.ml
```diff
@@ -58,33 +58,79 @@ let compute_tasks token sandbox =
...
+
+let compute_exec_tasks sandbox =
+  let open Promise.Syntax in
+  let+ executables =
+    let dune_describe =
+      Sandbox.get_command
+        sandbox
+        "dune"
+        [ "describe"; "--format"; "sexp"; "--lang"; "0.1" ]
+        `Command
```
xvw: Nitpick: Could we imagine having a dune helper in the source-tree or at least externalize this `dune describe` in order to reuse it in standalone-file (since the code is almost the same)?

xvw (approved):
I just left a nitpicking. But as says @voodoos, can you change the PR description (with screenshot) and I guess if we can maybe localize Dune helpers in order to invoke them quickly from the two source, and avoiding hard-coding stuff like the dune lang format?

src/dune
```diff
@@ -12,6 +12,7 @@
   opam-file-format
   polka
   promise_jsoo
+  sexplib
```
xvw, line 15: Should be added to the dune-project (or OPAM file)?

src/dune_task_provider.ml
Tim-ats-d: Yes, I can move this command in `dune.ml`

src/dune
Tim-ats-d: Nice catch, should be fixed

src/dune_task_provider.ml
Tim-ats-d: Or perhaps I can make directly a `dune_describe.ml` which contains the parser and provide the command needed?

xvw: I am fine with dune. Just to localize in one place that kind of hard-coding

Tim-ats-d: Should be fixed by the last commit.

xvw (approved):
Nice work @Tim-ats-d !

src/dune_describe.ml
```diff
@@ -0,0 +1,99 @@
+open Import
+open Sexplib
+open Option.Monad_infix
+open Stdlib.Option.Syntax
+
+type 'a parser = Sexplib.Sexp.t -> 'a option
```
voodoos:
```suggestion
type 'a parser = Sexp.t -> 'a option
```

src/dune_describe.ml
```diff
@@ -0,0 +1,99 @@
...
+            in
+            let+ rel_path = mod_impl_path mod_sexp in
+            let mod_path = String.chop_prefix_if_exists rel_path ~prefix:build_context in
+            let exec_path = Stdlib.Filename.remove_extension mod_path ^ ".exe" in
+            let exec_path =
+              if String.is_prefix exec_path ~prefix:"/"
+              then "." ^ exec_path
+              else "./" ^ exec_path
+            in
+            { name = exe_name; mod_path; exec_path })
```
voodoos, line 86: This does not look portable. You should use adequate functions from the `Filename` stdlib module like `concat` or, maybe better in that case, the bindings to nodejs' [path](https://nodejs.org/api/path.html#pathjoinpaths).

src/runnable.ml
```diff
@@ -0,0 +1,137 @@
...
+    ()
+  |> Promise.map (function
+    | [] -> Unknown
+    | _ -> Dune)
+;;
+
+let find_executables sandbox project_ctx =
+  let open Promise.Syntax in
+  match project_ctx with
+  | Dune ->
```
voodoos, line 20: Not sure how to materialize it, but this Dune branch does much more than handling "standalone" files...

The title of this module is misleading.

src/runnable.ml
```diff
@@ -0,0 +1,137 @@
...
+        ml_files
+    in
+    Some execs
+;;
+
+let exec_cmd sandbox project_ctx (exec : Dune_describe.executable) args =
+  let program, args =
+    match project_ctx with
+    | Dune -> "dune", [ "exec"; exec.exec_path; "--" ] @ args
+    | Unknown -> "ocaml", [ "-I"; "+str"; "-I"; "+unix"; exec.mod_path ] @ args
```
voodoos, line 60: Why specificaly `+str` ? Is that a common thing to do ?

src/standalone_file.ml
```diff
@@ -0,0 +1,137 @@
...
+let executable_choice_menu (execs : Dune_describe.executable list) =
+  let choices =
+    List.map
+      ~f:(fun exec ->
+        QuickPickItem.create ~label:exec.name ~detail:exec.exec_path (), exec)
+      execs
+  and options =
+    QuickPickOptions.create
+      ~canPickMany:false
+      ~placeHolder:"Which executable do you want to execute?"
```
voodoos:
```suggestion
      ~placeHolder:"Which executable do you want to run?"
```

src/runnable.ml
Tim-ats-d: I think nobody use the module Str, particularly in teaching context. Do you want me to remove it?

src/runnable.ml
Tim-ats-d: What do you think of renaming `standalone_file.ml` to `runnable.ml`?

src/runnable.ml
xvw: I think that it is a built-in module (not like `Graphics` now) and it save from errors.

Tim-ats-d: We agreed that it wouldn't cost anything to add it

src/dune_describe.ml
```diff
@@ -0,0 +1,99 @@
...
+                modules
+            in
+            let+ rel_path = mod_impl_path mod_sexp in
+            let mod_path = String.chop_prefix_if_exists rel_path ~prefix:build_context in
+            let exec_path = Stdlib.Filename.remove_extension mod_path ^ ".exe" in
+            let exec_path =
+              if Stdlib.Filename.is_relative exec_path
+              then exec_path
+              else Stdlib.Filename.current_dir_name ^ exec_path
+            in
```
voodoos, line 85: Let's comment on the fact that moving to using the [Node JS Path API](https://nodejs.org/api/path.html) would be a safer solution if issues are encountered in the future.

voodoos (approved):

voodoos:
@Tim-ats-d I just tested and I have a question:

There are two different ways to run things:
1. The standard `Tasks: Run Build Task` command that runs the chosen task in a terminal
2. The "Run button" that runs the executable and shows the output in an "output" panel

The issue is, if I am running a server, how do I kill it in case 2 ?
Also the output panel does not supports terminal colors, so it's a worse experience.

I think 2 should also run in a terminal.

src/dune_task_provider.ml
```diff
@@ -58,33 +58,66 @@ let compute_tasks token sandbox =
...
+    in
+    let name = Printf.sprintf "build %s" relative_path in
+    let execution =
+      let cwd = Stdlib.Filename.dirname (Uri.fsPath dune) in
+      let options = ShellExecutionOptions.create ~env ~cwd () in
+      get_shell_execution sandbox ~sub_cmd:"build" options
+    in
+    let task =
+      Task.make
+        ~definition
```
voodoos, line 75: We should use a different definition than "build". Maybe "exec" ?

voodoos:
And of course there is the lockfile issue: https://github.com/ocaml/dune/issues/15631

voodoos:
And also, but this one is for later, JS executables, like `vscode_ocaml_platform` are also part of the list but they cannot be run using the `.exe` extensions. And dune-describe does not tell us the "mode" so there is no way we can do something specific right now. Can you open issues on both vscode-ocaml-platform and dune about this once this PR is merged ?

src/dune_task_provider.ml
voodoos: Let's ignore that comment.

PR #2193, opened 2026-06-29
Use the command `dune pkg enabled` to detect Dune Package Management context

Tim-ats-d:
Previously, we detect DPM context by checking for a `dune.lock` lock file in the current workspace. This method is not reliable.


PizieDust (approved):
Thanks @Tim-ats-d , LGTM

PR #2199, opened 2026-06-30
New workflow for setting Dune Package Manager as a "sandbox"

PizieDust:
In this PR: 
## summary
we define two scopes for dune:
- global: by running `eval $(opam env --revert) dune --version` and
- switch: checking which opam switches have dune installed
    - ordering them by global (by version)
    - then by local (by version)

This is then compiled in a list and displayed to the user in the order:

- dune in currently active switch
- dune in system path
- custom path to dune
- dunes in global switches
- dunes in local switches

## usage
Activating DPM now happens in two steps:

1) Selecting Dune Package Manager as the sandbox
image

2) After selecting DPM, a list of all dune versions available is listed. The one the user selects is then registered as the dune binary to be used in the project.
image

This PR is only concerned with setting DPM as the current "sandbox" and updating the `.vscode` config. Proper usage of the dune binary after that is not handled in this PR. Subsequent PR's will be done to evaluate and make any corrections if needed.

### DIFF
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -99,11 +99,11 @@ type t =
   ; bin : Cmd.spawn
   }
 
-let make ~root_dir ~dune_path =
+let make ~working_dir ~dune_path =
   let open Promise.Syntax in
   let* spawn = Cmd.check_spawn { Cmd.bin = dune_path; args = [] } in
   match spawn with
-  | Ok bin -> Promise.return (Some { root = root_dir; bin })
+  | Ok bin -> Promise.return (Some { root = working_dir; bin })
   | Error _ ->
     log_chan
       `Info
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -220,9 +220,9 @@ let get_system_dune_path () =
        See: https://github.com/ocaml/dune/issues/11161 *)
     Promise.return None
   | Darwin | Linux | Other ->
-    let bash_script =
-      "if command -v opam >/dev/null 2>&1; then eval $(opam env --revert || true); fi; "
-      ^ "command -v dune && dune --version"
+    let shell_script =
+      "if which opam >/dev/null 2>&1; then eval $(opam env --revert || true); fi; "
+      ^ "which dune && dune --version"
     in
-    run_cmd "bash" [ "-c"; bash_script ]
+    run_cmd "sh" [ "-c"; shell_script ]
 ;;
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -106,7 +106,7 @@ let make ~root_dir ~dune_path =
   | Ok bin -> Promise.return (Some { root = root_dir; bin })
   | Error _ ->
     log_chan
-      `Error
+      `Info
       ~section:"dune package management"
       "Dune not found in the environment.";
     Promise.return None
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -97,8 +97,6 @@ type t =
   ; bin : Cmd.spawn
   }
 
-let binary = Path.of_string "dune"
-
 let make ~root ~path =
   let open Promise.Syntax in
   let* spawn = Cmd.check_spawn { Cmd.bin = path; args = [] } in
@@ -110,12 +108,6 @@ let make ~root ~path =
       ~section:"dune package management"
       "Dune not found in the environment.";
     Promise.return None
-;;
-
-let is_project_locked t =
-  (* Path to the dune.lock dir *)
-  let dune_lock_path = Path.join t.root (Path.of_string "dune.lock") in
-  Fs.exists (Path.to_string dune_lock_path)
 ;;
 
 let command t ~args = Cmd.Spawn (Cmd.append t.bin args)
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -179,7 +179,6 @@ let is_dune_in_switch opam switch =
 ;;
 
 let all_opam_switches_with_dune opam switches =
-  let open Promise.Syntax in
   Promise.List.filter_map (fun switch -> is_dune_in_switch opam switch) switches
 ;;
 
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -87,7 +87,7 @@ module Dune_version = struct
   let is_valid version =
     match version with
     | Release (major, minor, patch) ->
-      Stdlib.compare (major, minor, patch) (3, 20, 0) >= 0
+      Stdlib.compare (major, minor, patch) (3, 24, 0) >= 0
     | Preview (y, m, d) -> Stdlib.compare (y, m, d) (2026, 6, 11) >= 0
   ;;
 end
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -105,10 +105,7 @@ let make ~working_dir ~dune_path =
   match spawn with
   | Ok bin -> Promise.return (Some { root = working_dir; bin })
   | Error _ ->
-    log_chan
-      `Info
-      ~section:"dune package management"
-      "Dune not found in the environment.";
+    log_chan `Info ~section:"dune package management" "Dune not found in the environment.";
     Promise.return None
 ;;
 
--- a/src/dune.ml
+++ b/src/dune.ml
@@ -5,6 +5,24 @@ module Dune_version = struct
     | Release of int * int * int
     | Preview of int * int * int
 
+  let to_string t =
+    match t with
+    | Release (major, minor, patch) -> Printf.sprintf "release-%d.%d.%d" major minor patch
+    | Preview (y, m, d) -> Printf.sprintf "nightly-%d-%d-%d" y m d
+  ;;
+
+  let compare v1 v2 =
+    match v1, v2 with
+    | Release (major1, minor1, patch1), Release (major2, minor2, patch2) ->
+      let open Stdlib in
+      compare (major1, minor1, patch1) (major2, minor2, patch2) |> Ordering.of_int
+    | Preview (y1, m1, d1), Preview (y2, m2, d2) ->
+      let open Stdlib in
+      compare (y1, m1, d1) (y2, m2, d2) |> Ordering.of_int
+    | Release _, Preview _ -> Ordering.of_int 1
+    | Preview _, Release _ -> Ordering.of_int (-1)
+  ;;
+
   let parse_release_version version_str =
     let extract_int s =
       let digits = String.filter s ~f:Char.is_digit in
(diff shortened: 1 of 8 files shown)


src/dune.ml
```diff
@@ -109,36 +141,89 @@ let is_ocamllsp_present t =
...
+  Opam.var opam switch ~args:[ "dune:bin" ]
+  |> Cmd.output
+  >>= function
+  | Error err ->
+    log_chan
+      `Error
+      ~section:"dune package management"
+      "Dune is not installed in switch %s: %s"
+      (Opam.Switch.name switch)
+      err;
```
voodoos, line 163: That's not an error, right ? More an info(rmation)

src/dune.ml
```diff
@@ -109,36 +141,89 @@ let is_ocamllsp_present t =
...
+    Promise.return None
+  | Ok path ->
+    command
+      { root = Path.of_string ""
+      ; bin = { Cmd.bin = Path.(of_string (Stdlib.String.trim path) / "dune"); args = [] }
```
voodoos: Why did _you_ prefix the `String.trim` function here ? Base just reexport the Stdlib one anyway.

src/dune.ml
```diff
@@ -109,36 +141,89 @@ let is_ocamllsp_present t =
...
+;;
+
 let root t = t.root
+let dune_path t = t.bin.bin
+
+let is_dune_in_switch opam switch =
+  let open Promise.Syntax in
+  Opam.var opam switch ~args:[ "dune:bin" ]
+  |> Cmd.output
+  >>= function
```
voodoos, line 156: I'm less used to these operators these days and find monadic lets more readable, but feel free to keep them like it if you prefer that style.

src/dune.ml
```diff
@@ -109,36 +141,89 @@ let is_ocamllsp_present t =
...
+    match output with
+    | Ok output ->
+      (match String.split_lines (String.strip output) with
+       | [ path; version_str ] ->
+         let version = Dune_version.from_string (String.strip version_str) in
+         (match version with
+          | Some v -> Promise.return (Some (String.strip path, v))
+          | None -> Promise.return None)
+       | _ -> Promise.return None)
+    | Error _ -> Promise.return None
```
voodoos: If every branch of your matching finished with a `Promise.return` then you should have used `let+`

src/dune.ml
```diff
@@ -109,36 +141,89 @@ let is_ocamllsp_present t =
...
+  | Win32 ->
+    (* Dune package management is not supported on Windows at the moment
+       See: https://github.com/ocaml/dune/issues/11161 *)
+    Promise.return None
+  | Darwin | Linux | Other ->
+    let bash_script =
+      "if command -v opam >/dev/null 2>&1; then eval $(opam env --revert || true); fi; "
+      ^ "command -v dune && dune --version"
+    in
+    run_cmd "bash" [ "-c"; bash_script ]
```
voodoos: I don't think we should expect bash to be installed on the user system.

I think [`Sys.command`](https://ocaml.org/manual/5.5/api/Sys.html#VALcommand) is the correct way to do this sort of things. But you might alwo use `which` because `command -v` might be shell specific (not an expert).

src/sandbox.ml
```diff
@@ -90,6 +90,7 @@ let to_string = function
   | Opam (_, switch) -> Printf.sprintf "opam(%s)" (Opam.Switch.name switch)
   | Global -> "global"
   | Custom _ -> "custom"
+  (* TODO: it should be dune(<version>) *)
```
voodoos: todo ?

src/sandbox.ml
```diff
@@ -258,14 +258,11 @@ let of_settings () : t option Promise.t =
...
-    (match dune with
-     | None ->
-       not_available `Dune;
-       Promise.return None
-     | Some dune -> Promise.return (Some (Dune dune)))
+  | Some (Dune path) ->
+    let open Promise.Option.Syntax in
+    let* root = workspace_root () |> Promise.return in
+    let* dune = Dune.make ~root ~path in
+    Promise.return (Some (Dune dune))
```
voodoos, line 264: Use `let+`

src/sandbox.ml
```diff
@@ -305,9 +302,9 @@ let detect_opam_sandbox ~project_root opam () =
   Opam (opam, switch)
 ;;
 
-let detect_dune_pkg ~project_root _dune () =
+let detect_dune_pkg ~project_root () =
   let open Promise.Syntax in
-  Dune.make (Some project_root) ()
+  Dune.make ~root:project_root ~path:project_root
```
voodoos: I am not sure I understand the difference between root and path. Maybe they could have more descriptive names ?

src/sandbox.ml
```diff
@@ -394,25 +394,197 @@ module Candidate = struct
...
+    let task ~progress:_ ~token:_ =
+      let* opam_dunes = Dune.get_opam_dunes opam in
+      let* current_switch =
+        let open Promise.Option.Syntax in
+        let* opam = Promise.return opam in
+        Opam.switch_show ~cwd:root opam
+      in
+      let* system_dune = Dune.get_system_dune_path () in
+      data := Some (opam_dunes, current_switch, system_dune);
+      Promise.return ()
```
voodoos: Use `let+`

src/sandbox.ml
```diff
@@ -394,25 +394,197 @@ module Candidate = struct
...
+               let description =
+                 switch_kind_s ^ " | Currently active switch in project root"
+               in
+               Some
+                 ( create
+                     ~label
+                     ~description
+                     ~detail:(Printf.sprintf "%s" (Stdlib.String.trim path ^ "/dune"))
+                     ()
+                 , `Opam_dune Path.(of_string (Stdlib.String.trim path) / "dune") ))
```
voodoos: Same remark about Stdlib prefixing

src/sandbox.ml
```diff
@@ -394,25 +394,197 @@ module Candidate = struct
...
+           ( create
+               ~label:
+                 (Printf.sprintf
+                    "Dune %s, via opam switch %s"
+                    (Dune.Dune_version.to_string version)
+                    switch_name)
+               ~detail:(Printf.sprintf "%s" (Stdlib.String.trim path ^ "/dune"))
+               ()
+           , `Opam_dune Path.(of_string (Stdlib.String.trim path) / "dune") )
+         in
```
voodoos: Same about Stdlib. Also, is there a nice way to factor a bit this part  with the previous one ? The tricky path handling is done separately multiple time which is a bit scary.

src/sandbox.ml
```diff
@@ -394,25 +394,197 @@ module Candidate = struct
...
+             Printf.sprintf
+               "Dune %s, via system PATH"
+               (Dune.Dune_version.to_string version)
+           in
+           Some
+             ( create ~label ~description:"System Dune" ~detail:path ()
+             , `Opam_dune (Path.of_string path) )
+         | None -> None
+       in
+       let custom_dune_item =
```
voodoos, line 559: This function is already quite large, maybe you could move the mechanic for selecting a custom path in it's own function.

src/sandbox.ml
```diff
@@ -394,25 +394,197 @@ module Candidate = struct
...
+                 | Some version when Dune.Dune_version.is_valid version ->
+                   Promise.return None
+                 | Some version ->
+                   let msg =
+                     Printf.sprintf
+                       "Dune version %s is not supported. Please use a more recent \
+                        version."
+                       (Dune.Dune_version.to_string version)
+                   in
+                   Promise.return (Some msg)))
```
voodoos: Use `let+`...

src/sandbox.ml
```diff
@@ -484,21 +656,17 @@ let sandbox_candidates ~workspace_folders =
...
-    in
-    List.concat dune_projects
-    |> List.map ~f:(fun dune -> { Candidate.sandbox = Dune dune; status = Ok () })
+    match Platform.t with
+    | Win32 -> Promise.return None
+    | Darwin | Linux | Other ->
+      let dummy_path = Path.of_string "" in
+      let+ dummy_dune = Dune.make ~root:dummy_path ~path:dummy_path in
+      (match dummy_dune with
+       | Some dummy_dune -> Some { Candidate.sandbox = Dune dummy_dune; status = Ok () }
```
voodoos: What is the purpose of the dummy dune ?

src/dune.ml
PizieDust: we don't have a `trim` function in `String` or `Base.String`

voodoos:
That's strange, it appeared to work in my testing. Anyway, the base équivalent is strip !
https://ocaml.org/p/base/latest/doc/base/Base/String/index.html#val-strip

2 juil. 2026 11:23:21 PixieDust ***@***.***>:

src/sandbox.ml
PizieDust: originally this was just there so the user always sees a `Dune Package Manager` in their sandbox list. but now, i just refactored it so that the user can only see `Dune Package Manager` in the list if there's at least one valid dune version.

src/dune.ml
voodoos: Ah my bad, I must have bortched my testing. With Base the correct replacement appears to be `String.strip`

src/dune.mli
```diff
@@ -40,7 +42,7 @@ val is_ocamllsp_present : t -> bool Promise.t
 (** Check if amy two instances of dune pkg management projects are equal *)
 val equal : t -> t -> bool
 
-val make : root:Path.t -> path:Path.t -> t option Promise.t
+val make : root_dir:Path.t -> dune_path:Path.t -> t option Promise.t
```
voodoos: I still don't understand of what this is the `root` 🙂

Tim-ats-d (approved):

src/dune.mli
Tim-ats-d: It's essentially the working directory where Dune process should be launched, what about naming it `working_dir`?

voodoos: Ah I see, yes I think `working_dir` would be clearer in that case.

src/sandbox.ml
PizieDust: I moved some functionality into their own functions: particularly:
- getting all dune binaries
- finding dune in the active switch 
- finding dune in all the other global and local switches
- validating input given in the custum dune textbox

src/dune.ml
PizieDust: `Sys.commad` only returns integers.

voodoos: Still, this won't work on windows, will it ?

I think the correct way around is the same as the one we discussed for starting the ocaml-lsp-server in slack:
Run `opam env --revert` and parse it's output. Feed it to `spawn ~env`. 

cc @Tim-ats-d

PizieDust: yes this is not expected to work on windows. the dune team said windows is not supported yet, so in this function we actually match what platform the extension is running on, and if its windows, we return none

voodoos: When it comes to compatibility running a shell is always tricky. But if you tested this thoroughly and works on all supported platforms we can keep it. If you want to aim for robustness then feeding `spawn` the env from `opam env --revert` is the way to go. It's your call :-)

voodoos:
This looks good overall, tell me when it's ready for another round of review :-) (next week)

src/dune.ml
voodoos: Oh and there is this function that looks interesting even if we are not going for windows support today: `Cmd.to_spawn`

PizieDust: I think the latest commit addresses this :)

voodoos (approved):
Thanks, looks reasonable.

voodoos:
I wanted to test this PR but the extension fails to start on my machine:
```
Activating extension 'ocamllabs.ocaml-platform' failed: Cannot read properties of undefined (reading 'delimiter').
$onExtensionActivationError @ mainThreadExtensionService.ts:108
_doInvokeHandler @ rpcProtocol.ts:458
_invokeHandler @ rpcProtocol.ts:443
_receiveRequest @ rpcProtocol.ts:373
_receiveOneMessage @ rpcProtocol.ts:299
(anonymous) @ rpcProtocol.ts:161
_deliver @ event.ts:1364
fire @ event.ts:1395
fire @ ipc.net.ts:658
(anonymous) @ localProcessExtensionHost.ts:421
```

@smorimoto is this a known issue ?

PR #2208, opened 2026-07-02
Enforce users to enable dune package management if not enabled in a DPM sandbox

PizieDust:
This PR is checked from https://github.com/ocamllabs/vscode-ocaml-platform/pull/2199

when DPM is selected as the sanbox and a dune binary chosen, if DPM is not enabled in the workspace, then it will display a modal for the user to either: 
- generate a lockfile (the recommended way by the dune team) to enable DPM
- select a different sandbox

pressing escape will display the select_Sandbox quickpick ui 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fix the detection of Dune Package Management to use the recommended `dune pkg enabled` command. Previously, we were checking to see if a `dune.lock` file exists. (#2193)
 - Fix the launch of Utop REPL in context of Dune Package Management. (#2198)
 - Redesign how Dune Package Manager is handled. Users now use a two step process, first to select DPM as a sandbox and then to select which dune binary they wish to use in the project (#2199)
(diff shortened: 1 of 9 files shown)


voodoos (commented):
The last three commits which are the ones specific to this PR look good to me.

PizieDust:
I think `.vscode/settings` still get's updated once the user has selected a binary. will have to figure a way to prevent this if the user doesn't lock their project.

Tim-ats-d (approved):

PizieDust:
> I think `.vscode/settings` still get's updated once the user has selected a binary. will have to figure a way to prevent this if the user doesn't lock their project.

This is now fixed.

voodoos:
Can you rebase this now that the other pr is merged ?

PizieDust:
> Can you rebase this now that the other pr is merged ?

Thanks. it's done now :)

voodoos:
There are still multiple commits unrelevant to that PR...

voodoos (approved):

PR #2217, opened 2026-07-08
Rely on `dune tools env` to launch ocamllsp in the context of DPM.

Tim-ats-d:


src/extension_instance.ml
```diff
@@ -121,15 +150,37 @@ end = struct
...
+    let open Promise.Syntax in
     let args = Settings.(get server_args_setting) |> Option.value ~default:[] in
-    let command = Sandbox.get_command t.sandbox "ocamllsp" args `Tool in
+    let command =
+      match t.sandbox with
+      | Dune dune ->
+        (* FIXME: there is currently a bug in DPM causing the compilers to be recompiled
+           each time we try to execute OCaml LSP. As a workaround, we run ocamllsp
+           without the "dune tools exec" command. *)
+        Cmd.Spawn { bin = Path.of_string "ocamllsp"; args }
```
voodoos: In the end, we probably will stick with this version to start the server: it would not be suitable to have a slow process that install or update the server here.

src/extension_instance.ml
```diff
@@ -121,15 +150,37 @@ end = struct
...
     in
+    let+ env =
+      match t.sandbox with
+      | Dune dune ->
+        let+ output = Cmd.output ~cwd:dune.root (Dune.env dune) in
+        let paths =
+          String.drop_suffix
+            (Result.ok_or_failwith output)
+            1 (* Remove the last line feed. *)
+          |> String.chop_prefix_exn ~prefix:"export PATH="
```
voodoos: What if dune env changes and add more variables than PATH ? 
I suggest you make this more robust either by stopping explicitly at the end of the path (somthing like `Scanf.sscanf "export PATH=/Users/ulysse/tmp" "export PATH=%s" (fun s -> print_string s);;`, or by acting as if the output was actually a list of multiple exports and properly parse all `VAR=VAL` pairs you encounter and add all of them to env.

Tim-ats-d: Fixed ty.

voodoos (approved):

PR #2218, opened 2026-07-09
Add ocamlgrep command (OCaml: Search expressions with ocamlgrep)

### LINKED ISSUE #2160: Ocamlgrep integration
Please see https://github.com/ocaml/ocaml-lsp/issues/1607

The code was so far generated by Claude Code and I haven't reviewed it. If you want to see it anyway, it's at https://github.com/mjambon/vscode-ocaml-platform/pull/1

mjambon:
[Ocamlgrep](https://github.com/LexiFi/ocamlgrep) is a command that searches for OCaml expressions that match a query pattern which may contain holes and may contain type constraints. We [announced](https://discuss.ocaml.org/t/ann-ocamlgrep-0-1-1/18350) it earlier this week on Discuss.

This PR makes ocamlgrep search available in VSCode.

How it works: the Dune root is determined from the active file, then an `ocamlgrep` command is launched and scans the `.cmt` files of the whole Dune workspace.

Previous work tried to integrate ocamlgrep in Merlin, then in the LSP server, only to conclude that it's simpler to directly invoke it from here and consume its JSON output.

Demo:

https://github.com/user-attachments/assets/55164f70-129e-481d-9299-267f73ce5d1a

Test plan:

* Build and install normally using `make build` && `make install`
* `make test`
* Interactive testing: Open VSCode, use Ctrl+Shift+P to run "OCaml: Search expressions with ocamlgrep" - it should tell you to install ocamlgrep
* `opam install ocamlgrep`
* Open an OCaml project, run "OCaml: Search expressions with ocamlgrep". Enter an OCaml expression such as `failwith` or `__ = __` or `(__ : string)` or any expression you like, and check that you get results that make sense, grouped by matching file.

Status: ready for third-party review


src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,199 @@
...
+         ~args:
+           [ [%js.of: Uri.t] anchor_uri
+           ; [%js.of: Position.t] anchor_pos
+           ; [%js.of: Location.t array] locations
+           ])
+;;
+
+let scan_folder query (scan_root : Path.t) : Custom_requests.Ocamlgrep.response Promise.t =
+  let cmd =
+    Cmd.(Spawn { bin = Path.of_string "ocamlgrep"; args = [ "--format"; "json"; query ] })
```
smorimoto: Spawning the executable directly only searches the Extension Host's PATH. In the usual setup where ocamlgrep is installed only in the selected Opam, Esy, or Dune sandbox, this reports that the executable is missing and the command cannot be used. Please obtain the sandbox from instance and use Sandbox.get_command for both the availability check and the search invocation.

src/custom_requests.ml
```diff
@@ -157,6 +157,52 @@ module Merlin_jump = struct
...
+    ; warnings : string list
+    ; errors : string list
+    }
+
+  let decode_response ~resolve_path response =
+    let open Jsonoo.Decode in
+    let decode_pos json =
+      let row = field "row" int json in
+      let col = field "column" int json in
+      Position.make ~line:row ~character:col
```
smorimoto: The upstream JSON schema defines column as a zero-based byte offset within the line, whereas VS Code's Position.character is a UTF-16 code-unit offset. Any non-ASCII text before a match therefore shifts the reference range and may navigate to the wrong text. Please read the relevant line and convert the byte offset to a UTF-16 offset before constructing the Position.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,199 @@
...
+      "ocamlgrep is not installed. Please install it with: opam install ocamlgrep";
+    Promise.return ()
+  | Ok _ -> func ()
+;;
+
+(* This ignores VSCode workspace boundaries and may return a root folder
+   that is an ancestor of one of the VSCode workspaces. *)
+let get_scan_root (text_editor : TextEditor.t) =
+  let active_file_path =
+    text_editor |> TextEditor.document |> TextDocument.uri |> Uri.path |> Path.of_string
```
smorimoto: Uri.path is the URI path component rather than a platform-native file-system path, so it does not correctly represent Windows drive-letter or UNC paths for fs.realpath. This causes Dune root discovery to fail on Windows. Please use Uri.fsPath here, and consider rejecting non-file URI schemes explicitly.

src/ext_ocamlgrep.ml
mjambon: oh, nice. Done. Runs `opam exec --switch=... -- ocamlgrep ...` in the case of an opam sandbox.

src/custom_requests.ml
mjambon: Done. Please let me know if you see a cleaner or more fitting way of going about this (I'm still new to js_of_ocaml).

src/ext_ocamlgrep.ml
mjambon: Done. It works on Linux... and hopefully on Windows too.

mjambon:
In its current state, a missing `ocamlgrep` results in a purely informational message that tells the user to run `opam install ocamlgrep`. It's not as good as showing a button to perform the installation and another to change the sandbox like it's done for the LSP server.

Do you prefer that I do it here or in a separate pull request?

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,202 @@
...
+           [ [%js.of: Uri.t] anchor_uri
+           ; [%js.of: Position.t] anchor_pos
+           ; [%js.of: Location.t array] locations
+           ])
+;;
+
+let scan_folder sandbox query (scan_root : Path.t)
+  : Custom_requests.Ocamlgrep.response Promise.t
+  =
+  let cmd = Sandbox.get_command sandbox "ocamlgrep" [ "--format"; "json"; query ] `Exec in
```
smorimoto: Please add the -- end-of-options delimiter before query. Cmdliner otherwise treats patterns beginning with a hyphen as options, so valid OCaml expressions such as -__ fail with an 'unknown option' error. Passing --format json -- QUERY ensures that every pattern is parsed as the positional PATTERN argument.

mjambon: nice catch, thanks. Fixed.

mjambon:
I'm not sure what's going on with the Nix build. There seems to be a confusion regarding which `Promise` module is being used (?).
```
[3](https://github.com/ocamllabs/vscode-ocaml-platform/actions/runs/29561488927/job/88207049343#step:4:594)
File "src/ext_ocamlgrep.ml", line 137, characters 41-42:
137 |                      if Ojs.has_property e "message"
                                               ^
Error: The value e has type Promise.error
```

smorimoto:
You can ignore the Nix build! It's failing because it's picking up an old OCaml version, but I don't use Nix, so I'm not planning to fix it. If none of the contributors actively use Nix, I'm even considering removing it altogether...

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+    Command_api.Command_errors.text_editor_must_be_active
+      extension_name
+      ~expl:"An OCaml file must be open to determine the project root."
+    |> show_message `Error "%s"
+  | Some text_editor when not (is_valid_text_doc (TextEditor.document text_editor)) ->
+    show_message
+      `Error
+      "Invalid file type. This command only works in OCaml source files."
+  | Some text_editor ->
+    with_check_for_ocamlgrep sandbox (fun ocamlgrep_cmd ->
```
smorimoto, line 201: `ocamlgrep_cmd` is never used. The default Dune development profile treats warning 27 as an error, so `dune build` and `make test` stop here even though the release-profile packaging job succeeds. Since `with_check_for_ocamlgrep` passes `()`, please destructure the argument as `fun () ->`.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+           ; [%js.of: Position.t] anchor_pos
+           ; [%js.of: Location.t array] locations
+           ])
+;;
+
+let scan_folder sandbox query (scan_root : Path.t)
+  : Custom_requests.Ocamlgrep.response Promise.t
+  =
+  let cmd =
+    Sandbox.get_command sandbox "ocamlgrep" [ "--format"; "json"; "--"; query ] `Exec
```
smorimoto, line 79: `Sandbox.Custom` expands `$args` by joining them with spaces without shell quoting. A query such as `__ = __` therefore reaches `ocamlgrep` as multiple positional arguments, while patterns containing parentheses or shell metacharacters may be interpreted by the shell. Please preserve each argument boundary when constructing custom sandbox commands, preferably through structured spawning or proper platform-specific escaping.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+    in
+    previous := Some disposable;
+    InputBox.show input_box
+;;
+
+let with_check_for_ocamlgrep sandbox func =
+  (* Check for the presence of the ocamlgrep command. It may be wrapped
+     in an 'opam exec --switch=... -- ocamlgrep ...' invocation. *)
+  let cmd = Sandbox.get_command sandbox "ocamlgrep" [ "--help" ] `Exec in
+  let* result = Cmd.run cmd in
```
smorimoto, line 158: For a Dune sandbox this probe becomes `dune exec ocamlgrep -- --help`, but it inherits the Extension Host process working directory while the actual scan later uses `scan_root`. If the host was started outside the project, an executable available through the project lock is reported as missing and the scan is never attempted. Please determine the scan root first and run this probe with the same `~cwd:scan_root`.

smorimoto (commented):
I found three issues worth addressing in the direct ocamlgrep command path.

src/ext_ocamlgrep.ml
smorimoto, line 158: **Availability probe runs without `~cwd`, breaking the Dune (DPM) sandbox**

The installed-check runs `Cmd.run cmd` without `~cwd`, so cwd-dependent sandbox commands — for Dune Package Management, `dune exec ocamlgrep -- --help` — execute in the extension host's cwd instead of the workspace, unlike the actual scan which passes `~cwd:scan_root`. Compare `Dune.is_dpm_enabled` in `src/dune.ml`, which passes `~cwd:t.root` for exactly this reason.

Failure scenario: with the DPM sandbox selected and ocamlgrep available, the probe runs in the extension host's cwd (e.g. `/`), where dune fails with "I cannot find the root of the current workspace/project" and a nonzero exit code, so the user unconditionally gets the "ocamlgrep is not installed" toast and can never reach the query input box — even though the scan itself (run with `~cwd:scan_root`) would have worked.

src/ext_ocamlgrep.ml
smorimoto, line 79: **Custom sandbox: the query is spliced unquoted into a shell string (shell injection)**

For the Custom sandbox, `Sandbox.get_command` splices the raw user-typed query into a shell string via `String.concat ~sep:" "` and runs it with `sh -c` (`Cmd.run`'s Shell branch uses `child_process.exec`), so the query is neither passed as a single argument nor shell-escaped.

Failure scenario: with a Custom sandbox template (e.g. `opam exec -- $prog $args`), the placeholder query `List.filter __ __` reaches ocamlgrep as three separate arguments instead of one query; a typical pattern like `List.map (fun x -> x)` is a shell syntax error ("unexpected token `('") so the command fails with exit code 2; and a query containing `$(...)` or backticks executes arbitrary shell commands.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+  : Custom_requests.Ocamlgrep.response Promise.t
+  =
+  let cmd =
+    Sandbox.get_command sandbox "ocamlgrep" [ "--format"; "json"; "--"; query ] `Exec
+  in
+  Cmd.log cmd;
+  let+ result : ChildProcess.return = Cmd.run ~cwd:scan_root cmd in
+  (* Exit code 1 ("no matches") is not an error. *)
+  match result.exitCode with
+  | 0 | 1 ->
```
smorimoto, line 85: **Wrapper exit code 1 is conflated with ocamlgrep's own "no matches" convention**

`scan_folder` treats exit code 1 as ocamlgrep's "no matches" convention, but the observed exit code is the sandbox wrapper chain's: `dune exec` (DPM) also exits 1 when the workspace build or target resolution fails, so wrapper failures are routed into the JSON-parse path instead of the stderr-reporting error branch.

Failure scenario: under a Dune sandbox, the project fails to build: `dune exec ocamlgrep ...` exits 1 with the real error on stderr and nothing on stdout, `Jsonoo.try_parse_opt` returns `None`, and the user is shown the misleading warning "ocamlgrep: unexpected output from workspace root ..." while the actual dune build error (on stderr) is never surfaced in the popup.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+      ~expl:"An OCaml file must be open to determine the project root."
+    |> show_message `Error "%s"
+  | Some text_editor when not (is_valid_text_doc (TextEditor.document text_editor)) ->
+    show_message
+      `Error
+      "Invalid file type. This command only works in OCaml source files."
+  | Some text_editor ->
+    with_check_for_ocamlgrep sandbox (fun ocamlgrep_cmd ->
+      with_scan_root text_editor (fun root -> show_query_input sandbox root text_editor))
+    |> ignore
```
smorimoto, line 203: **Uncaught `realpath` rejection makes the command silently do nothing**

`Dune_root.find` calls `Node.Fs.realpath` (which rejects when the path does not exist — see the comment in `src/dune_root.ml`) with no catch anywhere on the chain, and `callback` discards the resulting promise with `|> ignore`, so the rejection is silently swallowed.

Failure scenario: the user has `foo.ml` open, then its directory disappears from disk (e.g. `git checkout` to a branch without that directory, a rename/delete in a terminal, or any dangling-symlink component). Running "Search expressions with ocamlgrep" then does nothing at all — no input box, no error message — because the realpath rejection propagates through `with_scan_root`/`with_check_for_ocamlgrep` into the ignored promise as an unhandled rejection.

src/ext_ocamlgrep.ml
smorimoto, line 201: **`ocamlgrep_cmd` parameter is bound to `()` and unused**

The callback parameter `ocamlgrep_cmd` is bound to `()` and unused — `with_check_for_ocamlgrep` invokes `func ()`. It should be `fun () ->`, and the current name falsely implies the resolved sandbox command is passed to the continuation (it is not; `scan_folder` re-resolves it).

Note: under dune's default dev-profile flags, warning 27 (unused variable) is fatal, so `make build` may fail on this line depending on the applied flags.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+;;
+
+let scan_folder sandbox query (scan_root : Path.t)
+  : Custom_requests.Ocamlgrep.response Promise.t
+  =
+  let cmd =
+    Sandbox.get_command sandbox "ocamlgrep" [ "--format"; "json"; "--"; query ] `Exec
+  in
+  Cmd.log cmd;
+  let+ result : ChildProcess.return = Cmd.run ~cwd:scan_root cmd in
```
smorimoto, line 82: **Custom sandbox: `child_process.exec`'s 1 MiB `maxBuffer` default truncates large results**

For the Custom sandbox, `Cmd.run` takes the Shell branch (`child_process.exec`) and `ChildProcess.Options.create` exposes no `maxBuffer`, so Node's documented 1 MiB default applies: if exceeded, the child process is terminated and output is truncated. ocamlgrep's JSON, which embeds the full source lines of every finding, easily exceeds that on broad queries.

Failure scenario: a Custom-sandbox user searches a common pattern (e.g. `List.map __ __`) in a large workspace; ocamlgrep produces more than 1 MiB of JSON, Node SIGTERMs the child mid-output, and the truncated stdout fails `Jsonoo.try_parse_opt`, so real matches are replaced by the misleading "unexpected output from workspace root ..." error toast. Spawn-based sandboxes (Global/opam/esy/dune) are unaffected.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+  let cmd = Sandbox.get_command sandbox "ocamlgrep" [ "--help" ] `Exec in
+  let* result = Cmd.run cmd in
+  match result.exitCode with
+  | 0 -> func ()
+  | _exit_status ->
+    (* TODO: Show buttons to perform the install in the current sandbox or
+       to change sandbox, just like we do for ocaml_lsp_server. *)
+    show_message
+      `Error
+      "ocamlgrep is not installed. Please install it with: opam install ocamlgrep";
```
smorimoto, line 166: **Install hint hardcodes opam regardless of the active sandbox**

The not-installed message hardcodes "Please install it with: opam install ocamlgrep" regardless of the active sandbox, which is wrong advice for Esy and Dune Package Management sandboxes where the probe command doesn't consult the opam switch at all.

Failure scenario: an Esy-sandbox user sees the message, runs `opam install ocamlgrep` as instructed, retries — and gets the identical error again, because the probe runs `esy -P <manifest> ocamlgrep --help`, which never sees opam-installed binaries.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+              (let+ response =
+                 scan_folder sandbox query scan_root
+                 |> Promise.catch ~rejected:(fun (e : Promise.error) ->
+                   let msg =
+                     if Ojs.has_property e "message"
+                     then [%js.to: string] (Ojs.get_prop_ascii e "message")
+                     else "unknown error"
+                   in
+                   show_message `Error "ocamlgrep: %s" msg;
+                   Promise.return Custom_requests.Ocamlgrep.empty_response)
```
smorimoto, line 144: **Decoder failure produces two contradictory notifications**

The rejection handler shows an Error toast, then returns `empty_response`, which flows into `display_results` whose `[], []` branch shows a second, contradictory Info toast "ocamlgrep: no matches found.". Additionally, an OCaml exception (`Jsonoo.Decode_error`) has no JS `message` property, so the first toast reads "ocamlgrep: unknown error".

Failure scenario: a future/older ocamlgrep emits JSON whose shape differs from the decoder's expectations (e.g. a renamed field), so `decode_response` raises and the promise rejects: the user gets "ocamlgrep: unknown error" (Error) immediately followed by "ocamlgrep: no matches found." (Info) — two contradictory notifications, with no diagnostic detail recorded anywhere.

src/dune_root.ml
```diff
@@ -0,0 +1,93 @@
...
+  | Dune_workspace of Path.t
+  | Dune_project of Path.t
+
+(* Unlike the rest of the Path module, this is not a pure syntactic operation
+   so I'm not sure if there's a better place to expose this function.
+
+   It resolves symlinks and returns the canonical absolute path.
+   Raises a JS exception if the path does not exist.
+*)
+let realpath (x : Path.t) : Path.t Promise.t =
```
smorimoto, line 25: **Physical (realpath'd) root walk diverges from dune's logical path walk**

`Dune_root` walks the physical (realpath'd) directory chain, whereas the dune reference implementation it cites (`bin/workspace_root.ml`, which uses `Filename.dirname` on the logical cwd and never calls realpath) walks the logical path. As a result the scan root — and every finding URI resolved against it in `ext_ocamlgrep.ml`'s `resolve_path` — becomes a physical path that can differ from the paths VS Code has open.

Failure scenario: a workspace opened through a symlinked path (e.g. macOS `/tmp` → `/private/tmp`, or `~/proj` → `/data/proj`): the scan root realpaths to `/data/proj`, ocamlgrep emits paths relative to that cwd, and every References-panel entry gets a `/data/proj/...` URI; clicking a result for the file the user is editing opens a duplicate editor tab outside the workspace instead of jumping within the open `~/proj/...` document. Secondarily, a source subtree symlinked from physically outside any dune project makes `find` fail with "cannot find a Dune root folder" where dune itself (logical walk from the same directory) finds the root.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+    ~prompt:
+      "Enter an OCaml expression pattern. Use __ as a wildcard, __1/__2 for \
+       metavariables, and (e : t) to constrain types."
+    ()
+;;
+
+let log_to_output lines =
+  let channel = Lazy.force Output.extension_output_channel in
+  List.iter lines ~f:(fun line -> OutputChannel.appendLine channel ~value:line);
+  OutputChannel.show channel ~preserveFocus:true ()
```
smorimoto, line 30: **Output panel is force-revealed on every search**

`log_to_output` calls `OutputChannel.show` on every invocation, so the "OCaml Platform Extension" Output panel is force-revealed both when the input box opens (scan-root log) and again when every search completes, including fully successful ones.

Failure scenario: every time the user runs an ocamlgrep search — even one that succeeds and shows its results in the References panel — the bottom Output panel pops open (twice per search), covering part of the editor until the user manually closes it each time. Consider only calling `show` on the error paths.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+      | None -> ()
+      | Some d -> Disposable.dispose d
+    in
+    let disposable =
+      InputBox.onDidAccept
+        input_box
+        ~listener:(fun () ->
+          match InputBox.value input_box with
+          | Some query when String.length query > 0 ->
+            InputBox.hide input_box;
```
smorimoto, line 133: **No busy indication or cancellation while a scan is running**

The accept listener hides the input box immediately and provides no in-flight indication or cancellation for the potentially long-running scan, dropping the busy/enabled guard that the `Search_by_type` input-box pattern this code follows applies while its request is pending (`InputBox.set_busy true` / `set_enabled false` in `extension_commands.ml`).

Failure scenario: on a large workspace the subprocess takes many seconds; after pressing Enter the box vanishes and no UI element indicates a search is running, so the user assumes the command failed and re-invokes it, spawning concurrent ocamlgrep processes whose results race to overwrite the References panel (and there is no way to cancel a scan mistakenly started with an expensive pattern).

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+                     else "unknown error"
+                   in
+                   show_message `Error "ocamlgrep: %s" msg;
+                   Promise.return Custom_requests.Ocamlgrep.empty_response)
+               in
+               display_results query text_editor response)
+          | _ -> ())
+        ()
+    in
+    previous := Some disposable;
```
smorimoto, line 150: **Potential race: overlapping invocations on the singleton input box**

`show_query_input` replaces the accept-listener on the module-level singleton `input_box` with last-async-writer-wins semantics, so two overlapping invocations can leave the surviving listener wired to the earlier invocation's `scan_root`/`text_editor`.

Possible scenario: invoke the command in editor A of project A; while its pre-flight probe and `Dune_root.find` are still pending, switch to editor B in project B and invoke again. B's chain resolves first and registers B's listener; then A's slower chain disposes B's listener (via the `previous` ref) and registers A's closure capturing A's `scan_root`/editor. The visible input box is the same singleton with an identical static title, so the user believes they are searching project B, but pressing Enter runs ocamlgrep with cwd = project A's root — results for the wrong project with no indication of the mix-up.

package.json
```diff
@@ -299,6 +299,11 @@
         "category": "OCaml",
         "title": "Search a value by type or polarity"
       },
+      {
+        "command": "ocaml.ocamlgrep",
```
smorimoto, line 303: **`ocaml.ocamlgrep` is missing a commandPalette `when` clause**

The new `ocaml.ocamlgrep` command has no `menus.commandPalette` `when` clause, unlike its sibling `ocaml.search-by-type` which is gated on `editorLangId == ocaml || ...`, so it is offered in the palette in any editor context.

Failure scenario: a user editing a JSON or Markdown file opens the command palette, sees and runs "OCaml: Search expressions with ocamlgrep", and gets the error popup "Invalid file type. This command only works in OCaml source files." — whereas the analogous "Search a value by type or polarity" command is hidden in the same context.

src/ext_ocamlgrep.ml
```diff
@@ -0,0 +1,204 @@
...
+   which closes on focus loss.
+*)
+let display_results query text_editor (response : Custom_requests.Ocamlgrep.response) =
+  let { Custom_requests.Ocamlgrep.findings; warnings; errors } = response in
+  log_to_output
+    ((sprintf "ocamlgrep %S: %d finding(s)" query (List.length findings)
+      :: List.map warnings ~f:(fun w -> "  " ^ w))
+     @ List.map errors ~f:(fun e -> "  Error: " ^ e));
+  match errors, findings with
+  | first_error :: _, _ ->
```
smorimoto, line 46: **Matches are discarded when the response also contains an error**

`display_results` matches `| first_error :: _, _ ->` first, so when ocamlgrep returns both findings and an error, the found matches are discarded and never shown in the References panel. ocamlgrep's `search` (`lib/Ocamlgrep.cppo.ml`) returns all findings collected before a failure together with `error = Some msg`, so this combination is a real output shape.

Failure scenario: ocamlgrep scans half the workspace, collects 40 matches, then aborts with an error (e.g. an unreadable directory); the JSON contains both findings and `error`. The user sees only the warning toast "ocamlgrep: ... (see output panel for details)" and none of the 40 matches, even though they were successfully found and decoded. Consider showing the warning *and* the partial results.

src/custom_requests.ml
```diff
@@ -157,6 +157,91 @@ module Merlin_jump = struct
...
+  (* string length in UTF-16 code units *)
+  include [%js: val js_string_length : string -> int [@@js.get "length"]]
+
+  (* Translate an offset in bytes into an offset in UTF-16 code units
+     i.e. the length of the corresponding JavaScript string *)
+  let translate_column ~line_array ~first_line_index ~line_index ~column_bytes =
+    let substr =
+      try
+        let line = line_array.(line_index - first_line_index) in
+        String.sub line ~pos:0 ~len:column_bytes
```
smorimoto, line 185: **`translate_column`'s fallback to column 0 can invert a range**

`translate_column` maps any out-of-range byte column or missing line to UTF-16 character 0 via the catch-all `Invalid_argument` handler, which can invert the range when only the end position is broken.

Possible scenario: ocamlgrep reports a finding whose `end.column` exceeds the text provided in `lines` (or whose `lines` array doesn't cover every row of the range): `String.sub` raises, the end character silently becomes 0 while the start character stays e.g. 10, VS Code's `Range` constructor swaps the inverted positions, and the References panel highlights the wrong span (column 0 up to the match start) instead of the actual match.

PR #2238, opened 2026-07-22
Move the OCaml version check to prevent compiler building freeze

Tim-ats-d:
In Dune Package Management context, if your project is set up, but you haven't already launched the initial `dune build`, `ocamlc` is not present in your path. In order to check if there is no newer version of `ocaml-lsp-server` for a given version of OCaml, we run `ocamlc -version`. Within DPM, we run `dune exec ocamlc -- -version` which triggers silently a compilation of the compiler without giving feedback to the user and makes fail all Dune command ran after (due to the lock on `_build/`).

This PR fix this problem by executing the OCaml version update just after checking that OCaml LSP is present (having OCaml LSP within your DPM project necessary means your compiler is already built). 

This means:
- If you are in a setuped DPM project without having launched the initial `dune build`, opening this project in VSCode will trigger the OCaml LSP installation via (`dune tools install ocamllsp`)
- If OCaml LSP is present, launch it

This PR also remove some dead code related to fetching the OCaml version which is not used anymore.

cc @voodoos @PizieDust


voodoos (commented):
That looks good overall. I am just a bit stressed by the use of `ocaml_version_exn` now that we fill it late.

It looks like it is okay but I think we should expose a function that returns an option instead of raising. Then just do nothing or show a warning in  `ocaml_lsp_doesnt_support_type_selection` if it is none.

And `ocaml_lsp_doesnt_support_type_selection` is a very bad workaround for a capability by the way, we should fix this at some point.

PizieDust:
@Tim-ats-d While testing the PR I got the log:

```
$ /home/ndahi/work/tarides/new/_opam/bin/dune pkg lock
Solution for dune.lock

Dependencies common to all supported platforms:
- compiler-cloning.enabled

Additionally, some packages will only be built on specific platforms.

arch = arm64; os = linux:
- ocaml.5.5.0
- ocaml-base-compiler.5.5.0
- ocaml-compiler.5.5.0

arch = arm64; os = macos:
- ocaml.5.4.1
- ocaml-base-compiler.5.4.1
- ocaml-compiler.5.4.1
- relocatable-compiler.5.4.1.20251109.1

arch = x86_64; os = linux:
- ocaml.5.5.0
- ocaml-base-compiler.5.5.0
- ocaml-compiler.5.5.0

arch = x86_64; os = macos:
- ocaml.5.4.1
- ocaml-base-compiler.5.4.1
- ocaml-compiler.5.4.1
- relocatable-compiler.5.4.1.20251109.1

$ /home/ndahi/work/tarides/new/_opam/bin/dune tools which ocamllsp
_build/_private/default/.dev-tool/ocaml-lsp-server/target/bin/ocamllsp

$ /home/ndahi/work/tarides/new/_opam/bin/dune exec ocamlc -- -version
Warning: As this is not the main instance of Dune it is unable to locate the
executable "ocamlc" within this project. Dune will attempt to resolve the
executable's name within your PATH only.
Error: Program 'ocamlc' not found!
```

The warning message, is that expected?
```
Warning: As this is not the main instance of Dune it is unable to locate the
executable "ocamlc" within this project. Dune will attempt to resolve the
executable's name within your PATH only.
```

pitag-ha:
FWIW:

> Warning: As this is not the main instance of Dune it is unable to locate the
executable "ocamlc" within this project. Dune will attempt to resolve the
executable's name within your PATH only.

This is an issue in Dune: ocaml/dune/issues/13676

PizieDust:
I addressed the review comments here: https://github.com/Tim-ats-d/vscode-ocaml-platform/pull/1

voodoos (approved):

PR #2244, opened 2026-07-22
Kill alive processes at desactivation of the extension

Tim-ats-d:
This PR adds the ability to attempt to kill all the child processes spawned by the extension. This is particularly useful in DPM context where the user can close VSCode and leave alive the compiler build process in the background.

Spawned process get their PID and their associated `ChildProcess` object memorized and they are removed from the hash table once they exit. We export a new function `deactivate` in the extension entry point that kills all residual process once the extension is deactivated.

See https://code.visualstudio.com/api/get-started/extension-anatomy#extension-entry-file for more information.


voodoos (approved):
That looks reasonable to me. There are some merge conflicts however.

PR #2245, opened 2026-07-22
Update available OCaml LSP version.

Tim-ats-d:
It is used to suggest upgrading OCaml LSP server and seems staled since releases compatible with 5.3.

cc @voodoos


voodoos (approved):
Thanks !

PR #2246, opened 2026-07-23
Walthrough auto-next

PizieDust:
This PR implements an auto-next feature for the ocaml walkthrough. 

image

The idea is that once a user has completed a step, they should not have to manually click to go to the next step. We should open it for them.

### DIFF
--- a/src/extension_commands.ml
+++ b/src/extension_commands.ml
@@ -32,6 +32,16 @@ let text_editor_command handle callback =
   command
 ;;
 
+let advance_walkthrough_step step_id =
+  let category = "ocamllabs.ocaml-platform#ocaml-onboarding" in
+  let (_ : unit Promise.t) =
+    let open Promise.Syntax in
+    let* () = Node.setTimeout' ~wait_ms:500 in
+    Command_api.(execute Vscode.open_walkthrough) (category, step_id)
+  in
+  ()
+;;
+
 let _select_sandbox =
   let callback (instance : Extension_instance.t) () =
     let open Promise.Syntax in
@@ -43,7 +53,8 @@ let _select_sandbox =
         Extension_instance.set_sandbox instance new_sandbox;
         let* () = Sandbox.save_to_settings new_sandbox in
         let* () = Extension_instance.update_ocaml_info instance in
-        Extension_instance.start_language_server instance
+        let+ () = Extension_instance.start_language_server instance in
+        advance_walkthrough_step "install-platform-tools"
     in
     ()
   in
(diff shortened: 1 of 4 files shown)


src/extension_commands.ml
```diff
@@ -314,7 +325,7 @@ let _install_opam =
             show_message `Error "An error occured while installing opam %s" err
         in
         let+ () = Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task in
-        ()
+        advance_walkthrough_step "init-opam"
```
voodoos, line 328: You probably don't want to "auto-next" if something went bad right ?

src/extension_commands.ml
```diff
@@ -399,7 +411,9 @@ let _init_opam =
       | Error err -> show_message `Error "An error occured while initializing opam %s" err
     in
     let (_ : unit Promise.t) =
-      Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task
+      let open Promise.Syntax in
+      let+ () = Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task in
+      advance_walkthrough_step "activate-opam-switch"
```
voodoos, line 416: Same here, the task should return information about the status (a bool or an error) so that we don't auto next if it went bad

src/extension_commands.ml
```diff
@@ -437,7 +451,9 @@ let _install_ocaml_dev =
         show_message `Error "An error occured while installing packages %s" err
     in
     let (_ : unit Promise.t) =
-      Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task
+      let open Promise.Syntax in
+      let+ () = Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task in
+      advance_walkthrough_step "check-installation"
```
voodoos, line 456: Same here

src/extension_commands.ml
```diff
@@ -472,7 +488,9 @@ let _open_utop =
       | Error err -> show_message `Error "An error occured while opening utop %s" err
     in
     let (_ : unit Promise.t) =
-      Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task
+      let open Promise.Syntax in
+      let+ () = Vscode.Window.withProgress (module Interop.Js.Unit) ~options ~task in
+      advance_walkthrough_step "finish-onboarding"
```
voodoos, line 493: Same here

voodoos (commented):
Looks cool but we don't want to auto-next when things go badly, and I think that is what would happen here.

PR #2254, opened 2026-07-31
Some fixes to `run-file` features

Tim-ats-d:
- Use terminal instead of an output channel (this adds support for colored compiler diagnostics)
- Dedicate a terminal per executable; if the user has already tried to run it, the previous terminal is reused. If the user closes the previously used term, we create another one.
- Always show the executable corresponding to the current viewed file in the executable dropdown if present


voodoos (approved):
Thanks !

PR #2276, opened 2026-09-04
Fix portability issues in TextMate grammars

Alhadis:
This pull-request fixes several structural issues with this extension's grammars that affect their portability to other TextMate-compatible highlighting engines… specifically, GitHub's [own](https://github.com/github-linguist/linguist/blob/main/CONTRIBUTING.md#fixing-syntax-highlighting) syntax highlighting.

The site currently uses [`textmate/ocaml.tmbundle`](https://github.com/textmate/ocaml.tmbundle) for highlighting OCaml syntax (and that repository hasn't been updated in 11 years), so I plan on replacing it with yours (which is clearly actively-maintained). However, some non-portable syntax needs to be fixed first (and don't worry, it won't have any user-facing changes).

### DIFF
--- a/syntaxes/atd.json
+++ b/syntaxes/atd.json
@@ -22,7 +22,9 @@
             "1": { "name": "keyword.other.atd" },
             "2": { "name": "keyword.other.atd" }
           },
-          "endCaptures": [{ "name": "keyword.other.atd" }],
+          "endCaptures": {
+            "1": { "name": "keyword.other.atd" }
+          },
           "patterns": [{ "include": "$self" }]
         }
       ]
--- a/syntaxes/atd.json
+++ b/syntaxes/atd.json
@@ -23,7 +23,7 @@
             "2": { "name": "keyword.other.atd" }
           },
           "endCaptures": {
-            "1": { "name": "keyword.other.atd" }
+            "0": { "name": "keyword.other.atd" }
           },
           "patterns": [{ "include": "$self" }]
         }
--- a/syntaxes/cram.json
+++ b/syntaxes/cram.json
@@ -135,24 +135,36 @@
         {
           "begin": "\\[\\^?",
           "end": "\\]",
-          "beginCaptures": [{ "name": "keyword.other.regex" }],
-          "endCaptures": [{ "name": "keyword.other.regex" }],
+          "beginCaptures": {
+            "1": { "name": "keyword.other.regex" }
+          },
+          "endCaptures": {
+            "1": { "name": "keyword.other.regex" }
+          },
           "contentName": "source.regex",
           "patterns": [{ "name": "keyword.other.regex", "match": "-" }]
         },
         {
           "begin": "\\((\\?[:=!])?",
           "end": "\\)",
-          "beginCaptures": [{ "name": "keyword.other.regex" }],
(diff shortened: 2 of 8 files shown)


Alhadis:
~~The `oxfmt` check failed. It would be more helpful if CI described where and what the formatting offences were, because I'm not an OCaml user, and I'd rather avoid the rigmarole of installing a toolchain  I've no familiarity with (none of the other JSON files I edited provoked an `oxfmt` error, so I suspect this is probably an easy fix.~~

**UPDATE:** I noticed one of the grammar rules was missing a space after its opening brace, unlike very other single-line object literal that's formatted that way. I'm blindly assuming that this was what `oxfmt` was complaining about.

I've no idea how to address the changelog-related CI failure, so somebody will need to add the `no changelog` label to flag these modifications as transparent to the extension's users.

smorimoto:
The array form `beginCaptures: [{ "name": … }]` means capture **0** (the whole match), not group 1 — so converting it to `{ "1": … }` only preserves behaviour where the regex actually has a group 1.

36 of the converted blocks don't, so they now reference a group that doesn't exist and the scope is silently dropped. Counting dangling capture references: 4 on `715202f` (all pre-existing, in `mlx.json`/`reason.json`) vs 40 here. Affected: `dune.json:3906` (`%{…}`), `menhir.json` (16), `ocamlbuild.json` (6), `cram.json` (6), `ocaml.json:132` + `mlx.json:291` (Cinaps), `ocamllex.json:69` (2), `atd.json:20`. Using `"0"` fixes all of them — `mlx.json:73` already does. (`cram.json`'s `\((\?[:=!])?` is a 37th case: group 1 is *optional*, so token boundaries shift.)

Separately, `menhir.json:314`: `(?<![[:word:]])[[:space:]]*\(` no longer excludes `foo (`, because the lookbehind is now tested at the `(`, which is preceded by a space — so `opt (x) = maybe` parses as a destructuring binding. `(?<![[:word:][:space:]])[[:space:]]*\(` is fixed-width and keeps the exclusion.

Alhadis:
> The array form `beginCaptures: [{ "name": … }]` means capture **0** (the whole match), not group 1

Oops, my bad. I'm surprised that didn't occur to me. I've fixed the incorrectly-numbered captures in eb47d0f.

> Separately, `menhir.json:314`: `(?<![[:word:]])[[:space:]]*\(` no longer excludes `foo (`,

But the [existing pattern](https://github.com/ocamllabs/vscode-ocaml-platform/blob/75ca9658f7cad56b61e88347d34989d69e4ab9a6/syntaxes/menhir.json#L281) only matches whitespace optionally (`(?<![[:word:]][[:space:]]*)\(`), and the rule doesn't assign any scopes to the contents of the matched text, so `(?<=\w)\s*\(` and `(?<=\w\s*)\(` are effectively identical in this scenario.

Alhadis:
@smorimoto How do you propose I proceed with the `(?<![[:word:]][[:space:]]*)\(` issue, given my feedback?

Alhadis:
@smorimoto …okay, I've an even better question: How the hell do I fix this changelog check? It ostensibly wants me to add a label to this PR, which <em>only users with write-access to this repository can do</em>.

